PR #4, opened 2017-02-14
Catch known built-in SSL provider errors

Safihre:
We (well, our users) have been using CherryPy with the built-in SSL module quite extensively and a number of errors have come up that are thrown by the built-in provider that can just be ignored.
Using the `pyopenssl` module, these are caught by `pyopenssl`, but not by `ssl`.

An easy example is running the http://testssl.sh/ script against a CherryPy server that uses the built-in provider with any type of certificates.

Maybe we can add some logging about dropped connections to the access log? I wasn't sure, because nowhere in these modules the logging is used.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v5.2.0
+======
+
+Prevent tracebacks and drop bad HTTPS connections in the
+``BuiltinSSLAdapter``, similar to ``pyOpenSSLAdapter``.
+
 v5.1.0
 ======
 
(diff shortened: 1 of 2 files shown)


Safihre:
No feedback? :pensive:

jaraco:
I don't really have a comment on whether this is right or not. It seems suitable.

You're right that you can't log to the cherrypy access log, because this package is meant to be agnostic to cherrypy and used by other servers.

Can you add a changelog entry to CHANGES.txt explaining the change?

CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+v5.2.0
```
webknjaz: @jaraco shouldn't this be patch version bump?

Safihre:
:sleepy:

webknjaz:
@Safihre could you please resolve conflicts?

webknjaz:
I wish there were automated tests for this.
Thanks for contribution!

PR #8, opened 2017-02-25
PEP8 improvements

krysros:
Fixes some code style warnings / errors generated by flake8


cheroot/test/test_conn.py
```diff
@@ -8,8 +8,12 @@
 import six
 import pytest
 
-from cheroot._compat import HTTPConnection, HTTPSConnection, NotConnected, BadStatusLine
-from cheroot._compat import ntob, urlopen
+from cheroot._compat import HTTPConnection
```
webknjaz: What is the motivation for this change?

cheroot/test/test_core.py
```diff
@@ -316,25 +316,29 @@ def testSlashes(self):
         # Make sure GET params are preserved.
```
webknjaz: Changes of this module are potentially useless, since in contains tests related to cherrypy, which are likely to be removed

cheroot/test/test_conn.py
krysros: 1. Imports should usually [be on separate lines](https://www.python.org/dev/peps/pep-0008/#imports).
2. The length of the line 11 [exceeds 79 characters](https://www.python.org/dev/peps/pep-0008/#maximum-line-length).

cheroot/test/test_core.py
krysros: OK

cheroot/test/test_conn.py
webknjaz: From the link you provided:
> It's okay to say this though:
> 
> from subprocess import Popen, PIPE

so for `from ... import ...` it's OKay to list a few of them

they may be split into several lines using several `from ... import ...` statements or just multilined like
```python
from some_lib import (
    importable1, importable2,
    importable3, importable4,
    etc,
)
```

webknjaz: Regarding 79 chars limitation. Nowadays people usually set higher threshold, as we don't have 80 char terminals limitation today. Still it's a good practice to avoid long lines to make code better readable.

cheroot/wsgi.py
```diff
@@ -53,9 +53,11 @@ def __init__(self, bind_addr, wsgi_app, numthreads=10, server_name=None,
         self.request_queue_size = request_queue_size
         self.timeout = timeout
         self.shutdown_timeout = shutdown_timeout
-        self.requests = threadpool.ThreadPool(self, min=numthreads or 1, max=max,
+        self.requests = threadpool.ThreadPool(
+            self, min=numthreads or 1, max=max,
             accepted_queue_size=accepted_queue_size,
-            accepted_queue_timeout=accepted_queue_timeout)
+            accepted_queue_timeout=accepted_queue_timeout
+            )
```
webknjaz: I think this paren should be one tab left, according to PEP8

cheroot/test/test_conn.py
krysros: I agree.

PR #9, opened 2017-02-26
Fix import errors of test suite

rim99:


rim99:
Ignore the 2nd commit please. 
It may not be appropriate.
.

rim99:
The 1st commit CAN make `testFavicon` in `test_core.py` pass.

README.md
```diff
@@ -0,0 +1,3 @@
+## Branch: "cheroot-py3"
```
webknjaz: Please revert the recent commit

webknjaz:
> which cannot be deleted 

Why?

PR #11, opened 2017-02-26
Fix import errors

rim99:


cheroot/test/helper.py
```diff
@@ -40,7 +41,7 @@ class CherootWebCase(webtest.WebCase):
     scheme = 'http'
 
     available_servers = {
-        'wsgi': cheroot.server.WSGIServer,
+        'wsgi': cheroot.wsgi.WSGIServer,
```
webknjaz: I think it's `cheroot.wsgi.Server`

rim99: At the end of the `wsgi.py` file, `WSGIServer` class was defined as an alias of the `Server` class. So both are right.

```
# compatibility aliases
globals().update(
    WSGIServer=Server,
    WSGIGateway=Gateway,
    WSGIGateway_u0=Gateway_u0,
    WSGIGateway_10=Gateway_10,
    WSGIPathInfoDispatcher=PathInfoDispatcher,
)
```

webknjaz: Yes, but it seems those names are added for backwards compatibility and may be a subject of deprecation in future. So it's always better to use real names.

@jaraco am I missing anything?

rim99: Right, I get your point.

So should I just make a new commit on this one, or pull another request?

I think the latter is better?

webknjaz: When doing pull requests you may update it by pushing new commits into your branch or rewrite it with rebase and force push. There's no need to create new PR after each review :) 
Moreover, if you update commented lines, the related comments will automatically become collapsed

rim99: Yeah, I also feel weird for what I just did.

 I start to get familiar with GitHub quite recently. Thank you for all these hints. ^o^

PR #12, opened 2017-02-26
Set Server.version to Cheroot version instead of CherryPy version

### LINKED ISSUE #5: wsgi.Server reports CherryPy version
As I understand it, Cheroot brings a standalone server with seperate versioning (e.g. 5.1), but 
`cheroot.wsgi.Server.version` returns either "CherryPy/undefined" or "CherryPy/10.1.1", depending on the additional cherrypy package beeing installed or not.

mar10:
Fix #5

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -89,12 +89,6 @@ if 'win' in sys.platform and hasattr(socket, 'AF_INET6'):
         socket.IPV6_V6ONLY = 27
 
 
-try:
-    cp_version = pkg_resources.require('cherrypy')[0].version
-except Exception:
-    cp_version = 'unknown'
-
-
 if six.PY3:
     def ntob(n, encoding='ISO-8859-1'):
         """Return the given native string as a byte string in the given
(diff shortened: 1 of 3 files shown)


cheroot/server.py
```diff
@@ -1130,7 +1130,8 @@ class HTTPServer(object):
     timeout = 10
     """The timeout in seconds for accepted connections (default 10)."""
 
-    version = 'CherryPy/' + cp_version
+    # version = 'CherryPy/' + cp_version
+    version = 'Cheroot/' + __version__
```
webknjaz: I'm not sure about branding (whether it should begin with capital `C`, @jaraco should it?), but this looks good :)

mar10:
What would be a good place to add tests?
Cheroot status says
> the test suite is currently disabled. Currently, the only way to validate the test
suite is to run it against CherryPy.

I als could not get the CherryPy tests to work:
```
(cp35) $ python setup.py test
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
running test
(cp35) $ 
```

Running tox stops at this line
```
(cp35) $ tox
...
collecting 0 items
```
Maybe I am missing some dependencies?

webknjaz:
Oh.. that's a tricky question. I'll merge this w/o tests. They aren't fully set up yet.
But in case you'll want to add them in the future, they'll go to https://github.com/cherrypy/cheroot/tree/master/cheroot/test

jaraco:
Thanks both!

Issue #3, opened 2017-01-23
Port tests from CherryPy


jaraco:
Following the refresh of cheroot from CherryPy and the new release of Cheroot 5.0, the codebase is stable and suitable for release, but the test suite is [basically disabled](https://github.com/cherrypy/cheroot/blob/master/pytest.ini#L2).

We'll need to invest some time to adapt the test suite to work with the cheroot code and get the tests passing again, at which point most if not all of those tests can be removed from CherryPy, which can rely on the tests executed here.

PR #14, opened 2017-02-26
Auto create static files in test

rim99:


cheroot/test/test_core.py
```diff
@@ -717,6 +729,13 @@ def test_expose_decorator(self):
         self.assertStatus(200)
         self.assertBody('Mr. and Mrs. Watson')
 
+    @classmethod
+    def teardown_class(cls):
+        cls.supervisor.stop()
+        # remove the temp static files
+        for p in [favicon_ico_file, index_html_file]:
```
webknjaz: It's not a good idea to create files globally in module outside of the class and remove it in testcase class.

cheroot/test/test_core.py
```diff
@@ -16,8 +17,12 @@
 
 
 localDir = os.path.dirname(__file__)
-favicon_path = os.path.join(os.getcwd(), localDir, '../favicon.ico')
```
webknjaz: You are mixing `os.path` methods in, while having imported pathlib you could've reused its features of path construction

cheroot/test/test_core.py
```diff
@@ -19,10 +19,8 @@
 localDir = os.path.dirname(__file__)
 base_path = os.path.join(os.getcwd(), localDir)
 favicon_path = os.path.join(base_path, '../favicon.ico')
-
-favicon_ico_file = pathlib.Path(os.path.join(base_path, '../favicon.ico'))
-static_dir = pathlib.Path(os.path.join(base_path, 'static/'))
-index_html_file = pathlib.Path(os.path.join(base_path, 'static/index.html'))
+static_dir = os.path.join(base_path, 'static/')
+index_html_path = os.path.join(base_path, 'static/index.html')
```
webknjaz: I don't think it's a good place for having test_method related file declared. Please move it to the corresponding test.

cheroot/test/test_core.py
```diff
@@ -512,70 +514,76 @@ def testFlatten(self):
...
-            # "When this status code is returned for a byte-range request,
-            # the response SHOULD include a Content-Range entity-header
-            # field specifying the current length of the selected resource"
-            self.assertHeader('Content-Range', 'bytes */14')
-        elif cherrypy.server.protocol_version == 'HTTP/1.0':
-            # Test Range behavior with HTTP/1.0 request
-            self.getPage('/ranges/slice_file', [('Range', 'bytes=2-5')])
-            self.assertStatus(200)
-            self.assertBody('Hello, world\r\n')
+        os.mkdir(static_dir)
```
webknjaz: It's not appropriate think to do. You may create static dir on class setup, but normally it should just exist in the repo already.

cheroot/test/test_core.py
rim99: But at the line 34, `favicon_path` was referenced when set up the server. 
So either put `favicon_path` separately in different places, or remain being a global const.

Only there two variables as shown above can be moved to the test.

cheroot/test/test_core.py
```diff
@@ -19,15 +19,18 @@
...
 #                             Client-side code                             #
 
 
 class CoreRequestHandlingTest(helper.CPWebCase):
 
     @staticmethod
     def setup_server():
+
+        static_dir = os.path.join(base_path, 'static/')
+        if not os.path.exists(static_dir):
```
webknjaz: This is not needed, create a dir in git instead

cheroot/test/test_core.py
```diff
@@ -19,15 +19,18 @@
...
-index_html_path = os.path.join(base_path, 'static/index.html')
 #                             Client-side code                             #
 
 
 class CoreRequestHandlingTest(helper.CPWebCase):
 
     @staticmethod
     def setup_server():
+
+        static_dir = os.path.join(base_path, 'static/')
```
webknjaz: You may define `static_dir` in module scope

cheroot/test/test_core.py
```diff
@@ -280,6 +286,17 @@ def commas(self):
 
         cherrypy.tree.mount(root)
 
+    @staticmethod
+    @contextlib.contextmanager
+    def temp_file(file, content):
```
webknjaz: This method is completely detached from the class, I think it is better to turn it into a function in module level scope.

cheroot/test/test_core.py
```diff
@@ -500,70 +517,73 @@ def testFlatten(self):
...
 
     def testFavicon(self):
-        # favicon.ico is served by staticfile.
-        icofilename = os.path.join(localDir, '../favicon.ico')
-        icofile = open(icofilename, 'rb')
-        data = icofile.read()
-        icofile.close()
+        with self.temp_file(favicon_path, 'Test file'):
+            # favicon.ico is served by staticfile.
+            icofilename = os.path.join(localDir, '../favicon.ico')
```
webknjaz: This path is already stored in the `favicon_path` variable.

cheroot/test/test_core.py
```diff
@@ -500,70 +517,73 @@ def testFlatten(self):
...
+            icofile = open(icofilename, 'rb')
```
webknjaz: Could you please work with file using a context manager? (`with` syntax)

cheroot/test/test_core.py
```diff
@@ -17,20 +17,24 @@
 
 
 localDir = os.path.dirname(__file__)
-base_path = os.path.join(os.getcwd(), localDir)
-favicon_path = os.path.join(base_path, '../favicon.ico')
+temp_dir = os.path.join(os.getcwd(), localDir, 'temp')
```
webknjaz: Why did you rename it? It's used to be `static` since times of being in `cherrypy` repo's tests. I believe there are other tests which still rely on `static` folder.

rim99: Well I think `temp` is a more suitable name consider its usage. And I try run `pytest` on the whole test suite, which turns out to be all passed.

```
➜  test git:(cheroot-fixbug-auto-create-static-file) ✗ pytest
================= test session starts ===========================
platform darwin -- Python 3.5.1, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /Users/rim99/PycharmProjects/baige, inifile: pytest.ini
collected 56 items 

test_compat.py s.
test_config_server.py ssss
test_conn.py sssssssssssssssss
test_core.py s..................
test_http.py .s.......
test_wsgiapps.py .....

============ 32 passed, 24 skipped in 92.70 seconds ===============

```

webknjaz:
We need to have testing configured properly before proceeding w/ merging this.

rim99:
Changes needed?

webknjaz:
`.travis.yaml` should be properly configured and it must be a separate PR. I'll look into the implementation a bit later. Travis CI is currently not fully functioning due to yesterday's issues at Amazon.

webknjaz:
@rim99 please, resolve conflicts

webknjaz:
It looks like @jaraco cleaned up this test contents in scope of #3. So this PR isn't relevant anymore.

Issue #17, opened 2017-03-07
New option to drop headers that contain underscores


mar10:
I'd like to propose a new option to drop headers that contain underscores ('_').

Cheroot / CherryPy convert HTTP headers for WSGI to upper case and replace hyphen ('-') with  underscore ('_'):

```py
env.update(
    ('HTTP_' + bton(k).upper().replace('-', '_'), bton(v))
    for k, v in req.inheaders.items()
)
```   

So this headers

```py
headers = {"Sm-User": "val1",
           "Sm_User": "val2",}
res = requests.get("http://127.0.0.1:8080", headers=headers)
```

will result in a **single** WSGI environment entry `"HTTP_SM_USER": "val1"` *or* `"HTTP_SM_USER": "val2".` 
(The concrete behavior is undefined, since the order of dict key iteration is hard to control.)

### Proposed solution

This *could* be fixed by modifying the wsgi server code above, but I would suggest to drop those headers in the core code.

While underscores in header fields are allowed ([rfc7230](https://tools.ietf.org/html/rfc7230#section-3.2)), they are uncommon, AFAIK.  
For example Apache and nginx seem to drop them by default:
http://stackoverflow.com/a/18205049/19166
http://stackoverflow.com/a/22856867/19166

So a solution could look like this:
```py
def read_headers(rfile, hdict=None, drop_underscores=False):
    ...
    while True:
        line = rfile.readline()
            ...
            if drop_underscores and '_' in k:
                continue
            k = k.strip().title()
            v = v.strip()
            hname = k
            ...
```

I tried to prepare a PR, but I could need some advice how and where to make this configurable:

```py
        try:
            read_headers(self.rfile, self.inheaders)
            # Configurable using "`server.underscores_in_headers": False`
            drop_underscores = not self.server.server_adapter.underscores_in_headers
            read_headers(self.rfile, self.inheaders, drop_underscores)
        except ValueError:
            ex = sys.exc_info()[1]
            self.simple_response('400 Bad Request', ex.args[0])
```

For example I tried adding this to cheroot.HTTPServer:

```py
class HTTPServer(object):
    ...
    underscores_in_headers = True
    """If False, silently discards headers containing underscores ('_')."""
```

and then make it configurable from cherrypy like so:

```py
cherrypy.config.update({"server.underscores_in_headers": False})
cherrypy.quickstart(HelloWorld())
``` 
=> this seems to set the attribute on the ServerAdapter instead of the HttpServer.

or 

```py
[...]

PR #18, opened 2017-03-09
Drop headers containing underscores

mar10:
Update #17

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -142,7 +142,7 @@ if not hasattr(logging, 'statistics'):
     logging.statistics = {}
 
 
-def read_headers(rfile, hdict=None):
+def read_headers(rfile, hdict=None, drop_underscores=False):
     """Read headers from the given stream into the given header dict.
 
     If hdict is None, a new header dict is created. Returns the populated
@@ -177,6 +177,8 @@ def read_headers(rfile, hdict=None):
                 k, v = line.split(COLON, 1)
             except ValueError:
                 raise ValueError('Illegal header line.')
+            if drop_underscores and '_' in k:
+                continue
             # TODO: what about TE and WWW-Authenticate?
             k = k.strip().title()
             v = v.strip()
@@ -654,7 +656,10 @@ class HTTPRequest(object):
 
         # then all the http headers
         try:
-            read_headers(self.rfile, self.inheaders)
+            # Configurable using "`server.drop_underscore_headers": True`:
+#             drop_underscores = self.server.server_adapter.drop_underscore_headers
+            drop_underscores = self.server.drop_underscore_headers
+            read_headers(self.rfile, self.inheaders, drop_underscores)
         except ValueError:
             ex = sys.exc_info()[1]
             self.simple_response('400 Bad Request', ex.args[0])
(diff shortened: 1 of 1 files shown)


cheroot/server.py
```diff
@@ -654,7 +656,10 @@ def read_request_headers(self):
 
         # then all the http headers
         try:
-            read_headers(self.rfile, self.inheaders)
+            # Configurable using "`server.drop_underscore_headers": True`:
+#             drop_underscores = self.server.server_adapter.drop_underscore_headers
```
webknjaz: I assume the `#` at the beginning of this line is there by mistake

cheroot/server.py
```diff
@@ -654,7 +656,10 @@ def read_request_headers(self):
...
+            drop_underscores = self.server.drop_underscore_headers
```
webknjaz: I'd use `getattr()` instead to safely fallback to feature disabled state.

cheroot/server.py
```diff
@@ -1146,6 +1151,10 @@ class HTTPServer(object):
     nodelay = True
     """If True (the default since 3.1), sets the TCP_NODELAY socket option."""
 
+    # issue #17
```
webknjaz: please add such comments in the end of the line they refer to in order to avoid lines decoupling over time

cheroot/server.py
webknjaz: (Are there such cases?)

cheroot/server.py
mar10: No, I left this comment for discussion, since I noticed that this code
```py
    cherrypy.config.update({"server.drop_underscore_headers": True})
    cherrypy.quickstart(HelloWorld())
```
modified the attribute on `self.server.server_adapter.drop_underscore_headers` (not `self.server.server_adapter.drop_underscore_headers`).
So I am not sure if we should declare this option on the adapter instead of the server object.

cheroot/server.py
mar10: will do.

cheroot/server.py
```diff
@@ -652,7 +654,10 @@ def read_request_headers(self):
 
         # then all the http headers
         try:
-            read_headers(self.rfile, self.inheaders)
+            # Configurable using `"server.drop_underscore_headers": True`:
+#             drop_underscores = getattr(self.server.server_adapter, 'drop_underscore_headers', False)
```
webknjaz: I think this comment shouldn't go into merge.
Btw, what is the difference with `server_adapter` version?

mar10: No this should not be merged. I just proposed two variants for discussion. I'll add a comment to the issue #17

cheroot/server.py
```diff
@@ -1144,6 +1149,9 @@ class HTTPServer(object):
     nodelay = True
     """If True (the default since 3.1), sets the TCP_NODELAY socket option."""
 
+    drop_underscore_headers = False  # issue #17
```
jaraco: In general, I try to avoid solutions that involve boolean flags, especially if those flags trickle through function call parameters.

What I'd rather see here is that instead of having a global `read_headers` function, that instead we consider creating a HeaderReader interface (or similar), and a given cheroot.server.Server can be configured with an instance of a HeaderReader (or some reference to one). In this way, cheroot provides an interface that allows this customization to be made or other customizations to be made without each one needing a boolean switch. I'll draft what I mean.

jaraco:
I've pushed a draft to [this branch](/cherrypy/cheroot/tree/feature/header-reader) demonstrating what I have in mind by the comment above. I'm not entirely happy with it as it is, but it demonstrates the concept... that all of the details about how to customize header reading are contained in one class that can be replaced by other classes.

jaraco:
I'll send a PR with my draft for comments.

PR #22, opened 2017-03-16
add attribute "ciphers" to ssladapter

LenovoCindywang:
Hi Team,
Our security team got a request recently, need to dynamiclly set cipher suites of https connections.
Since Python(version >= 2.7.9)support "context.set_ciphers()", then we can add attribute "ssl_ciphers", and transfer it to "ssl_builtin.py". Thus, if cherrypy use builtin ssl lib, it can dynamicly configure the cipher suites.

I have a solution to add the attribute, could you please merge it once you pass the review?

Thanks,
Cindy

### DIFF
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -59,7 +59,7 @@ class BuiltinSSLAdapter(Adapter):
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
-            if(self.ciphers != None):
+            if self.ciphers is not None:
                 self.context.set_ciphers(ciphers)
 
     def bind(self, sock):
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -50,10 +50,9 @@ class BuiltinSSLAdapter(Adapter):
     def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None):
         if ssl is None:
             raise ImportError('You must install the ssl module to use HTTPS.')
-        self.certificate = certificate
-        self.private_key = private_key
-        self.certificate_chain = certificate_chain
-        self.ciphers=None
+
+        super(BuiltinSSLAdapter, self).__init__(certificate, private_key, certificate_chain, ciphers)
+
         if hasattr(ssl, 'create_default_context'):
             self.context = ssl.create_default_context(
                 purpose=ssl.Purpose.CLIENT_AUTH,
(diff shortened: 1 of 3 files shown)


cheroot/ssl/builtin.py
```diff
@@ -43,20 +43,25 @@ class BuiltinSSLAdapter(Adapter):
...
         self.private_key = private_key
         self.certificate_chain = certificate_chain
+        self.ciphers=None
         if hasattr(ssl, 'create_default_context'):
             self.context = ssl.create_default_context(
                 purpose=ssl.Purpose.CLIENT_AUTH,
                 cafile=certificate_chain
             )
             self.context.load_cert_chain(certificate, private_key)
+            if(self.ciphers != None):
```
webknjaz: `if self.ciphers is not None:`

cheroot/ssl/builtin.py
```diff
@@ -43,20 +43,25 @@ class BuiltinSSLAdapter(Adapter):
...
     context = None
 
-    def __init__(self, certificate, private_key, certificate_chain=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None):
         if ssl is None:
             raise ImportError('You must install the ssl module to use HTTPS.')
         self.certificate = certificate
         self.private_key = private_key
         self.certificate_chain = certificate_chain
+        self.ciphers=None
```
webknjaz: `self.ciphers = None`

webknjaz: self.ciphers = ciphers

cheroot/ssl/builtin.py
```diff
@@ -43,20 +43,25 @@ class BuiltinSSLAdapter(Adapter):
     """The ssl.SSLContext that will be used to wrap sockets where available
     (on Python > 2.7.9 / 3.3)
     """
+    ciphers = None
+    """The ciphers list of SSL."""
```
The-Compiler: I've never seen a docstring used to document variables FWIW

webknjaz: https://www.python.org/dev/peps/pep-0258/#attribute-docstrings

The-Compiler: Note that pep was rejected though :wink:

webknjaz: At least this is supported by linters. Perhaps @jaraco could link smth relevant.

webknjaz: The active PEP257 still [has attribute docstrings mentions](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)

jaraco:
Released as 5.4.0.

cheroot/ssl/builtin.py
jaraco: Yes, attribute docstrings exist. I endorse them despite the modest adoption in the wild.

PR #25, opened 2017-04-23
HeaderReader class to replace read_headers

jaraco:
Based on #18 and addressing #17, this approach presents an mechanism by which headers may be filtered or otherwise manipulated as read from the request.

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -132,6 +132,10 @@ if not hasattr(logging, 'statistics'):
 
 
 class HeaderReader(object):
+    """
+    Interface and default implementation of an object for reading
+    headers from an HTTP request.
+    """
 
     def __call__(self, rfile, hdict=None):
         """
@@ -193,6 +197,11 @@ class HeaderReader(object):
 
 
 class DropUnderscoreHeaderReader(HeaderReader):
+    """
+    Custom HeaderReader to exclude any headers with underscores
+    in them.
+    """
+
     def _allow_header(self, key_name):
         orig = super(DropUnderscoreHeaderReader, self)._allow_header(key_name)
         return orig and '_' not in key_name
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -192,6 +192,12 @@ class HeaderReader(object):
         return key_name.strip().title()
 
 
+class DropUnderscoreHeaderReader(HeaderReader):
+    def _allow_header(self, key_name):
+        orig = super(DropUnderscoreHeaderReader, self)._allow_header(key_name)
+        return orig and '_' not in key_name
+
+
 class SizeCheckWrapper(object):
     """Wraps a file-like object, raising MaxSizeExceeded if too large."""
 
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -173,6 +173,9 @@ class HeaderReader(object):
                 k = self._transform_key(k)
                 hname = k
 
+            if not self._allow_header(k):
+                continue
+
             if k in comma_separated_headers:
                 existing = hdict.get(hname)
                 if existing:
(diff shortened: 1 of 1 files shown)


cheroot/server.py
```diff
@@ -131,53 +131,71 @@ def bton(b, encoding='ISO-8859-1'):
...
+                existing = hdict.get(hname)
+                if existing:
+                    v = b', '.join((existing, v))
+            hdict[hname] = v
 
-    return hdict
+        return hdict
+
+    def _allow_header(self, key_name):
+        return key_name
```
webknjaz: `return True`?

cheroot/server.py
```diff
@@ -131,53 +131,71 @@ def bton(b, encoding='ISO-8859-1'):
...
+            else:
+                try:
+                    k, v = line.split(COLON, 1)
+                except ValueError:
+                    raise ValueError('Illegal header line.')
+                v = v.strip()
+                k = self._transform_key(k)
+                hname = k
+
+            if not self._allow_header(k):
```
webknjaz: It looks you've lost one level of indentation starting here till line 183.

webknjaz: Also, what if one wants to run a check before transformation?

cheroot/server.py
```diff
@@ -131,53 +131,71 @@ def bton(b, encoding='ISO-8859-1'):
     logging.statistics = {}
 
 
-def read_headers(rfile, hdict=None):
-    """Read headers from the given stream into the given header dict.
+class HeaderReader(object):
```
webknjaz: Please add a docstring here.

cheroot/server.py
jaraco: Hmm. Perhaps. Unless we want to filter out empty headers. I guess the previous implementation allowed for empty headers, so this refactor should as well.

cheroot/server.py
jaraco: If I add a level of indentation, that would put all of that code within the else block. That code wasn't in the else block before. Why should it be now?

jaraco: If one wants to run a check before the transformation, they will either need to override _transform_key to perform the check there, or they should describe why that feature would be useful and add an appropriate hook to enable it.

webknjaz: @jaraco oh.. I assumed it was.

`k` variable is created in `else`-block, meaning we'll get `NameError` when `if` part happens, while referencing a variable before assignment. Am I overlooking smth?

cheroot/server.py
```diff
@@ -131,53 +131,71 @@ def bton(b, encoding='ISO-8859-1'):
...
+                # It's a continuation line.
+                v = line.strip()
+            else:
+                try:
+                    k, v = line.split(COLON, 1)
+                except ValueError:
+                    raise ValueError('Illegal header line.')
+                v = v.strip()
+                k = self._transform_key(k)
+                hname = k
```
webknjaz: why not assign `hname = self._transform_key(k)` and use just this var name? do we need two vars for the same purpose?

cheroot/server.py
jaraco: Yes, the code is complex. The other part of the else block (the if block) is indicating a 'continuation line' which is assuming k was set in a previous line (previous iteration of the while loop). So yes, you would get a name error if the first line contained a space or tab, but apparently that's not an issue in practice, as the original code used the same logic.

cheroot/server.py
jaraco: Maybe, although in a loopy/branchy code like this, I've decided to limit my changes to the refactoring in order to avoid potential issues with other optimizations. Since the previous code set k and hname, I decided to retain that. We can optimize that away later if appropriate.

cheroot/server.py
webknjaz: got it, thanks :)

PR #26, opened 2017-04-28
Run Travis CI tests in OS X based environment

webknjaz:
* PoC of running tests under MacOS
* reusable YAML anchors reduce config duplication dramatically
* completely pyenv based python installations


webknjaz:
@jaraco I've found a working way to run tests under OS X in TravisCI. Should we run them across all envs or maybe just a few?

.travis.yml
```diff
@@ -47,26 +49,97 @@ _base_envs:
...
+    - PYTHON_VERSION=3.7-dev
+    - *env_pyenv
+    - *env_path
+  - <<: *osx_python_base
+    python: pypy2-5.7.1
+    env:
+    - PYTHON_VERSION=pypy2-5.7.1
+    - *env_pyenv
+    - *env_path
+  # Fails under OS X:
```
jaraco: I generally try to avoid commented code. I suggest either marking this test as an allowed failure (if you intend for it to work in the near future) or just remove it (and let the source code history remember it in case we choose to revive it).

jaraco:
When you merge this PR, please also cut a new release. Thanks.

webknjaz:
Actually I was worried about the feedback time period from CI: it largely increases waiting time of "Okay, tests are green" or "Oops.. Smth is wrong with your changes in specific env". That's my primary motivation of asking about leaving just a few OS X tests.

Regarding "the latest 2.7" (etc.) I've tried that, but pyenv has it just for a few envs, so I wanted to be consistent across config, given that I follow declarative manner of specifying versions (although I could actually script version lookup and match via `pyenv install --list`).

PR #29, opened 2017-05-19
Change Travis CI config to use newly presented build stages feature

### LINKED ISSUE #28: [TODO] Change Travis CI config to use newly presented build stages feature
Refs:
* https://blog.travis-ci.com/2017-05-11-introducing-build-stages
* https://docs.travis-ci.com/user/build-stages#Examples

*Note: I'm going to move PYPI deployment to separate stage, so it would wait for tests in all envs to complete.*

webknjaz:
Closes #28.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ _base_envs:
       - xz-utils
       - tk-dev
 - &pyenv_base
+  stage: test
   python: pypy
   env:
   - PYTHON_VERSION=pypy2-5.7.1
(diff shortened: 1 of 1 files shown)


webknjaz:
[postponed]

Ref bug: https://github.com/travis-ci/beta-features/issues/11#issuecomment-302766298

webknjaz:
Ref bug: travis-ci/travis-ci#7789

webknjaz:
That bug has been fixed.

webknjaz:
@jaraco what about moving `TOXENV=pre-commit` job to separate pre-flight check stage? It runs faster than others and will fail the build faster, preventing waste on resources for more complex tests, which won't run at all if linters fail?

.travis.yml
```diff
@@ -140,6 +140,20 @@ matrix:
...
     after_failure: skip
+  - stage: upload new version of python package to PYPI (only for tagged commits)
+    python: 3.6
+    deploy:
+      provider: pypi
+      server: https://upload.pypi.org/legacy/
+      on:
+        tags: true
+        all_branches: true
+        python: 3.6
```
jaraco: Is it possible to eliminate this selection characteristic now?

webknjaz: Just remove it? Technically, yes. But I would leave it here for safety. We could use YAML anchors to reduce version duplication.

PR #30, opened 2017-06-12
Avoid infinite loop while reading file from request body

mewalig:


webknjaz:
Ref: cherrypy/cherrypy#1602

cheroot/server.py
```diff
@@ -387,6 +387,10 @@ def _fetch(self):
 
     def read(self, size=None):
         data = EMPTY
+        
+        if size == 0:
+            return data
+        
         while True:
             if size and len(data) >= size:
```
webknjaz: Perhaps, it would be better to add the above check here.

PR #32, opened 2017-06-18
Add more built-in SSL errors to ignore

Safihre:
Users have reported these 2 new errors that get thrown specifically by the built-in SSL provider compared to PyOpenSSL.

* `unknown error`: If testing a CherryPy server using https://testssl.sh/: https://github.com/sabnzbd/sabnzbd/issues/820#issuecomment-281463671
* `https proxy request`: When someone tries to use a CherryPy server as a Proxy server: https://github.com/sabnzbd/sabnzbd/issues/860


webknjaz:
Released as v5.5.2

PR #33, opened 2017-06-18
Default HTTP Server header to Cheroot version str

webknjaz:
I don't see the need to use hostname as default `Server` header: it's designed to advertising software running HTTP server (i.e. cherrypy/cheroot).


webknjaz:
@jaraco?

webknjaz:
I've decided to give it a try, since it's not something breaking the flow. Users still can explicitly define the server name and if they want, it's fairly easy to pass `server_name` argument.

PR #37, opened 2017-06-25
wsgi: add peer UNIX credential to a non-standard environment variable

bazsi:
Signed-off-by: Balazs Scheidler <balazs.scheidler@balabit.com>

It adds the UNIX UID of a unix domain socket that connects to our HTTP service over socket.

none

peer credentials are not available to apps even if they bind to a unix domain socket 

I've added a non-standard _REMOTE_UID value to environ, that is available to apps. I've originally did this patch to an older version of cherrypy and this is a forward port.

* **Other information**:


bazsi:
I've attempted to fix automatic tests, but the ones in circle ci seems to be an infrastructure problem. The rest of the tests were green.

I'd be really grateful about the feature, especially the naming of the environment variable. I'd like to deploy something like this in production, and I'd like to use a name that will match the 'official' one.

thanks.

bazsi:
Any comments?

webknjaz:
Oh, sorry @bazsi.. I was busy and didn't check notifications during past few days.

Nevermind about CircleCI, it's not configured yet.
Regarding your change, it's on my list to check it. I'll do this ASAP once I have some free time.

cheroot/wsgi.py
```diff
@@ -276,6 +278,11 @@ def get_environ(self):
 
         return env
 
+    def _get_peer_uid(self):
+        creds = self.req.server.socket.getsockopt(socket.SOL_SOCKET, socket.SO_PEERCRED, struct.calcsize('3i'))
```
webknjaz: There's no `SO_PEERCRED` in Python 2: https://utcc.utoronto.ca/~cks/space/blog/python/AbstractUnixSocketsAndPeercred

cheroot/wsgi.py
```diff
@@ -276,6 +278,11 @@ def get_environ(self):
 
         return env
 
+    def _get_peer_uid(self):
```
webknjaz: It isn't obvious what this function purpose is, so it's a good practice to leave comments + docstring.

cheroot/wsgi.py
```diff
@@ -254,6 +255,7 @@ def get_environ(self):
             # AF_UNIX. This isn't really allowed by WSGI, which doesn't
             # address unix domain sockets. But it's better than nothing.
             env['SERVER_PORT'] = ''
+            env['_REMOTE_UID'] = self._get_peer_uid()
```
webknjaz: Why do you need this as an environment variable?

cheroot/wsgi.py
webknjaz: Can this be moved to the request entity?

bazsi: Done in 831ec2db35533bed15b2d9bf38894ccbaab2af00

cheroot/wsgi.py
bazsi: also done in 831ec2db35533bed15b2d9bf38894ccbaab2af00

I would be squashing that into the base commit as review is complete.

cheroot/wsgi.py
bazsi: I felt this was a lot more connected to REMOTE_ADDR and the likes, as I've felt it is a property of the connection rather than the request itself.

In our internal HTTP based service, we are using the environment to populate our internal data structures and by that time the connection itself is not available.

If you meant that _REMOTE_UID stays in the environment, but the implementation itself moves to HTTPConnection, that could make sense. Is that what you are thinking?

Thanks for the feedback.

bazsi:
I have now pushed a couple of fixups to address your notes. If you are ok with them, I would rebase the branch and squash them to the same patch.

cheroot/wsgi.py
```diff
@@ -26,8 +26,9 @@ def my_crazy_app(environ, start_response):
 """
 
 import sys
-
+import struct
```
webknjaz: It's unused in this module

cheroot/wsgi.py
```diff
@@ -254,6 +255,9 @@ def get_environ(self):
             # AF_UNIX. This isn't really allowed by WSGI, which doesn't
             # address unix domain sockets. But it's better than nothing.
             env['SERVER_PORT'] = ''
+            uid = self.req.conn.get_peer_uid()
+            if uid is not None:
+                env['_REMOTE_UID'] = str(uid)
```
webknjaz: Are there any other WSGI implementations providing similar env var?

cheroot/wsgi.py
```diff
@@ -26,8 +26,9 @@ def my_crazy_app(environ, start_response):
 """
 
 import sys
-
+import struct
 import six
+import socket
```
webknjaz: Unused as well

cheroot/server.py
```diff
@@ -1155,6 +1156,28 @@ def close(self):
...
 
+    def get_peer_uid(self):
+        """Returns the UID of the peer socket in the case of UNIX domain sockets
+
+        This function uses SO_PEERCRED to query the UNIX UID of the peer, which
+        is only available if the bind address is a UNIX domain socket.
+        """
+
+        if self.socket.family != socket.AF_UNIX:
+            return None
```
webknjaz: I would prefer raising exception rather than returning "magic values"

cheroot/server.py
```diff
@@ -1155,6 +1156,28 @@ def close(self):
...
+        # NOTE: the value for SO_PEERCRED can be architecture specific, in
+        # which case the getsockopt() will hopefully fail.  The arch
+        # specific value could be derived from platform.processor()
+        SO_PEERCRED = getattr(socket, 'SO_PEERCRED', 17)
+        try:
+            creds = self.socket.getsockopt(socket.SOL_SOCKET, SO_PEERCRED, struct.calcsize('3i'))
+            pid, uid, gid = struct.unpack('3i', creds)
+            return uid
+        except socket.error:
+            pass
```
webknjaz: Please, don't swallow exception. Log it and re-raise appropriate one.

cheroot/server.py
```diff
@@ -1155,6 +1156,28 @@ def close(self):
             # Apache does, but not today.
             pass
 
+    def get_peer_uid(self):
```
webknjaz: I'd prefer a method returning the whole tuple of creds and 3 `@property` methods returning each of 'em.

cheroot/server.py
```diff
@@ -1155,6 +1156,28 @@ def close(self):
...
+        is only available if the bind address is a UNIX domain socket.
+        """
+
+        if self.socket.family != socket.AF_UNIX:
+            return None
+
+        # NOTE: the value for SO_PEERCRED can be architecture specific, in
+        # which case the getsockopt() will hopefully fail.  The arch
+        # specific value could be derived from platform.processor()
+        SO_PEERCRED = getattr(socket, 'SO_PEERCRED', 17)
```
webknjaz: I'd add this as a monkey-patch next to similar win-specific code @ line 80 to be consistent

cheroot/wsgi.py
webknjaz: Could you please share your snippet where you cannot access HTTPConnection and still can access env vars? I'm really not sure whether all other users need this as it adds extra syscall overhead.

webknjaz:
@bazsi it would be very helpful to have tests for this feature as well

cheroot/wsgi.py
bazsi: I've found a number of WSGI/HTTP servers/libs that support unix domain sockets:
* https://github.com/eventlet/eventlet/pull/320/files
* nginx with its proxy_pass directive
* https://github.com/jakubroztocil/httpie/issues/209

With that said, the only HTTP over UNIX domain socket implementation, that I've found to  also support SO_PEERCRED based authentication is cups, which has its embedded http implementation (in C, brr) and does not use CGI like environment variables.

I've also looked up rfc3875, the original specification of CGI, which seems to be the origin for these environment variables: http://www.ietf.org/rfc/rfc3875. In the RFC, there's a standard REMOTE_USER environment variable, which seems to be what we aim for, but would require username lookup instead of just passing on the UID. username resolution could be expensive, especially as it would need to be done on every request.

That same RFC, also specifies that non-standard extensions should be prefixed with an X_.

With all this said, I think:
 * unix domain based authentication makes sense, there is at least one implementation, and a use-case that matches mine.
 * environment wise, we have three options:
    1) use X_REMOTE_UID and pass the uid only
    2) resolve the uid to username (with caching?), and reuse the standard REMOTE_USER attribute
    3) disregard the CGI origin, and use something different.

I've also found this document about WSGI: http://wsgi.readthedocs.io/en/latest/specifications/simple_authentication.html. This says, that the username needs to go to REMOTE_USER, thus

My opinion and the above spec point to option number 2) above.

If you agree, I can change my patches this way.

cheroot/server.py
bazsi: I actually wanted to ignore errors and in those cases not add this to the environment. It seems to be a Linux specific feature. I wouldn't want UNIX domain sockets not to work if this is not available.

perhaps a one-time warning would suffice?

bazsi:
I'll do another iteration and push the updated patches.

cheroot/server.py
webknjaz: Alright, warning would be ok (mabe even  `NOTICE`). But I still prefer raising exception instead of returning `None`.

cheroot/wsgi.py
webknjaz: 1) This shouldn't be a default behavior to avoid overhead for other users who may not want such logic to be triggered, so I would prefer that the calling logic go to CherryPy tool, which would be easy to turn on or switch off easily via config option.
2) It would be enough to provide `@properties`, which evaluate creads lookup lazily, once per request as I commented earlier.
3) It would be ok to have some function/method populating all of `REMOTE_USER`, `X_REMOTE_UID`, `X_REMOTE_GID`, `X_REMOTE_PID` or just some of them as configured.

bazsi: hmm.. cherrypy does not really have access to HTTPConnection objects, so I don't see how that would go back to the socket to do SO_PEERCRED. This is how it looks like (obviously I don't have intimate knowledge of these parts, I just did a quick browse of the related source code):

 * cherrypy instantiates a cheroot.wsgi.Server instance (in ServerAdapter). wsgi.Server is derived from the "plain" HTTP server, cheroot.server.HTTPServer
 * as a connection arrives, a HTTPConnection is instantiated to encapsulate the connection specific socket
 * we parse the incoming request, and a cheroot specific HTTPRequest is created, to encapsulate the low-level HTTP details. Here we still have a reference to the connection/server objects.
 * then a cheroot.wsgi.Gateway is created, one that takes the low-level HTTPRequest, and 
populates a WSGI environment, and passes it to the the WSGI application via a __call__ method. 
 * The environment is the only input to the WSGI app, even headers are translated into and from environment name-value pairs. (with ugly translations like Accept-Language becoming HTTP_ACCEPT_LANGUAGE)

There seems to be an intention to decouple the HTTP specific pieces and the app side, and the means to exchange information is the environment itself. It is thus not possible to go back from the app to HTTPConnection & HTTPRequest where the information is available.

The Native interface is one abstraction short (e.g. environment is not created and HTTPRequest is directly translated to cherrypy request objects), but that is not the default how cherrypy operates. That interface does not support authentication (e.g. request.login is not populated at all).

I could inject the socket into the environment (with a key like "wsgi.socket" to match other attributes), but my opinion is that this would be ugly and would derail the entire idea to decouple apps from HTTP specifics.

Short of adding this ugly hack to environment, we need to do the authentication on the cheroot side and pass the information via environment variables.

Still, REMOTE_USER looks to be the best solution, as that would work out of the box with cherrypy (e.g. request.login is filled automatically).

I could make the username resolution stuff optional by adding an additional constructor parameter to cheroot.wsgi.Server that could be driven by a cherrypy config variable (e.g. server.socket_file_auth True/False).

I can also amortize the performance impact by caching the username lookup, because in most cases, it would only be a _very_ limited number of users that would use the interface.

I can see these courses of action at this point if we are to use REMOTE_USER:
1) additional parameter w/o caching
2) additional parameter w/ caching
3) no additional parameter w/ caching

OR we can use X_REMOTE_UID, GID and PID as you suggested, in which case username resolution is not needed at all, so performance impact should be small. But in that case cherrypy.request.login will not be set automatically.

What do you think?

bazsi:
I still have not progressed with this. coming back once I have a timeslice.

webknjaz:
No worries. I'll be waiting for updates from your side and maybe even push a few commits to your branch if I have some free time for this.

webknjaz:
@bazsi I've rebased your branch on top of current master

webknjaz:
@bazsi I've added bits of refactoring, but it's not considered as complete yet.

I'd like to address:
- [ ] https://github.com/cherrypy/cheroot/pull/37#pullrequestreview-46783369 discussion
- [ ] play around with abstract sockets

webknjaz:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-0153c32f6bf970027939776a00602844c9992d1e)

**new alerts:**

* 1 for Unnecessary pass

---

webknjaz:
I still hope to find some time for this eventionally.

webknjaz:
unstale

webknjaz:
unstale

webknjaz:
Hi @bazsi!

Sorry for being silent for so long. I think I came up with a proper solution now.
I've reconsidered feature switching problem a bit: now I think it should be server-wide. This way it's easy to make it switchable in CherryPy as well (I've tested it). Also, it doesn't make sense on per-request basis or per-URI, because it operates on the lower level and has nothing to do with those.

It's bypassed through an adapter in CherryPy to Server instance in cheroot and then it pre-populates it into connections. WSGI env tries to get this data from connection and it succeeds when enabled and supported by the runtime.

Option in CherryPy would be `server.peercreds_enabled` in global section. I'm also thinking about additional option to enable resolving PID/GID from this auth data into string names.

Do you mind taking another round of testing and finish this with me?

webknjaz:
@bazsi I think, PoC is ready. 

Additionally to https://github.com/cherrypy/cheroot/pull/37#pullrequestreview-46783369 I'm thinking that we could also looking at [`REMOTE_IDENT` (RFC 3875 section 4.1.10., RFC 1413)](http://www.ietf.org/rfc/rfc3875)

I'm almost ready to merge this, but it would be great to see some tests around it.

webknjaz:
@bazsi 
Our test client doesn't support making HTTP requests over UNIX socket, so I'll have to postpone covering this with tests and doing any other modifications in the code base withing this PR.
I'm going to merge and release it now and expect that we could polish/work it out with bugfix releases if necessary.
Feel free to send new PRs with adding tests/fixing test client to support unix sockets etc.
I'm going to wire up CherryPy config options to this feature a bit earlier.

Issue #38, opened 2017-07-11
cheroot.server.Server() does not correctly parse the request_uri and set the query string.


bertjwregeer:
* **I'm submitting a bug report**

* **What is the current behavior?**

Unfortunately the code is broken and doesn't correctly set `self.qs`.

https://github.com/cherrypy/cheroot/blob/60ec5b6bd665b97ed8b46b34dc5a86738764d4d7/cheroot/server.py#L709-L724

Does not currently correct parse the `reques_uri` and set `self.qs`.

Anyone using cheroot with CherryPy will not have `environ['QUERY_STRING']` set correctly, which was fun to debug.

* **What is the expected behavior?**

To parse the `request_uri` correctly and set `self.qs` so that it may be set in the WSGI `environ`.

* **What is the motivation / use case for changing the behavior?**

`urlparse` will not return the query string in the path, so it needs to be appended again, or returned from `parse_request_uri`, then the check on line https://github.com/cherrypy/cheroot/blob/60ec5b6bd665b97ed8b46b34dc5a86738764d4d7/cheroot/server.py#L723 will no longer be necessary.

* **Please tell us about your environment:**

- Python version: 2.7.x
- Cheroot version: 5.7.0 (issue has existed since 5.0.0: https://github.com/cherrypy/cheroot/blob/v5.0.0/cheroot/server.py#L785)

PR #39, opened 2017-07-12
Bugfix/query string

bertjwregeer:
Fix for #38


cheroot/server.py
```diff
@@ -874,14 +870,14 @@ def parse_request_uri(self, uri):
             # If there's a scheme (and it must be http or https), then:
             # http_URL = "http:" "//" host [ ":" port ] [ abs_path [ "?" query
             # ]]
-            return parsed.scheme, parsed.netloc, parsed.path
+            return parsed.scheme, parsed.netloc, parsed.path, parsed.query
```
webknjaz: Could you please alter the docsrting of this method to match new behavior?

bertjwregeer:
I can update the docstring and the return line as requested.

I will need help writing test cases for the rest of it, I don't see any unit tests for this piece of code to verify that it works as intended (unless I missed them in my grepping). I am not sure that a full on HTTP request will hit all of the points it currently attempts to support.

bertjwregeer:
Docstring and return statement added.

webknjaz:
@bertjwregeer possible test cases to cover are `uri in ['*', '/some_abs_uri', 'some_rel_uri']`. Feel free to add 'em next to the one you already wrote. Thanks in advance :)

bertjwregeer:
- `*` is only valid in the case of an OPTIONS request, and can't be tested with a simple GET
- `relative_uri` are never valid in the HTTP spec
- `absolute_path` is already tested, since it is the default
- `absolute_uri` is valid for a proxy and not for a general HTTP server and should be rejected unless the server is a proxy server.

See https://tools.ietf.org/html/rfc7230#section-5.3

I have added the tests,  but I don't believe that `getPage()` will pass them through without modifying them by prepending a `/`.

bertjwregeer:
I also noticed that this is false: https://github.com/bertjwregeer/cheroot/blob/791f126947bd9406fe5648a089be778106672544/cheroot/server.py#L722

Since the returned path from urlparse will have it's params stripped (at least for the last component), since it is parsed. You'll want to modify `parse_request_uri` to return those too. Since it is part of "parsed.params". See https://docs.python.org/2/library/urlparse.html

bertjwregeer:
Actually Python recommends using urlsplit instead: https://docs.python.org/2/library/urlparse.html#urlparse.urlsplit since it will leave the path parameters alone for later processing.

bertjwregeer:
Alright. I'm done.

webknjaz:
Sorry, I'm at EuroPython currently. I'll check the changes ASAP once have some free time. ETA: this week.
If I fail to give you a feedback or release changes, please feel free to ping me on Monday.

P.S. Thank you for your efforts! You've done much more than I expected, I really appreciate this.

josephtate:
@webknjaz do you still have objections?

webknjaz:
Sorry, I wrote a few comments and forgot to hit publish button. Yes, there are a couple of nitpicks.

bertjwregeer:
You have push permissions to the branch on my repo as a maintainer. Feel free to make those modifications and push there; or pull these changes into a local branch and fixup before pushing back to a new PR.

I already maintain Waitress, I really have no want or desire to maintain or spend even more time fixing yet another Python HTTP server.

webknjaz:
Deal, thanks for your contribution anyway :)

cheroot/server.py
```diff
@@ -698,42 +705,73 @@ def read_request_line(self):
...
+                atoms = [
+                    unquote_to_bytes(x)
+                    for x in quoted_slash.split(path)
+                ]
+            except ValueError as ex:
+                self.simple_response('400 Bad Request', ex.args[0])
+                return False
+            path = b'%2F'.join(atoms)
+
+        if path[0] != b'/':
```
webknjaz: Wow! It turned out that `path[0]` returns `int`, not `bytes`, which is `47` for `/`, thus prepending `/` unconditionally.

(just a note)

webknjaz: Ref: [PEP 358](https://www.python.org/dev/peps/pep-0358/)

```python
.__getitem__(int | slice) -> int | bytes
```

cheroot/server.py
```diff
@@ -598,12 +599,17 @@ class HTTPRequest(object):
     A HeaderReader instance or compatible reader.
     """
 
-    def __init__(self, server, conn):
+    def __init__(self, server, conn, proxy_mode=False, strict_mode=True):
```
jaraco: I'm not a fan of boolean parameters, especially when they represent a switch on behavior, because they often are limited in what they can describe and lead to branchy and unmaintainable code. Often it's better to use subclasses or handler functions or enumerated values to signal behavior. I'm not sure what's the best approach here without digging deeper, but I wanted to signal this potential issue.

webknjaz: I see your point, but currently I cannot suggest cleaner solution. So for now I'm going to merge this as is.

webknjaz:
@bertjwregeer I've cut the tag [v5.8.0](https://github.com/cherrypy/cheroot/releases/tag/v5.8.0), it will be available @ PYPI once [this build](https://travis-ci.org/cherrypy/cheroot/builds/259849925) succeeds.

webknjaz:
**JFI:** merging this broke some things and integration with CherryPy, so this is going to be fixed.

PR #41, opened 2017-07-26
Catch error of builtin ssl when used with newest OpenSSL

Safihre:

Bug fix

https://github.com/sabnzbd/sabnzbd/issues/853
https://forums.sabnzbd.org/viewtopic.php?f=3&t=22425&p=112784#p112778

When using the builtin SSL provider in Python 2 but with a new version of OpenSSL (1.10) that is not *really* supported by Python 2, this error will be thrown but connections will then later still work.


Safihre:
This fix turns out not to work, the error thrown is of the general `Error` class and not `SSLError`. 
Sorry!

PR #44, opened 2017-08-09
Correctly catch EPROTOTYPE on macOS

Safihre:
@webknjaz @The-Compiler 
A helpful user that got this error a lot debugged this for me. 
`cheroot.errors.plat_specific_errors('EPROTOTYPE')` produces an array, so appending an array to an array gives...
```
socket_errors_nonblocking = [35, [41]]
```
So seems to me the original patch never worked either?


The-Compiler:
Oh, my bad... I think you're right, in retrospective I'm not sure how this ever could've worked 😊

webknjaz:
I wish there were tests for this error handling

PR #52, opened 2017-10-13
Require client certificates

dpeschman:
Feature

cherrypy/cherrypy#1001 (CherryPy) — partial implementation

Using the built-in SSL adapter, client certificates are neither required nor validated with a CA.  Client certificate fields (subject, issuer) are not passed downstream via WSGI environment.

New parameter `verify_mode` on BuiltinSSLAdapter can be one of none/optional/required to specify whether client certificates are required.  These values map to Python's verify_mode values [1].  The default is "none" for backwards compatibility.

Subject and issuer of the client certificate are passed downstream via WSGI environment.  For instance, certificate subject is passed down as SSL_CLIENT_S_DN_* fields.

* **Other information**:
This patch was inspired by, and much of the code derived from, the patch in CherryPy cherrypy/cherrypy#1001.  That change includes additional features, such as hostname verification and parallel support for pyOpenSSL, which are omitted here.

There will be a forthcoming PR to cherrypy with an `ssl_verify_mode` on server adapter to pass along to the SSL adapter.

I plan to incorporate the tests from CherryPy cherrypy/cherrypy#1001, but I expect most of them belong in CherryPy.  I'm happy to add tests here for these changes, though, if you'd like.

CONTRIBUTING.rst has a note to _ensure you set up pre-commit utility correctly and TravisCI tests pass_, but it is not clear to me how to do this, and your guidance is appreciated.

[1] https://docs.python.org/3.5/library/ssl.html?highlight=ssl#constants

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/52)


webknjaz:
@dpeschman Thank you for your contribution!

> CONTRIBUTING.rst has a note to _ensure you set up pre-commit utility correctly and TravisCI tests pass_, but it is not clear to me how to do this, and your guidance is appreciated.

Actually, just `pip install pre-commit` and then `pre-commit install` before committing of `pre-commit run --all-files` to run it manually.
CI already runs that, so you can look up errors from logs there. Installing it manually is just for convenience, so that you could catch errors early :)

webknjaz:
Just a friendly tip: here's how to refer other repo's issue — `cherrypy/cherrypy#1001`. JFYI, i've fixed it in your post

cheroot/ssl/__init__.py
```diff
@@ -17,12 +17,14 @@ class Adapter(object):
     """
 
     @abstractmethod
-    def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None):
+    def __init__(self, certificate, private_key, certificate_chain=None, ciphers=None,
+                 verify_mode=None):
```
webknjaz: You don't have to have 80 chars wrap: this project's linters are configured to allow up to 120-chars lines. We don't work in 80-char terminals nowadays :)

cheroot/ssl/builtin.py
```diff
@@ -117,8 +128,73 @@ def get_environ(self, sock):
...
         """Return socket file object."""
         return MakeFile(sock, mode, bufsize)
+
+    VERIFY_MODES = {
+        'none': ssl.CERT_NONE,
+        'optional': ssl.CERT_OPTIONAL,
+        'required': ssl.CERT_REQUIRED,
+    }
+
+    def get_verify_mode_constant(self, spec):
```
webknjaz: I'd prefer pythonic way implementing `@property` setter/getter.

webknjaz:
@dpeschman I didn't review the PR in detail, but I've posted a few comments, so that we could proceed with discussion/reaction :)

dpeschman:
@webknjaz Thanks for the review and the tips n' tricks!
I've implemented the changes you suggested.

cheroot/ssl/builtin.py
```diff
@@ -45,12 +45,43 @@ class BuiltinSSLAdapter(Adapter):
...
+        None: ssl.CERT_NONE,
+        'none': ssl.CERT_NONE,
+        'optional': ssl.CERT_OPTIONAL,
+        'required': ssl.CERT_REQUIRED,
+    }
+
+    _verify_mode = None
+
+    @property
+    def verify_mode(self):
```
webknjaz: I think, we should keep any method declarations below `__init__`

cheroot/ssl/builtin.py
```diff
@@ -117,8 +150,56 @@ def get_environ(self, sock):
             # SSL_VERSION_INTERFACE     string  The mod_ssl program version
             # SSL_VERSION_LIBRARY   string  The OpenSSL program version
         }
+
+        client_cert = sock.getpeercert()
+        if client_cert:
+            for cert_key, env_var in self.CERT_KEY_TO_ENV.items():
+                ssl_environ.update(
+                    self.env_dn_dict(
```
webknjaz: Plz do this a bit more flat for the sake of readability

cheroot/ssl/builtin.py
```diff
@@ -45,12 +45,43 @@ class BuiltinSSLAdapter(Adapter):
...
+        - "required": peer cert required (ssl.CERT_REQUIRED)
+        """
+        return self._verify_mode
+
+    @verify_mode.setter
+    def verify_mode(self, value):
+        if value not in self.VERIFY_MODES:
+            raise ValueError('Invaid verify_mode "%s"; must be one of %s' % (
+                             value, self.VERIFY_MODES.keys()))
+        self._verify_mode = value
```
webknjaz: please transform `'none'` string to `None` here, so that `if self.verify_mode:` check would evaluate stable result

cheroot/ssl/builtin.py
```diff
@@ -117,8 +150,56 @@ def get_environ(self, sock):
             # SSL_VERSION_INTERFACE     string  The mod_ssl program version
             # SSL_VERSION_LIBRARY   string  The OpenSSL program version
         }
+
+        client_cert = sock.getpeercert()
```
webknjaz: I'd add `if self.verify_mod` check in order to avoid unnecessary syscall when this feature is disabled.

cheroot/ssl/builtin.py
```diff
@@ -117,8 +150,56 @@ def get_environ(self, sock):
...
+        if client_cert:
+            for cert_key, env_var in self.CERT_KEY_TO_ENV.items():
+                ssl_environ.update(
+                    self.env_dn_dict(
+                        env_var, client_cert.get(cert_key)))
+
         return ssl_environ
 
+    CERT_KEY_TO_ENV = {
```
webknjaz: Please keep class properties at the top, above method declarations

cheroot/ssl/builtin.py
```diff
@@ -117,8 +150,56 @@ def get_environ(self, sock):
...
+        """
+        if not cert_value:
+            return {}
+
+        env = {}
+        # dn = []
+        for rdn in cert_value:
+            for attr_name, val in rdn:
+                attr_code = self.CERT_KEY_TO_LDAP_CODE.get(attr_name)
+                if attr_code:
```
webknjaz: Doing try/except rather then if/else is more pythonic way.

cheroot/ssl/builtin.py
```diff
@@ -117,8 +150,56 @@ def get_environ(self, sock):
...
+                    env['%s_%s' % (env_prefix, attr_code)] = val
+                    # dn.append("%s=%s" % (attr_code, val))
```
webknjaz: do you still need this comment?

webknjaz:
Additionally, I'd really appreciate submitting tests, cause SSL is fairly unstable part of Cheroot.

webknjaz:
This pull request introduces 1 alert - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-ce83b8735b6f30adb128832433f8afe7282eb9bc)

**new alerts:**

* 1 for Comparison using is when operands support __eq__

---

jaraco:
> ensure you set up pre-commit utility correctly and TravisCI tests pass

What I do is run `tox -e pre-commit`.

cheroot/ssl/builtin.py
```diff
@@ -76,7 +123,8 @@ def wrap(self, sock):
                                     server_side=True, certfile=self.certificate,
                                     keyfile=self.private_key,
                                     ssl_version=ssl.PROTOCOL_SSLv23,
-                                    ca_certs=self.certificate_chain)
+                                    ca_certs=self.certificate_chain,
```
jaraco: I'm pretty sure based on other code I've seen that `ssl.wrap_socket` is the legacy way to handle SSL and that `context.wrap_socket` is the preferred way to handle it. The whole reason the concept of the `context` was added was to avoid scattering SSL parameters as attributes of every class that touches sockets. Instead, the context object encapsulates these parameters.

If that's true (and I'm not certain it is), any support for SSL cert validation should be provided first on self.context and then only if absolutely necessary, also provide legacy support. Perhaps the context object is sufficient to satisfy all modern deployments.

In fact, I'd  be tempted to say that the legacy technique for wrapping a socket should soon be deprecated and removed in favor of the context model.

Still, I don't think that eliminates the value of this PR. This PR could indeed adapt itself to demonstrate and test the context-based SSL certificatate validation support, but at the same time avoid the added complexity of additional pass-through parameters (with interpretation).

What do you think of that approach?

cheroot/test/test_https.py
```diff
@@ -0,0 +1,118 @@
...
+        self.assertStatus('200 OK')
+
+    def assert_reject(self, client_cert):
+        self.set_client_cert(client_cert)
+        with pytest.raises(ssl.SSLError) as context:
+            self.getPage()
+        return context
+
+
+@ddt.ddt
```
jaraco: I like the use of ddt, and if weren't married to unittest by way of webtest and could use pytest fixtures instead, I'd prefer those. But I'm glad to see these work here.

cheroot/ssl/builtin.py
webknjaz: @dpeschman

jaraco: > I'd be tempted to say that the legacy technique for wrapping a socket should soon be deprecated and removed in favor of the context model.

I have so many things on my mind that I hadn't connected that only a few days prior, I'd done just that in e4be9d6 (pending review in #60).

webknjaz:
@dpeschman this branch is out of sync, rebase and conflicts resolution is required

dpeschman:
@jaraco Providing my own SSL context would be a superior implementation. A comment in the code led me to believe ssl_context was only used for PyOpenSSL. I’ll try providing my own context directly next time I iterate on this project.

dpeschman:
@webknjaz Roger that.

dpeschman:
Rebase onto latest master; no substantial changes.

dpeschman:
@jaraco @webknjaz I got rid of the new SSL Adapter parameter.  This change now consists only of passing client certificate fields through to the WSGI environment, and unit testing of HTTPS with various SSL verify modes.

cheroot/test/test_https.py
webknjaz: And we fully support using pytest fixtures now :)

webknjaz: #67 (FTR)

cheroot/ssl/builtin.py
```diff
@@ -43,6 +43,21 @@ class BuiltinSSLAdapter(Adapter):
     ciphers = None
     """The ciphers list of SSL."""
 
+    CERT_KEY_TO_ENV = {
+        'subject': 'SSL_CLIENT_S_DN',
+        'issuer': 'SSL_CLIENT_I_DN'
```
webknjaz: Could you plz add a trailing comma?

cheroot/test/test_https.py
```diff
@@ -0,0 +1,146 @@
...
+    def setUp(self):
+        """TestCase setup."""
+        super().setUp()
+        self.ssl_context = ssl.create_default_context(
+            cafile='cheroot/test/ssl/ca.cert')
+        self.ssl_context.check_hostname = False
+
+    def getPage(self, *args, **kw):
+        """Fetch the page."""
+        return super().getPage(self.script_name, *args,
```
webknjaz: This is incompatible with Python 2.7 https://travis-ci.org/cherrypy/cheroot/jobs/336749016#L543

cheroot/test/test_https.py
```diff
@@ -0,0 +1,146 @@
...
+# vim: set fileencoding=utf-8 :
+
+import os
+import pytest
+import ssl
+
+from cheroot.ssl.builtin import BuiltinSSLAdapter
+from cheroot.test import helper
+import cheroot
+import ddt
```
webknjaz: @dpeschman we now rely on pytest, which means no need for ddt anymore

cheroot/test/ssl/ca.cert
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
webknjaz: I'd like to see those ssl stubs generated (and cleaned up) by pytest fixtures rather then be part of the repo.

cheroot/ssl/builtin.py
```diff
@@ -117,8 +132,34 @@ def get_environ(self, sock):
...
+        See SSL_CLIENT_S_DN_x509 at
+        https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#envvars.
+        """
+        if not cert_value:
+            return {}
+
+        env = {}
+        for rdn in cert_value:
+            for attr_name, val in rdn:
+                attr_code = self.CERT_KEY_TO_LDAP_CODE.get(attr_name)
```
webknjaz: I'd prefer try/except approach over if check. It's easier to ask for forgiveness, then for permission (which is considered more pythonic).

cheroot/test/ssl/ca.cert
webknjaz: (this way it would be like self-documentation for using this feature as well)

cheroot/test/test_https.py
jaraco: Since this PR was originally drafted on unittest-based tests, I'd be inclined to accept this change, rather than asking the contributor to chase our changes.

jaraco:
It looks to me like this effort was abandoned, perhaps out of frustration? In my estimation, the main request that needed to be addressed was the Python 2.7 compatibility.

cheroot/test/test_https.py
webknjaz: I think I could try rewriting this on top of https://github.com/cherrypy/cheroot/tree/experiment/ssl-tests (ref #95). And we also need solid design change of `makefile` module (ref #6).
Anyway, it's currently unmergable because of conflicts and syntax is sometimes incompatible with Python 2. (see my `super()` comment)

P.S. We could probably drop some of hardcoded certificates in favor of `trustme`, which seems to be providing a very smooth experience for TLS testing with fake trusted CA.

jaraco:
I see the pre-commit failing with:

```
Verifying PEP257 Compliance..............................................Failed
hookid: pydocstyle
cheroot/test/test_https.py:45 in public method `setup_server`:
        D401: First line should be in imperative mood; try rephrasing (found 'Setup')
cheroot/test/test_https.py:56 in public method `setUp`:
        D403: First word of the first line should be properly capitalized ('Testcase', not 'TestCase')
```

cheroot/test/test_https.py
jaraco: I've resolved the conflicts and Python 2 syntax issues. Happy to see iterations on the code to eliminate ddt and remove unittest constructs.

PR #53, opened 2017-10-16
Don't use SO_REUSEADDR on Windows

kyrias:

Bug fix(?)

cherrypy/cherrypy#1088

To bind using SO_REUSEADDR on all platforms

To not bind using SO_REUSEADDR on Windows.

* **Other information**:
https://msdn.microsoft.com/en-us/library/ms740621(v=vs.85).aspx

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1545,6 +1545,7 @@ class HTTPServer(object):
         self.socket = socket.socket(family, type, proto)
         prevent_socket_inheritance(self.socket)
         # Windows have different semantics for SO_REUSEADDR, so don't set it there.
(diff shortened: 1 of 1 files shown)


kyrias:
Do you want me to add the msdn link as well?  And should I squash the two commits?

webknjaz:
Yeah, it would be nice to have that link and maybe some test. Don't squash commits.

kyrias:
Not sure what kind of test you were thinking?  Try to bind an instance, and checking if the sockopt is set would work I guess, but it doesn't feel like that would be a very useful test, hm.

webknjaz:
Well, I'll accept this PR without additional test for now. Just wanted to hear your suggestions.

webknjaz:
@kyrias thank you!

PR #60, opened 2017-11-21
Drop support for Python 2.6, 3.1, 3.2, and 3.3. Also Python <2.7.9

jaraco:
I thought I'd submitted a PR for this but now I don't see one.

This request removes support for Python 2.7.8 and earlier on Python 2 and Python 3.2 on Python 3.

Removing support for older Pythons simplifies the code-base and gets us one step closer to supporting only Python 3.

These changes will also simplify the SSL story by supporting only the preferred SSLContext for wrapping sockets.

Python 2.7.9 was released three years ago, so I'm inclined to drop support for 2.7.8, but I'm not dead-set on it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,4 +6,4 @@ jobs:
     steps:
     - checkout
     - run: pip install tox
-    - run: tox -e py27,py33,py34,py35,py36
+    - run: tox -e py27,py34,py35,py36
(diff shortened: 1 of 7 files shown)


webknjaz:
Ref #65

jaraco:
Yeah, let's drop support for 3.3 at the same time.

webknjaz:
@jaraco I've rebased your topic branch and added deprecation of 3.3 there.

PR #66, opened 2017-12-02
Signal WSGI app that the input is terminated

webknjaz:
  * Feature

  * cherrypy/cherrypy#1664
  * Pylons/webob#278
  * unbit/uwsgi#1428

``'wsgi.input_terminated'`` WSGI env var is not set.

``'wsgi.input_terminated'`` WSGI env var is set if the Transfer-Encoding is chunked.

* **Other information**:
This PR sets ``env['wsgi.input_terminated'] = True`` if the request uses
chunked Transfer-Encoding, so that WSGI app would know that it's safe
to read from ``env['wsgi.input']`` stream.

It implements @mitsuhiko's proposal:
https://gist.github.com/mitsuhiko/5721547

///
@davidism and @jaraco please review.

### DIFF
--- a/cheroot/wsgi.py
+++ b/cheroot/wsgi.py
@@ -249,6 +249,7 @@ class Gateway_10(Gateway):
             'SERVER_SOFTWARE': req.server.software,
             'wsgi.errors': sys.stderr,
             'wsgi.input': req.rfile,
+            'wsgi.input_terminated': bool(req.chunked_read),
             'wsgi.multiprocess': False,
             'wsgi.multithread': True,
             'wsgi.run_once': False,


cheroot/wsgi.py
```diff
@@ -249,6 +249,7 @@ def get_environ(self):
             'SERVER_SOFTWARE': req.server.software,
             'wsgi.errors': sys.stderr,
             'wsgi.input': req.rfile,
+            'wsgi.input_terminated': bool(self.req.chunked_read),
```
davidism: Don't need `self` here.

davidism:
Is `chunked_read` the only time cheroot terminates input? If it also limits the stream when `Content-Length` is present, you can set `wsgi.input_terminated` then too.

cheroot/wsgi.py
webknjaz: Oh.. you're right

davidism:
@webknjaz this looks fine, my earlier comment was just that more could be done, but that isn't requried. This should be ok to merge.

webknjaz:
Thanks!

Issue #42, opened 2017-07-31
[TODO] Get rid of unittest.TestCase


webknjaz:
This is needed to fully take advantage of `py.test` testing framework.

webknjaz:
Ref: https://stackoverflow.com/a/35562401/595220
Ref: https://docs.pytest.org/en/latest/unittest.html

jaraco:
Fully support.

jaraco:
One consideration is some environments may be relying on WebCase as a unittest.TestCase subclass. I suggest separating the web testing functionality into a WebTest class that can be instantiated without any unittest functionality, but also use it as a mix-in for WebCase to continue to provide that functionality in the WebCase.

webknjaz:
I've dropped a few classes in #67

webknjaz:
Next thing to achieve on this is to deprecate webtest and try to make some public test fixture/helper module

jaraco:
In my experience - CherryPy's webtest is helpful, but it also masks errors (because every request happens on a remote side of a call, it's impossible to debug with something like pdb). I've found instead the [webtest](https://pypi.org/project/WebTest/) project and more interestingly [pyriform](https://pypi.org/project/pyriform/), though admittedly I haven't used the latter yet.

webknjaz:
I was also thinking of pylons.webtest and officially recommending to defer WSGI testing to third party, but didn't explore the possibilities and effect of such decision.

webknjaz:
So I looked through pyriform and it looks simple and easy to use. There's no pytest fixtures shipped though, which might be useful. On the other side, we could provide our own fixtures based on this library. I foresee that it would fit in our ecosystem :)

jaraco:
I don't think there is anything called [pylons.webtest](https://pypi.org/project/pylons.webtest/) and when I find [Pylons/webtest](/Pylons/webtest), it seems to be one and the same project as the one I referenced.

And pyriform builds on that same framework.

I know the developer of pyriform and I bet he'd accept pull requests for pytest fixtures, so something definitely worth considering contributing upstream.

webknjaz:
unstale

webknjaz:
unstale

jeffvanvoorst:
I would like to tackle this.  I am familiar with the requests, pytest, and unittest modules.

I did see that pyriform links the requests and WebTest modules.  Is it OK to have all 3 modules and dependencies as requirements for unit tests?

[...]

PR #67, opened 2017-12-02
WIP: Drop unused unittest components from test helpers and replace tests with pytest-based ones

webknjaz:
Cleanup

#42

TerseTestResult, TerseTestRunner and ReloadingTestLoader exist.

Drop above.

* **Other information**:


jaraco:
Although it's found in the 'tests' package, the webtest module in particular is part of the public interface of cheroot and cherrypy.

I did a [quick search](https://github.com/search?q=ReloadingTestLoader&type=Code&utf8=%E2%9C%93) for ReloadingTestLoader, which turns up quite a few results. If there are users relying on this functionality, this change will need to describe what the recommendation is for those users. Maybe the recommendation is they adapt their tests to a new model--a fairly harsh ask. Or maybe we release this functionality in a separate package. I can't decide whether it's worth an intermediate release deprecating this functionality first.

In any case, I'm largely in favor of this change as long as we can make it without causing undue disruption.

cheroot/test/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+import time
+
+import pytest
+
+import cheroot.server
+import cheroot.wsgi
+
+
+config = {
+    'bind_addr': ('127.0.0.1', 54583),
```
jaraco: Here I'd like to see support for IPv6 and I'd like not to see hard-coded port numbers... which could conflict with other applications or other instances of the tests running. I suggest using `portend.find_available_local_port` or binding to the ephemeral port 0 and then resolving the bind port for the client.

webknjaz:
> I did a quick search for ReloadingTestLoader, which turns up quite a few results.

I followed the search and found out that all (or nearly all) of the results are just copies of cherrypy repo.

cheroot/test/conftest.py
webknjaz: Yeah, I was thinking about it and prefer ephemeral port 0 approach as it's safer to defer port allocation to the OS and `portend` dependency is not even needed in this case.

I was trying to decouple existing `unittest.TestCase` based class into pieces, thus I didn't really take any steps to optimize or add new behavior at this stage. I didn't even replace the test suite module, but added new one because of this experimenting process.

webknjaz:
@jaraco 

Do you think anyone uses/needs `WebCase._handlewebError`?

If yes, do you have any ideas on how to reimplement this using pytest?

jaraco:
I don't know about _handlewebError, but I would recommend that in a pytest scenario, one shouldn't use _handlewebError, and the tests should just make assertions and let the test runner handle interactivity (such as through --pdb). Given that it's got the underscore prefix, I'd say it's sane to deprecate/remove it.

webknjaz:
@jaraco well, it's kinda public, because depending on the environment variable (which I would consider to be a public interface) it intercepts assertion failures and provides an interactive interface for accessing/printing HTTP request/response values and suppressing/raising a real failures during test runs.

On the other hand I don't really see any need for this interface.

Also, I'd like to provide some new public pytest-oriented fixtures/helpers in `cheroot.testing` module rather that mixing them into `cheroot.test.webtest`

cheroot/test/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+    'bind_addr': (NO_INTERFACE, EPHEMERAL_PORT),
+    'wsgi_app': None,
+}
+
+
+def cheroot_server(server_factory):
+    conf = config.copy()
+    bind_port = conf.pop('bind_addr')[-1]
+
+    for interface in ANY_INTERFACE_IPV6, ANY_INTERFACE_IPV4:
```
webknjaz: @jaraco turns out, if Linux box supports dual stack it will bind to `::`, but also be available as `0.0.0.0:sameport`. And this [raises `OSError: [Errno 99] Cannot assign requested address` on the client side](https://travis-ci.org/cherrypy/cheroot/jobs/320644912#L593).

One can emulate this on linux using `sudo sysctl net.ipv6.conf.all.disable_ipv6=1`, which will disable IPv6 addresses on interfaces, but `socket.has_ipv6` **will still remain `True`**, because *kernel is built with IPv6 support*.

@jaraco do you have any ideas on how can we treat this?

jaraco: I've [previously posted that Python should provide a routine for binding in dual-stack environments](https://bugs.python.org/issue25667), but in that ticket, I reference back to the CherryPy implementation, which is now part of this project.

You say that one can emulate this on Linux by disabling IPv6 - is that what's happening in Travis?

jaraco: Looks like there are recent, emergent [issues with IPv6 on Travis](https://github.com/travis-ci/travis-ci/issues/8891), and that's probably what's happening - IPv6 is only partially available. Perhaps this routine for creating a server should check that a connection can be established on IPv6 (not just that a connection can be bound on IPv6) when setting up the server... or maybe in the server_client, fallback to IPv4 as a workaround for IPv6-broken environments like Travis.

.travis.yml
```diff
@@ -158,6 +158,8 @@ script:
 - tox
 
 after_failure:
+- ip a
+- sysctl -a
```
webknjaz: @jaraco

> You say that one can emulate this on Linux by disabling IPv6 - is that what's happening in Travis?

Yes, and I've added this output to confirm that IPv6 is disabled.

cheroot/test/conftest.py
webknjaz: I've seen this kind of issue lots of times in Travis and don't want to rely on its state.

webknjaz: I was hoping that there's some clever syscall capable of telling us whether IPv6 is disabled in sysctl

webknjaz: > Perhaps this routine for creating a server should check that a connection can be established on IPv6 (not just that a connection can be bound on IPv6) when setting up the server... or maybe in the server_client, fallback to IPv4 as a workaround for IPv6-broken environments like Travis.

Yes, I was thinking about such options, but I still don't have a clear vision w/o hacks on how to do this check sanely. Technically, I can connect to the server if I replace `::1` with `127.0.0.1`, but I don't like such hardcoding. On one hand, I'd leave `bind_addr` representation as it comes from the OS in the server object, because it's how things work in such envs, on the other hand, I'd like to see some limitations/flags that it also listens to IPv4 on the same port to allow client decide on the best protocol available from its perspective.

.travis.yml
webknjaz: @jaraco please note that the same issue happens in CircleCI: https://circleci.com/gh/cherrypy/cheroot/223

cheroot/test/conftest.py
webknjaz: New info, to be clear:

According to https://wiki.archlinux.org/index.php/IPv6#Disable_functionality, there's `ipv6.disable=1` and `ipv6.disable_ipv6=1`. The first one is destined to disable the whole IPv6 network stack in kernel (also achieved by `rmmod` if it's been built as a separate module, rather than built-in to kernel itself). **The second one will keep the IPv6 stack functional but will not assign IPv6 addresses to network devices.** And this is exactly what's happening.

So I think it should be handled on the client side.

cheroot/test/pure_pytest/test_core.py
```diff
@@ -0,0 +1,368 @@
...
+
+
+@pytest.fixture
+def server_client(testing_server):
+    host, port = testing_server.bind_addr
+
+    interface = webtest.interface(host)  # :: -> ::1; 0.0.0.0 -> 127.0.0.1
+    host = socket.getfqdn(interface)  # {::1|127.0.0.1} -> localhost
+    # localhost -> {::1|127.0.0.1} :
+    interface = socket.gethostbyaddr(host)[-1].pop(0)
```
webknjaz: @jaraco look at how I fixed this issue for now

cheroot/test/pure_pytest/test_core.py
```diff
@@ -77,10 +77,27 @@ def testing_server(wsgi_server):
 def server_client(testing_server):
     host, port = testing_server.bind_addr
 
-    interface = webtest.interface(host)  # :: -> ::1; 0.0.0.0 -> 127.0.0.1
-    host = socket.getfqdn(interface)  # {::1|127.0.0.1} -> localhost
-    # localhost -> {::1|127.0.0.1} :
-    interface = socket.gethostbyaddr(host)[-1].pop(0)
+    interface = webtest.interface(host)
+
+    def probe_ipv6_sock(interface):
```
webknjaz: @jaraco Had to return to this solution, because of Travis has separate `ip6-localhost` entry in `/etc/hosts`

webknjaz:
@jaraco I've added `test_conn.py`, pure pytest version.

Do we need these `test_config_server.py` and `test_wsgiapps.py` at all? It looks like they are disabled, because they depend (test?) CherryPy.

webknjaz:
@jaraco could you plz share your opinion regarding public helpers for end-users (aka `cheroot.testing`)?

jaraco:
I agree, creating an officially-supported `testing` module and separating it from a private `tests` package makes a lot of sense. Of course, doing so would be a backward-incompatible change, but we can manage that with deprecations and compatibility shims and iterative releases.

cheroot/server.py
```diff
@@ -1600,6 +1603,7 @@ def bind(self, family, type, proto=0):
                 pass
 
         self.socket.bind(self.bind_addr)
+        self.bind_addr = self.socket.getsockname()[:2]  # TODO: keep separate
```
jaraco: This change seems dangerous to me. You're overwriting the state on the object... so where one could previously access `self.bind_addr` to determine what address was attempted to be bound, now one gets instead the addr than was bound. Plus, there's a TODO - not sure why not just do:

```
    @property
    def bound_addr(self):
        return self.socket.getsockname()[:2]
```

or even better, just let the caller call server.socket.getsockname() directly.

cheroot/server.py
```diff
@@ -1693,16 +1697,17 @@ def tick(self):
                 return
             raise
 
-    def _get_interrupt(self):
+    @property
+    def interrupt(self):
+        """Return interrupt Exception instance."""
```
jaraco: This docstring is the one that will be present on the property, whereas previously the docstring would have provided indication about setting the exception. The docstring should be copied from the setter and the setter should probably have no meaningful docstring, maybe something like "setter for interrupt".

cheroot/test/webtest.py
```diff
@@ -313,7 +213,16 @@ def interactive(self):
...
         """
         env_str = os.environ.get('WEBTEST_INTERACTIVE', 'True')
-        return bool(json.loads(env_str.lower()))
+        is_interactive = bool(json.loads(env_str.lower()))
+        if is_interactive:
+            warnings.warn(
+                'We are moving to pytest-based testing and thus deprecating '
+                'WEBTEST_INTERACTIVE environment variable support. '
+                'Interactive test failure interceptor is going to be removed '
+                'in Cheroot v7.0.0',  # TODO: decide whether it's v7 or v8
```
jaraco: I've learned that it's difficult to predict when a feature will be dropped unless you have a project with long iterations and scheduled releases, neither of which apply to cheroot. I would simply say "WEBTEST_INTERACTIVE is deprecated." ...and in a future, backward-incompatible release, which we will determine when we get to it, we drop support.

cheroot/server.py
webknjaz: Fair enough

webknjaz: But still the docstring should describe the property/getter

webknjaz: I'll try to rephrase that

cheroot/server.py
webknjaz: Yeah, it's just a dirty hack I did while refactoring tests. I wanted to postpone its resolution and to not dig into it while focusing on tests rewrite.
I don't want to mix this into the current PR, so I've logged this in #70

webknjaz:
@jaraco thanks for the review! :dizzy: :rocket: :robot:

webknjaz:
Confirmed that it's compatible with CherryPy

PR #74, opened 2018-01-26
WIP: Bypass URI as path only if absolute for OPTIONS

webknjaz:
I'm not sure whether it's right from the RFC point of view, but this allows to put path w/o query-string into PATH_INFO in case of origin-form.

Bug fix (possibly)

cherrypy/cherrypy#1662

PATH_INFO in WSGI env is assigned URI as is if proxy is enabled.

PATH_INFO is assigned to path part if URI is not absolute.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/74)


webknjaz:
@bertjwregeer @josephtate @jaraco Could you please share your opinion regarding RFC compliance?

bertjwregeer:
That change looks fine to me from a RFC perspective.

webknjaz:
@bertjwregeer thanks.
What about `proxy_mode`? Maybe it's better to raise an error when there's no absolute path present?

webknjaz:
Here's what I think:
* If we expect it to act like proxy, there's not enough info if only origin-form is present in Request-Line. absolute-form is needed for proxy to know where to connect to. So in case of enabled proxy mode I'd raise an error if absolute-form validation fails.
* It should be okay to set PATH_INFO to absolute-path in case of disabled proxy mode, but
  * [RFC encourages to accept absolute-form for all queries](https://tools.ietf.org/html/rfc7230#section-5.3.2) and instructs servers implementing Virtual Hosts to override Host header with a host from absolute-form. It looks like this should be handled on another level, but WSGI requires to have just absolute-path part there.
* Based on that, I'd just turn off proxy mode in cherrypy adapter.

jaraco:
I'm okay with this change. Will be good to stabilize the behavior reported in the ticket.

webknjaz:
Actually, I'm not okay with it now:) But haven't had time to put a new patch together

danlamanna:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thlank you for your contributions.

I would hate to see this issue/PR fall off the radar since it's caused us to pin CherryPy to an older version in our downstream project (https://github.com/girder/girder/).

Seeing as how the related issue has broken long lived expectations of CherryPy's handling of OPTIONS requests, is it possible to get this patched in the meantime?

webknjaz:
@danlamanna I need to re-do this patch in order to have it merged. I just didn't have time for this. I wish there were more maintainers who could help.

bertjwregeer:
Stale bot... go away. This is still an issue.

webknjaz:
Now bot won't close it with actual `bug` label.

webknjaz:
@jaraco should we _think_ of deferring HTTP parsing to https://github.com/python-hyper/h11?

fviard:
Hi, today I encountered this issue also using cherrypy and just trying to use query params...
That would be nice if that could be fixed soon... And also if there was a "hack" that can be used with existing versions.
FYI, applying the fix proposed here solve my issue.
thanks

duper51:
@fviard You're going to end up using a version of cheroot that is not RFC compliant, so you might have some user agents that misbehave and/or introduce yourself to various vulnerabilities.

cheroot/server.py
```diff
@@ -827,12 +827,15 @@ def read_request_line(self):
...
+        uri_is_absolute_form = (scheme or authority)
+
         if self.method == b'OPTIONS':
             # TODO: cover this branch with tests
             path = (
                 uri
                 # https://tools.ietf.org/html/rfc7230#section-5.3.4
-                if self.proxy_mode or uri == ASTERISK
+                if (self.proxy_mode and uri_is_absolute_form)
+                or uri == ASTERISK
```
jaraco: I'm pretty sure this line can be eliminated as `urllib.parse.urlsplit('*').path == '*'` anyway.

jaraco: Removed in dc9592b6.

webknjaz:
FTR I recall there have been some corner cases there in conjunction with CherryPy.

cheroot/server.py
webknjaz: @jaraco where did you see `urllib.parse.urlsplit('*').path == '*'`?

webknjaz: Because I don't see any code checking for `ASTERISK` now.

webknjaz: I think that this change is wrong. And the problem with this PR was this hack: https://github.com/cherrypy/cherrypy/blob/535605c/cherrypy/_cpwsgi_server.py#L31

It was temporary until this PR's figured out (https://github.com/cherrypy/cherrypy/commit/c87bc9f73740dfe35e4d2ca9082dccb6278bdc53). We should revert it to `proxy_mode=False` now and see if it's ok.

PR #75, opened 2018-02-08
Ensure response headers are being sent before closing the response

josejulio:
Fixes https://github.com/cherrypy/cherrypy/issues/1404

Bugfix

https://github.com/cherrypy/cherrypy/issues/1404

on_end_request hook is being call **before** sending a response on cherrypy

on_end_request hook is being call **after** sending a response on cherrypy

* **Other information**:

I found https://github.com/Lawouach/WebSocket-for-Python/issues/158 which lead me to this "PR" https://bitbucket.org/cherrypy/cherrypy/pull-requests/121/fixing-issue-1404-on_end_request-hook-is/diff 
I ported that to cheroot and addressed the last comment from  "Sylvain Hellegouarch"

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -973,9 +973,6 @@ class HTTPRequest(object):
 
         self.server.gateway(self).respond()
 
-        if (self.ready and not self.sent_headers):
-            self.sent_headers = True
-            self.send_headers()
         if self.chunked_write:
             self.conn.wfile.write(b'0\r\n\r\n')
 
@@ -1015,6 +1012,11 @@ class HTTPRequest(object):
         except socket.error as ex:
             if ex.args[0] not in errors.socket_errors_to_ignore:
                 raise
+
+    def write_headers(self):
+        """Write headers to the client if not already sent."""
+        self.sent_headers = True
+        self.send_headers()
 
     def write(self, chunk):
         """Write unbuffered data to the client."""
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -972,7 +972,7 @@ class HTTPRequest(object):
             self.rfile = KnownLengthRFile(self.conn.rfile, cl)
 
         self.server.gateway(self).respond()
-        self.ensure_headers_sent()
+        self.ready and self.ensure_headers_sent()
 
         if self.chunked_write:
             self.conn.wfile.write(b'0\r\n\r\n')
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -972,6 +972,7 @@ class HTTPRequest(object):
             self.rfile = KnownLengthRFile(self.conn.rfile, cl)
 
         self.server.gateway(self).respond()
+        self.ensure_headers_sent()
 
         if self.chunked_write:
             self.conn.wfile.write(b'0\r\n\r\n')
@@ -1013,10 +1014,11 @@ class HTTPRequest(object):
             if ex.args[0] not in errors.socket_errors_to_ignore:
                 raise
 
-    def write_headers(self):
-        """Write headers to the client if not already sent."""
-        self.sent_headers = True
-        self.send_headers()
+    def ensure_headers_sent(self):
+        """Ensure headers are sent to the client if not already sent."""
+        if not self.sent_headers:
+            self.sent_headers = True
+            self.send_headers()
 
     def write(self, chunk):
         """Write unbuffered data to the client."""
--- a/cheroot/test/test_core.py
+++ b/cheroot/test/test_core.py
@@ -351,3 +351,56 @@ def test_garbage_in(test_client):
         # "Connection reset by peer" is also acceptable.
         if ex.errno != errno.ECONNRESET:
             raise
+
+
+class CloseController(object):
+    """Controller for testing the close callback."""
+
+    def __init__(self):
+        """Sets the default values."""
+        self.sent_headers_before_closing = None
+        self.req = None
+
+    def __call__(self, environ, start_response):
+        """Gets the req to know header sent status."""
+        self.req = start_response.__self__.req
+        resp = CloseResponse(self.close)
+        start_response(resp.status, resp.headers.items())
+        return resp
+
+    def close(self):
+        """Hook for close, tests if headers already sent."""
+        self.sent_headers_before_closing = \
+            self.req is not None and self.req.sent_headers
+
+
+class CloseResponse(object):
+    """Dummy empty response to trigger the no body status."""
+    def __init__(self, close):
+        """Use some defaults to ensure we have a header."""
+        self.status = '200 OK'
+        self.headers = {'Content-Type': 'text/html'}
+        self.close = close
+
+    def __getitem__(self, index):
+        """Ensures we don't have a body."""
+        raise IndexError()
+
+    def output(self):
+        """Returns self to hook the close method."""
+        return self
+
+
+@pytest.fixture
+def testing_server_close(wsgi_server_client):
+    wsgi_server = wsgi_server_client.server_instance
+    wsgi_server.wsgi_app = CloseController()
+    wsgi_server.max_request_body_size = 30000000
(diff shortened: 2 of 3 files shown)


webknjaz:
Original PR authors/reviewers seem to be: @moigagoo @ronin8600 @Lawouach

webknjaz:
Thanks for the PR.
I cannot accept it w/o manually checking the flow you're changing, so I'll do this once I have some time. This should also trigger checks against CherryPy itself.
Covering this with tests might speed up the process.

josejulio:
Thanks @webknjaz 
I'll try to make time to write test. 
Basically this writes the headers before `closing` the request.
See [here](https://github.com/cherrypy/cheroot/pull/75/files#diff-08d927a6f63bcf494ed88cca2a6873d0R141) and [here](https://github.com/cherrypy/cheroot/pull/75/files#diff-b0366adf530cee9249c1888ba4f32260L976)

I suppose (haven't digged much on cherrypy code) cherrypy hooks on that [close](https://github.com/cherrypy/cheroot/pull/75/files#diff-08d927a6f63bcf494ed88cca2a6873d0R144) and triggers the `on_end_request`

josejulio:
@webknjaz I added a test, I verified that it fails without this patch - and it passes with the patch :) -

Let me know how this can be improved.

webknjaz:
@josejulio first of all, linters in CI fail: https://travis-ci.org/cherrypy/cheroot/jobs/340814348#L484
We use fail fast approach in our CI to save on computing resources and power, so tests are not even being run if fast checks crash.
Could you please correct this?

P.S. Also, it's possible to `pip install pre-commit && pre-commit install` locally to set up automatic linter running for each commit you do on your machine :) So that you won't wait for CI to be triggered and return the result.

cheroot/test/test_core.py
```diff
@@ -351,3 +351,56 @@ def test_garbage_in(test_client):
...
+    wsgi_server = wsgi_server_client.server_instance
+    wsgi_server.wsgi_app = CloseController()
+    wsgi_server.max_request_body_size = 30000000
+    wsgi_server.server_client = wsgi_server_client
+    return wsgi_server
+
+
+def test_send_header_before_closing(testing_server_close):
+    """Tests we are actually sending the headers before closing the response."""
+    testing_server_close.server_client.get('/')
```
webknjaz: How about checking http response headers in client?
```python
resp_status_line, resp_headers, resp_body = testing_server_close.server_client.get('/')
```

cheroot/test/test_core.py
```diff
@@ -351,3 +351,56 @@ def test_garbage_in(test_client):
...
+        """Gets the req to know header sent status."""
+        self.req = start_response.__self__.req
+        resp = CloseResponse(self.close)
+        start_response(resp.status, resp.headers.items())
+        return resp
+
+    def close(self):
+        """Hook for close, tests if headers already sent."""
+        self.sent_headers_before_closing = \
+            self.req is not None and self.req.sent_headers
```
webknjaz: This seems to be overengineered. Please see my comment below.

cheroot/test/test_core.py
josejulio: Sorry, maybe my description is wrong here or i didn't understand well.

The headers are always sent, the problem is that the `close` [1] method is being call before the headers are being sent.

This is a problem in cherrypy, because the `on_end_request` is being called before the headers are actually sent, anything sent to the client from this hook will reach before the headers, thus the headers will be malformed. I don't think I can test this from the client side unless i can get a raw response, but i would still need to send something on that `close` [1] method.

It's done in a similar fashion in pep-0333 [2] 

```python
    try:
        for data in result:
            if data:    # don't send headers until body appears
                write(data)
        if not headers_sent:
            write('')   # send headers now if body was empty
    finally:
        if hasattr(result, 'close'):
            result.close()

```

[1] https://github.com/cherrypy/cheroot/pull/75/files#diff-08d927a6f63bcf494ed88cca2a6873d0R144
[2] https://www.python.org/dev/peps/pep-0333/

josejulio:
@webknjaz I fixed linting issues and tried to simplify the test.

Here is the output of the test without the fix (commit: 742b1b9b9beb589afbf2a202623a21c68405cefa)
```
―――――――――――――――――――――――――――――――――――――― test_send_header_before_closing ――――――――――――――――――――――――――――――――――――――

testing_server_close = <cheroot.wsgi.Server object at 0x7fc57f98a350>

    def test_send_header_before_closing(testing_server_close):
        """Test we are actually sending the headers before calling 'close'."""
        _, _, resp_body = testing_server_close.server_client.get('/')
>       assert resp_body == 'hello'
E       AssertionError: assert 'helloHTTP/1....\r\n0\r\n\r\n' == 'hello'
E         + hello
E         - helloHTTP/1.1 200 OK\r
E         - Content-Type: text/html\r
E         - Transfer-Encoding: chunked\r
E         - Date: Fri, 16 Feb 2018 05:30:56 GMT\r
E         - Server: Cheroot/6.0.1.dev103+gc17bff6d\r
E         - \r...
E         
E         ...Full output truncated (3 lines hidden), use '-vv' to show

cheroot/test/test_core.py:402: AssertionError

```

josejulio:
@webknjaz  @jaraco @Lawouach 
Is there anything else I can do here? Do you have additional comments on the tests?
Thanks.

cheroot/server.py
```diff
@@ -973,9 +973,6 @@ def respond(self):
 
         self.server.gateway(self).respond()
 
-        if (self.ready and not self.sent_headers):
-            self.sent_headers = True
-            self.send_headers()
```
webknjaz: Why completely remove this? How will it affect non-WSGI plain HTTP server?

cheroot/test/test_core.py
```diff
@@ -351,3 +351,52 @@ def test_garbage_in(test_client):
...
+    def __call__(self, environ, start_response):
+        """Get the req to know header sent status."""
+        self.req = start_response.__self__.req
+        resp = CloseResponse(self.close)
+        start_response(resp.status, resp.headers.items())
+        return resp
+
+    def close(self):
+        """Hook for close, write hello."""
+        self.req.write('hello'.encode('utf-8'))
```
webknjaz: `b'hello'` would look nicer

cheroot/wsgi.py
```diff
@@ -137,6 +137,9 @@ def respond(self):
                     raise ValueError('WSGI Applications must yield bytes')
                 self.write(chunk)
         finally:
+            # Send headers if not already sent
+            if not self.req.sent_headers:
```
webknjaz: It's always paired with "if not sent" check. Try moving it inside the method and maybe rename it into smth like `ensure_headers_sent` or so

webknjaz:
@josejulio sorry for late reply, I've added some thoughts on this

josejulio:
> @josejulio sorry for late reply, I've added some thoughts on this

Better late than never :).
I'll address your comments later today, thanks!

cheroot/server.py
josejulio: I though that adding it to `self.server.gateway(self).respond()` would cover it

I suppose I can add it back to ensure we are always sending the headers in case the `gateway` is not sending the headers.

josejulio:
@webknjaz Addressed the comments, travis is green and this is ready for other review round, thanks!

cheroot/server.py
```diff
@@ -972,10 +972,8 @@ def respond(self):
             self.rfile = KnownLengthRFile(self.conn.rfile, cl)
 
         self.server.gateway(self).respond()
+        self.ensure_headers_sent()
```
webknjaz: It looks like you've lost ready check here. Could be:
```python
self.ready and self.ensure_headers_sent()
```

Is this intentional? What is the effect of a such change?

josejulio: Not intentional, missed that, thanks for spotting it!

josejulio:
Awesome, thank you!

PR #76, opened 2018-03-09
Actually use 'reponse' in _conditional_error() method

jonathn:
Bug fix

N/A

Bug was introduced during this refactoring:
  https://github.com/cherrypy/cheroot/commit/a00ad088e75940729a279ebfac9d857012a712fb

The response parameter is never used, instead the hardcoded string '408 Request Timeout' is always sent.

Sends 'response' instead

* **Other information**:

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1202,7 +1202,7 @@ class HTTPConnection(object):
             return
 
         try:
-            req.simple_response('408 Request Timeout')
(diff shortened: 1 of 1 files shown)


webknjaz:
Thanks! May I ask you to also add a test for it please?

webknjaz:
Nevermind, there's no easy way to add test now

PR #80, opened 2018-04-06
Support AF_UNIX in server.HTTPServer.bind_addr

wrmsr:
[This hotfix](https://github.com/cherrypy/cheroot/commit/81abec023851cc9f60443ec3b4d815e83fe9c39d) breaks AF_UNIX binding by leaving only the first two characters of the socket path in bind_addr (as getsockname [returns the socket file path in AF_UNIX](https://docs.python.org/3/library/socket.html#socket-families)). I caught this as my code then goes on to chmod bind_addr expecting it to be a correct path after binding.

This is imo a more urgent fix than the TODO of refactoring out the hack as the hack renders valid usecases broken in 6.1.0 without monkeypatching or equivalent.

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1607,11 +1607,15 @@ class HTTPServer(object):
                 pass
 
         self.socket.bind(self.bind_addr)
-        # TODO: keep separate
-        if family == socket.AF_UNIX:
-            self.bind_addr = self.socket.getsockname()
-        else:
-            self.bind_addr = self.socket.getsockname()[:2]
+        # TODO: keep requested bind_addr separate real bound_addr (port is
+        # different in case of ephemeral port 0)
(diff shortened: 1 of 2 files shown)


webknjaz:
Thanks for pointing this out! It might need a regression test to prevent it from breaking during further refactoring.
I'll try to accept this ASAP once I check that it won't affect anything else additionally, but if you're up to writing that test please let me know.

webknjaz:
@wrmsr do you directly use `bind_address` in your code? how does it break? any tracebacks? I'd like to document more details.

webknjaz:
Looks like this PR broke Windows build: https://ci.appveyor.com/project/CherryPy/cheroot/build/291

/cc: @jaraco

webknjaz:
Ref: #81

Issue #86, opened 2018-04-19
A stand-alone command for running a Cheroot server


scottp-dpaw:
* **I'm submitting a ...**
[ ] bug report
[x] feature request
[ ] question about the decisions made in the repository

* **What is the current behavior?**
Applications wishing to support Cheroot have to provide a loader stub to interface with the generic WSGI application.

* **What is the expected behavior?**
Applications can start a Cheroot instance for a WSGI application from the command line without any additional Python code.

* **What is the motivation / use case for changing the behavior?**
My organisation is in the process of migrating several Python WSGI apps to an automated deployment environment (Red Hat OpenShift + S2I). We would be interested in testing Cheroot as a server backend now it has been decoupled from CherryPy, but we do miss being able to spin up an instance from the command line in the manner of Gunicorn and uWSGI.
In short, having a stand-alone application is useful from a testing and deployment standpoint, as it removes the need to ship a dedicated Cheroot loader for a web app. All that is required is a Python WSGI application exposed as a module. 

* **Other information**
The attached pull request contains a proof of concept script (bin/cheroot) which starts a cheroot.wsgi.Server() based on command line arguments, similar in functionality to the standalone gunicorn and uwsgi programs. Given that this is an early stage I would love feedback on the viability of this proposal, the specific command line arguments, etc. (I am holding off on changing the documentation until the design is approved)

webknjaz:
Resolved by #87.

P.S. @scottp-dpaw please add smth like `Resolves #86` to your PRs and/or commits in future, so that GitHub will automatically close related issues upon merge to master :)

PR #87, opened 2018-04-19
Add stand-alone command for running server

scottp-dpaw:
New feature. This pull request adds the ability to run a new Cheroot instance from the command line, similar to the standalone tools provided by Gunicorn and uWSGI.

#86

Applications wishing to support Cheroot have to provide a loader stub to interface with the generic WSGI application.

Applications can start a Cheroot instance for a WSGI application from the command line without any additional Python code.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/87)


scottp-dpaw:
Thanks for the heads up. I saw contributor notes did mention "ensure you set up pre-commit utility correctly", but I didn't find any detail beyond that. Didn't occur to me there was a tool literally called "pre-commit".

cheroot/__main__.py
```diff
@@ -0,0 +1,103 @@
...
+    if ':' in string:
+        mod_path, app_path = string.split(':', 1)
+    else:
+        mod_path, app_path = string, 'application'
+
+    return mod_path, app_path
+
+
+def wsgi_bind_addr(string):
+    if ':' in string:
```
webknjaz: Mind adding ability to bypassing UNIX sockets via argument?

Here's how curl, httpie and requests do it:
```
curl --unix-socket /tmp/cp.sock http://localhost/
http http+unix://%2Fvar%2Frun%2Fdocker.sock/info
r = requests.get('http+unix://%2Fvar%2Frun%2Fdocker.sock/info')
```

Also, software tends to prefix abstract sockets with `@`:
https://unix.stackexchange.com/questions/206386/what-does-the-symbol-denote-in-the-beginning-of-a-unix-domain-socket-path-in-l?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

So I'd like it to accept unix socket starting with an `@` (or extra argument if you think it's better)

webknjaz:
Also maybe it makes sense to make use of click?

webknjaz:
> Didn't occur to me there was a tool literally called "pre-commit".

Thanks, I've improved that doc a bit now.

setup.py
```diff
@@ -103,6 +103,9 @@
         'Topic :: Internet :: WWW/HTTP :: WSGI :: Server',
     ],
     entry_points={
+        'console_scripts': [
+            'cheroot = cheroot.__main__:main'
```
webknjaz: @jaraco what do you think about this script naming? Do we want it to have trailing `d` to be consistent with CherryPy? Does it need a daemon mode?

cheroot/__main__.py
```diff
@@ -0,0 +1,103 @@
...
+        addr, port = string.split(':', 1)
+        port = int(port)
+    else:
+        addr, port = string, 8000
+    return addr, port
```
webknjaz: @jaraco I've been thinking about having a public struct for the interface in our codebase. What do you think?

It could be:
1. reused across the project
2. support several different concrete socket implementations (and maybe hold some list of features they have):
  - TCP sockets
  - UNIX file sockets (most unix-like OSs)
  - UNIX abstract namespace sockets (GNU/Linux starting v2.2.0 of Linux kernel)

webknjaz:
@scottp-dpaw thanks for your PR :tada:

ping me once linter errors are fixed and you react to comments and I'll get back to review ;)

cheroot/__main__.py
scottp-dpaw: I had a long think about this... the two supported inputs for bind_addr are either the tuple (host_or_ip, port), or a string as a path for a UNIX socket. As the string equivalent for the former ("host_or_ip:port") is pretty easy to match with a regex, it's possible to get away with treating any non-matches as a UNIX socket path. Main difference would be that passing just an IP address would no longer give you port 8000 for free, but chances are if you want to change the IP address you'll be wanting to change the port too.

Alternatively we could put in a case to deal with just a bare IP address so it doesn't make a socket, or require UNIX socket paths begin with '@' and throw an exception as the overall fallback. Also if the plan is to make a new struct type for bind_addr, any string deserializing logic should go in there.

scottp-dpaw:
@webknjaz: should be ready for review again. Is it okay if I ignore the remaining CodeClimate test? 25 lines is a difficult limit for argument parsing, it counts each arg as 5 lines!

I considered third party arg parsing modules; in the end I went with argparse because it didn't add a dependency and worked alright for printing/variable naming/type validation. (Also I once worked on a project which used docopt and almost went mad from trying to debug it)

webknjaz:
// this got folded, so copy-pasting from diff comments:

@jaraco I've been thinking about having a public struct for the interface in our codebase. What do you think?

It could be:
1. reused across the project
1. support several different concrete socket implementations (and maybe hold some list of features they have):
    * TCP sockets
    * UNIX file sockets (most unix-like OSs)
    * UNIX abstract namespace sockets (GNU/Linux starting v2.2.0 of Linux kernel)

setup.py
```diff
@@ -103,6 +103,9 @@
...
+        ]
```
webknjaz: Could you add a trailing comma here?

setup.py
webknjaz: and comma here

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # create a Server object based on the arguments provided
+    server_args = {
+        k: v for k, v in vars(raw_args).items()
+        if (not k.startswith('_')) and v is not None
+    }
+    server_args['wsgi_app'] = wsgi_app_path(raw_args._wsgi_app)
+    server = Server(**server_args)
+
+    try:
```
webknjaz: m/b safe_start? https://github.com/cherrypy/cheroot/blob/master/cheroot/server.py#L1594

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
+"""Command line tool for starting a Cheroot instance.
```
webknjaz: Let's clarify that it's WSGI.

Also, I think it might be good to also support starting plain HTTP thing accepting `cheroot.server.Gateway` instance or class descendant.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+
+Basic usage::
+
+    # Start a server on 127.0.0.1:8000 with the default settings
+    # for the WSGI app myapp/wsgi.py:application()
+    cheroot myapp.wsgi
+
+    # Start a server on 0.0.0.0:9000 with 8 threads
+    # for the WSGI app myapp/wsgi.py:main_app()
+    cheroot myapp.wsgi:main_app --bind 0.0.0.0:9000 --threads 8
```
webknjaz: I'd like to see file socket and abstract socket examples

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+import sys
+
+from .wsgi import Server
+
+RE_HOST_PORT = r'^(\S+)(?::)([0-9]+)$'
+
+
+def wsgi_app_path(string):
+    """Convert WSGI app string to a tuple."""
+    if ':' in string:
```
webknjaz: Refactor:
```python
mod_path, app_path = string.split(':', 1)
if ':' in app_path:
  raise ValueError('Application object name is invalid: it should not contain colon (:) character.')
elif not app_path:
    app_path = 'application'

File: cheroot/__main__.py [hunk 9]
```diff
@@ -0,0 +1,134 @@
...
+        mod_path, app_path = string.split(':', 1)
+    else:
+        mod_path, app_path = string, 'application'
+
+    module = import_module(mod_path)
+    app = eval(app_path, vars(module))
```
webknjaz: I don't tolerate `eval`s in code. Do:
```python
from importlib import import_module

import six

try:
    wsgi_app = getattr(import_module(mod_path), app_path)
    if not callable(wsgi_app):
        raise TypeError
except ImportError as imp_err:
    six.raise_from(NameError('Failed to find a module: {}'.format(mod_path)), imp_err)
except AttributeError as attr_err:
    six.raise_from(NameError('Failed to find application object: {}'.format(app_path)), attr_err)
except TypeError as type_err:
    six.raise_from(TypeError('Application must be a callable object'), type_err)
```

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+import os
+import re
+import sys
+
+from .wsgi import Server
+
+RE_HOST_PORT = r'^(\S+)(?::)([0-9]+)$'
+
+
+def wsgi_app_path(string):
```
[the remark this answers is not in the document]
webknjaz: Turns out that this function returns a WSGI callable, which is why it's better to rename it to `get_wsgi_app`

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    if app is None:
+        raise NameError(
+            'Failed to find application object: {}'.format(app_path))
+    elif not callable(app):
+        raise TypeError('Application must be a callable object')
+
+    return app
+
+
+def wsgi_bind_addr(string):
```
webknjaz: this argument name is pointless

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    """Convert bind address string to a format accepted by Server()."""
+    # try and match for an IP/hostname and port
+    match = re.match(RE_HOST_PORT, string)
+    if match:
+        addr, port = match.groups()
+        port = int(port)
+        return addr, port
+
+    # else, assume a UNIX socket path
+    return string
```
webknjaz: Never return different types of return values from a function.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    elif not callable(app):
+        raise TypeError('Application must be a callable object')
+
+    return app
+
+
+def wsgi_bind_addr(string):
+    """Convert bind address string to a format accepted by Server()."""
+    # try and match for an IP/hostname and port
+    match = re.match(RE_HOST_PORT, string)
```
webknjaz: it's easier to go through `urllib.parse.urlparse('//' + host_port)`, rather then invoking regexp

cheroot/__main__.py
webknjaz: Rename this function when it'll return smth consistent.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+        port = int(port)
+        return addr, port
+
+    # else, assume a UNIX socket path
+    return string
+
+
+def main(args=None):
+    """Run cheroot from the CLI."""
+    # read arguments
```
webknjaz: this comment is pointless.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    if match:
+        addr, port = match.groups()
+        port = int(port)
+        return addr, port
+
+    # else, assume a UNIX socket path
+    return string
+
+
+def main(args=None):
```
webknjaz: what is the point of accepting argument you never pass into?

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    """Run cheroot from the CLI."""
```
webknjaz: This docstring is misleading. This function is already being run invoked from CLI, it doesn't invoke a new one.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+    os.chdir(chdir)
+
+    # add working directory to the path so module importing will resolve
+    if chdir not in sys.path:
+        sys.path.insert(0, chdir)
+
+    # create a Server object based on the arguments provided
+    server_args = {
+        k: v for k, v in vars(raw_args).items()
+        if (not k.startswith('_')) and v is not None
```
webknjaz: this is too magical and making use of underscores does not exactly make sense.

cheroot/__main__.py
scottp-dpaw: My understanding was that the console_script created by setuptools (at one point) used to use it to funnel arguments... today those just seem to edit sys.argv in place, so it's probably redundant.

cheroot/__main__.py
```diff
@@ -0,0 +1,134 @@
...
+import re
+import sys
+
+from .wsgi import Server
+
+RE_HOST_PORT = r'^(\S+)(?::)([0-9]+)$'
+
+
+def wsgi_app_path(string):
+    """Convert WSGI app string to a tuple."""
```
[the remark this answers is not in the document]
scottp-dpaw: The perils of adding docstrings before the refactoring is finished

cheroot/__main__.py
```diff
@@ -114,20 +136,25 @@ def main(args=None):
...
     # create a Server object based on the arguments provided
     server_args = {
-        k: v for k, v in vars(raw_args).items()
-        if (not k.startswith('_')) and v is not None
+        'wsgi_app': wsgi_app,
+        'bind_addr': bind_addr,
     }
-    server_args['wsgi_app'] = wsgi_app_path(raw_args._wsgi_app)
+    for arg, value in vars(raw_args).items():
+        if (not arg.startswith('_')) and (value is not None):
```
webknjaz: We don't abuse parentheses too much in Python.

cheroot/__main__.py
```diff
@@ -114,20 +136,25 @@ def main(args=None):
...
+
     # create a Server object based on the arguments provided
     server_args = {
-        k: v for k, v in vars(raw_args).items()
-        if (not k.startswith('_')) and v is not None
+        'wsgi_app': wsgi_app,
+        'bind_addr': bind_addr,
     }
-    server_args['wsgi_app'] = wsgi_app_path(raw_args._wsgi_app)
+    for arg, value in vars(raw_args).items():
```
webknjaz: It's still too magical as it prevents code analyzers from identifying problems early and breaks debugging experience.

_[Explicit is better than implicit. (c) Zen Of Python](https://zen-of-python.info/explicit-is-better-than-implicit.html#2)_

cheroot/__main__.py
```diff
@@ -9,54 +9,76 @@
...
-    match = re.match(RE_HOST_PORT, string)
-    if match:
-        addr, port = match.groups()
-        port = int(port)
-        return addr, port
+    match = six.moves.urllib.parse.urlparse('//{}'.format(bind_addr_string))
+    try:
+        addr = match.hostname
+        port = match.port
+        if (addr is not None) and (port is not None):
```
webknjaz: Don't abuse parens

webknjaz: _[Beautiful is better than ugly.](https://zen-of-python.info/beautiful-is-better-than-ugly.html#1)_

cheroot/__main__.py
```diff
@@ -65,7 +87,7 @@ def main(args=None):
         help='WSGI application callable to use')
     parser.add_argument(
         '--bind', metavar='ADDRESS',
-        dest='bind_addr', type=wsgi_bind_addr,
+        dest='_bind_addr', type=str,
```
webknjaz: Why not just rely on automatically created property names? Why abuse underscore at all?

cheroot/__main__.py
```diff
@@ -9,54 +9,76 @@
...
+        if (addr is not None) and (port is not None):
+            return addr, port
+    except ValueError:
+        pass
 
     # else, assume a UNIX socket path
-    return string
+    # if the string begins with an @ symbol, use an abstract socket
+    if bind_addr_string.startswith('@'):
+        return '\0{}'.format(bind_addr_string[1:]), None
```
webknjaz: I bet this should be a byte string, shouldn't it?

cheroot/__main__.py
```diff
@@ -9,54 +9,76 @@
...
-        addr, port = match.groups()
-        port = int(port)
-        return addr, port
+    match = six.moves.urllib.parse.urlparse('//{}'.format(bind_addr_string))
+    try:
+        addr = match.hostname
+        port = match.port
+        if (addr is not None) and (port is not None):
+            return addr, port
+    except ValueError:
```
webknjaz: AFAICS this should never happen in such scenario.

cheroot/__main__.py
```diff
@@ -9,54 +9,76 @@
...
 
 
-def wsgi_app_path(string):
-    """Convert WSGI app string to a tuple."""
-    if ':' in string:
-        mod_path, app_path = string.split(':', 1)
+def get_wsgi_app(wsgi_app_path):
+    """Read WSGI app path string and import application module."""
+    if ':' in wsgi_app_path:
+        mod_path, app_path = wsgi_app_path.split(':', 1)
```
webknjaz: You don't need so much nested `if`s. Do split even without check.

cheroot/__main__.py
```diff
@@ -0,0 +1,161 @@
...
+            'Application must be a callable object'), type_err
+        )
+
+    return wsgi_app
+
+
+def parse_wsgi_bind_addr(bind_addr_string):
+    """Convert bind address string to a tuple."""
+    # try and match for an IP/hostname and port
+    match = six.moves.urllib.parse.urlparse('//{}'.format(bind_addr_string))
```
webknjaz: I didn't test this (wrote comment of the top of my head). Did it prove to be working? I'd prefer having tests for it.

setup.py
```diff
@@ -103,6 +103,9 @@
         'Topic :: Internet :: WWW/HTTP :: WSGI :: Server',
     ],
     entry_points={
+        'console_scripts': [
+            'cheroot = cheroot.__main__:main',
```
webknjaz: I'd want to move the implementation code to cli module and then reuse it also in `__main__`, because architecturally speaking both `__main__` and `console_scripts` are the entry points.

cheroot/__main__.py
```diff
@@ -0,0 +1,161 @@
...
+        if (addr is not None) and (port is not None):
+            return addr, port
+    except ValueError:
+        pass
+
+    # else, assume a UNIX socket path
+    # if the string begins with an @ symbol, use an abstract socket
+    if bind_addr_string.startswith('@'):
+        return '\0{}'.format(bind_addr_string[1:]), None
+    return bind_addr_string, None
```
webknjaz: Having plain tuple looks fragile. I'd prefer having dataclass for this or at least a `namedtuple`.

cheroot/__main__.py
scottp-dpaw: 1.) Explicit is better than implicit
2.) The use of underscore for the arguments was to make it possible to package up the majority of arguments from argparse using a simple loop rather than individually.

Basically if I unroll the loop, that will mean at least one if statement per argument to check for the None condition before adding it to the arguments list, which normally doesn't pass my definition of "beautiful is better than ugly". The other option would be to duplicate the class defaults in the argument parser, which hits up against the koan of "don't repeat yourself". Initially I thought the issue was with using the dictionary generator, but what I'm taking away now is that you'd prefer the args manually parsed to the Server() class. I'd like to hear your thoughts on which approach is the best course.

cheroot/__main__.py
scottp-dpaw: The Python socket module accepts both bytes and str, both of which can have \0 in it.

cheroot/cli.py
```diff
@@ -0,0 +1,212 @@
...
+        description='Start an instance of the Cheroot WSGI/HTTP server.')
+    parser.add_argument(
+        '_wsgi_app', metavar='APP_MODULE',
+        type=str,
+        help='WSGI application callable or cheroot.server.Gateway subclass '
+        'to use')
+    parser.add_argument(
+        '--bind', metavar='ADDRESS',
+        dest='_bind_addr', type=str,
+        default='127.0.0.1:8000',
```
jaraco: Default should be IPv6-ready, i.e. '::1:8000'

jaraco: Also, it seems odd to supply two parameters as one (and thus have to parse it).

jaraco:
I'll work on the codeclimate failures - bow to our robot overlords.

webknjaz:
This pull request **introduces 2 alerts** when merging fc4eb350be118157aa5deccfbf792f7aec1df039 into ad4cd1cecb935539413db8b59071cb8c3fed4e8f - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-87a84329e321a34bc5d28968529d8f434385b3a8)

**new alerts:**

* 1 for Missing call to \_\_init\_\_ during object initialization
* 1 for Wrong number of arguments in a call

---

webknjaz:
This pull request **introduces 1 alert** when merging d1c5841310edc2ef3dfb78d2749db87793ade7ab into ad4cd1cecb935539413db8b59071cb8c3fed4e8f - [view on lgtm.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-4a68689716454b315e606e66f6a2a467fcce863e)

**new alerts:**

* 1 for Wrong number of arguments in a call

---

cheroot/__main__.py
```diff
@@ -0,0 +1,6 @@
+"""Stub for accessing the Cheroot CLI tool."""
```
webknjaz: It's not a stub, but an entry point for running CLI

cheroot/cli.py
```diff
@@ -0,0 +1,233 @@
...
+class Application:
+    """Application."""
+
+    @classmethod
+    def resolve(cls, full_path):
+        """Read WSGI app/Gateway path string and import application module."""
+        mod_path, _, app_path = full_path.partition(':')
+        app = getattr(import_module(mod_path), app_path or 'application')
+
+        with contextlib.suppress(TypeError):
```
webknjaz: @jaraco this thing appeared in `3.4`, did I miss adding a compat lib for 2.7?

cheroot/cli.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def __init__(self, addr):
+        """Initialize."""
+        self.bind_addr = '\0{}'.format(self.abstract_socket)
+
+
+class Application:
+    """Application."""
+
+    @classmethod
```
webknjaz: Python... `3.7`?

cheroot/cli.py
webknjaz: Also, what exactly can cause `TypeError` here?

webknjaz: From what I see, it can simply be removed.

cheroot/cli.py
```diff
@@ -0,0 +1,233 @@
...
+    def __init__(self, wsgi_app):
+        """Initialize."""
+        if not callable(wsgi_app):
+            raise TypeError(
+                'Application must be a callable object or '
+                'cheroot.server.Gateway subclass'
+            )
+        self.wsgi_app = wsgi_app
+
+    def server_args(self, parsed_args):
```
webknjaz: This looks like a `@staticmethod` to me

cheroot/cli.py
```diff
@@ -0,0 +1,233 @@
...
+        metavar='APP_MODULE',
+        type=Application.resolve,
+        help='WSGI application callable or cheroot.server.Gateway subclass',
+    ),
+    '--bind': dict(
+        metavar='ADDRESS',
+        dest='bind_addr',
+        type=parse_wsgi_bind_addr,
+        default='[::1]:8000',
+        help='Network interface to listen on (default: [::1]:8000)',
```
webknjaz: Description should tell how to pass UNIX file and abstract sockets.

PR #103, opened 2018-06-20
Fade to black

webknjaz:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [x] refactoring
  - [x] other

N/A

N/A

black style enforced

* **Other information**:
@jaraco here's an idea of adopting black. I'm seeking on feedback, but in general, we would benefit, because contributors won't have to address style quirks, without having to figure out how to manually adjust style.

* **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/103)


jaraco:
meh. Some style changes are good. Other ones are removing meaningful or useful styling. But I guess we should practice deferring to our robot overlords. Do what you like.

webknjaz:
> Other ones are removing meaningful or useful styling.

Yeah, that's what I'm worried about. At least I've disabled this annoying double-quoting of everything and adjusted line-length to PEP8 standard.

jaraco:
I've found that selectively adopting a style system is sometimes worse than not using one at all. Then you have to manage the exclusions/selections plus each user to the project has to know what those are. I've never been a huge fan of flake8 or pep8, but the linter has protected me from many coding errors, which is why I've adopted it. I don't see that black will provide those same benefits. Maybe consider applying black just the once - and cherry-pick the patches to exclude. Then at least the styling will be generally consistent (and provide a template for others working on the project), but we don't have to apply it continuously or enforce violations of it.

webknjaz:
Well, I like 99% of changes, so I think I'm +1 for keeping it.

jaraco:
The consensus is rising on black. I'm happy to adopt it in cheroot/cherrypy too.

webknjaz:
@jaraco I'm still unhappy with certain changes it does. I'm not ready to accept diff-unfriendly changes and I don't like that it does it wholesale (breaking Git blame at least).
I think, style should be chosen consciously by the human author, who knows _why_ they choose to use a certain way of writing those code lines.

jaraco:
So I merged this branch with master, deferring to master for any conflicts, then re-ran `tox -e pre-commit` to re-black the code, but even after, the tests now fail PEP257 and flake8 hooks. I think that outcome reveals at least two issues with this approach.

webknjaz:
I don't like black style after all.

PR #110, opened 2018-08-30
Apply chmod on UNIX socket after creation

### LINKED ISSUE #93: Unix socket mode not set to 0o777
Currently, cheroot deletes any pre-existing AF_UNIX socket file, then attempts to chmod the now non-existent file to 0o777, and only then creates it.

The correct behaviour would be to chmod the socket after creating it.

webknjaz:
Fixes #93

  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#93 

code unlinks the socket file and then immediately tries to chmod it

it should chmod after bind

* **Other information**:
Very first appearance: https://github.com/cherrypy/cheroot/commit/6ced9fd2f7d23c9e4db78894d616ed8cc5e83738#diff-920de15fced9e5f9697efed08f3c207dR214

* **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/110)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1660,26 +1660,46 @@ class HTTPServer:
 
         # Select the appropriate socket
         self.socket = None
+        msg = 'No socket could be created'
         if os.getenv('LISTEN_PID', None):
             # systemd socket activation
             self.socket = socket.fromfd(3, socket.AF_INET, socket.SOCK_STREAM)
         elif isinstance(self.bind_addr, six.string_types):
             # AF_UNIX socket
 
+            if IS_WINDOWS:
+                # Trying to access socket.AF_UNIX under Windows
+                # causes an AttributeError
+                raise ValueError(  # or RuntimeError?
+                    'AF_UNIX sockets are not supported under Windows.'
+                )
+
             # So we can reuse the socket...
             try:
                 os.unlink(self.bind_addr)
-            except Exception:
-                pass
+            except OSError:
+                """
+                File does not exist, which is the primary goal anyway.
+                """
+
+            # TODO: call bind here
+            af, socktype, proto, sa = (
+                socket.AF_UNIX, socket.SOCK_STREAM, 0, self.bind_addr
+            )
+            try:
+                self.bind(af, socktype, proto)
+            except socket.error as serr:
+                msg = '%s -- (%s: %s)' % (msg, sa, serr)
+                if self.socket:
+                    self.socket.close()
+                self.socket = None
+                six.raise_from(socket.error(msg), serr)
 
             # So everyone can access the socket...
             try:
-                os.chmod(self.bind_addr, 0o777)
-            except Exception:
+                os.chmod(self.bind_addr, 0o777)  # TODO: allow changing mode
+            except OSError:
                 pass
-
-            info = [
-                (socket.AF_UNIX, socket.SOCK_STREAM, 0, '', self.bind_addr)]
         else:
             # AF_INET or AF_INET6 socket
             # Get the correct address family for our host (allows IPv6
@@ -1699,8 +1719,6 @@ class HTTPServer:
 
                 info = [(sock_type, socket.SOCK_STREAM, 0, '', bind_addr)]
 
-        if not self.socket:
-            msg = 'No socket could be created'
             for res in info:
                 af, socktype, proto, canonname, sa = res
                 try:
@@ -1712,8 +1730,8 @@ class HTTPServer:
                         self.socket.close()
                     self.socket = None
 
-            if not self.socket:
-                raise socket.error(msg)
+        if not self.socket:
+            raise socket.error(msg)
 
         # Timeout so KeyboardInterrupt can be caught on Win32
         self.socket.settimeout(1)
@@ -1806,6 +1824,7 @@ class HTTPServer:
                 pass
 
         self.socket.bind(self.bind_addr)
+
         # TODO: keep requested bind_addr separate real bound_addr (port is
         # different in case of ephemeral port 0)
         self.bind_addr = self.socket.getsockname()
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1791,20 +1791,32 @@ class HTTPServer:
 
     def bind(self, family, type, proto=0):
         """Create (or recreate) the actual socket object."""
-        self.socket = socket.socket(family, type, proto)
-        prevent_socket_inheritance(self.socket)
+        sock = self.prepare_socket(
+            self.bind_addr,
+            family, type, proto,
+            self.nodelay, self.ssl_adapter,
+        )
+        sock = self.socket = self.bind_socket(sock, self.bind_addr)
+        self.bind_addr = self.resolve_real_bind_addr(sock)
+        return sock
+
+    @staticmethod
+    def prepare_socket(bind_addr, family, type, proto, nodelay, ssl_adapter):
+        """Create and prepare the socket object."""
+        sock = socket.socket(family, type, proto)
+        prevent_socket_inheritance(sock)
         if not IS_WINDOWS:
             # Windows has different semantics for SO_REUSEADDR,
             # so don't set it.
             # https://msdn.microsoft.com/en-us/library/ms740621(v=vs.85).aspx
-            self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        if self.nodelay and not isinstance(self.bind_addr, str):
-            self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if nodelay and not isinstance(bind_addr, str):
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
 
-        if self.ssl_adapter is not None:
-            self.socket = self.ssl_adapter.bind(self.socket)
+        if ssl_adapter is not None:
+            sock = ssl_adapter.bind(sock)
 
-        host, port = self.bind_addr[:2]
+        host, port = bind_addr[:2]
 
         # If listening on the IPV6 any address ('::' = IN6ADDR_ANY),
         # activate dual-stack. See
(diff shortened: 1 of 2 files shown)


cheroot/server.py
```diff
@@ -1805,7 +1793,22 @@ def bind(self, family, type, proto=0):
                 # this machine's TCP stack
                 pass
 
+        # So we can reuse the socket...
+        try:
+            if not IS_WINDOWS and family == socket.AF_UNIX:
```
jaraco: Can we omit the `IS_WINDOWS` check? It feels like a redundant check, especially after having constructed a `socket.socket(AF_UNIX)`. It's also conceivable (though unlikely) that Windows could support this socket type. Better to just let operations fail, I think.

cheroot/server.py
```diff
@@ -1805,7 +1793,22 @@ def bind(self, family, type, proto=0):
                 # this machine's TCP stack
                 pass
 
+        # So we can reuse the socket...
```
jaraco: I dislike how this functionality is now embedded in the bind method and straddling the bind call.

I'm inclined to leave the unlink operation in prepare socket.

cheroot/server.py
```diff
@@ -1805,7 +1793,22 @@ def bind(self, family, type, proto=0):
...
+                os.unlink(self.bind_addr)
+        except Exception:
+            pass
+
         self.socket.bind(self.bind_addr)
+
+        # So everyone can access the socket...
+        try:
+            if not IS_WINDOWS and family == socket.AF_UNIX:
+                os.chmod(self.bind_addr, 0o777)
```
jaraco: Is 0o777 the right permissions to set unconditionally? Why not let the os set the permissions naturally, and leave it to the client to change the permissions if needed?

cheroot/server.py
webknjaz: As I recall accessing `socket.AF_UNIX` causes `AttributeError` under Windows, that's why this guard exists. Windows doesn't have such a concept at all.

cheroot/server.py
webknjaz: It guarantees that everyone can access it. Also, this PR is not intended to change the behavior. Only fix a mistake.

cheroot/server.py
webknjaz: Please take a look at this comment of the reporter: https://github.com/cherrypy/cheroot/issues/93#issuecomment-417620472
It looks like these things are quite tightly coupled. Also, it's needed exactly around `socket.bind()`.

What I don't like even more is that bypassing of the bind_addr is done via implicit assignment to the self object attribute.

webknjaz:
I'm not sure about complete removal of chmod recognizing that most of (or all?) other HTTP servers allow bypassing the config with these access privileges, while have other sufficient default. Also, I'm not sure about what side-effects this removal may cause.

webknjaz:
This pull request **introduces 1 alert** when merging d2fbe492d2a6717ea879e76927859fa16c62200c into 5edf9cb1cd4ebc8775971909a037c4d9dcd6688b - [view on LGTM.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-c1e36d2659b4d5f67dc4c72ac888344ac1479649)

**new alerts:**

* 1 for Wrong number of arguments in a call

---

*Comment posted by [LGTM.com](https://lgtm.com)*

webknjaz:
This pull request **introduces 1 alert** when merging bc58cdb924ab6d15b0e1fe2fce194683ef0957ef into 5edf9cb1cd4ebc8775971909a037c4d9dcd6688b - [view on LGTM.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-b9a1743e504abca0fdec880bbd2802e235601b52)

**new alerts:**

* 1 for Unused local variable

---

*Comment posted by [LGTM.com](https://lgtm.com)*

webknjaz:
@jaraco @hvenev I've redesigned this PR a bit. Could you please take a look?

cheroot/server.py
```diff
@@ -1773,20 +1762,93 @@ def error_log(self, msg='', level=20, traceback=False):
...
+
+        try:
+            # Make possible reusing the socket...
+            os.unlink(self.bind_addr)
+        except OSError:
+            """
+            File does not exist, which is the primary goal anyway.
+            """
+
+        sock = self.prepare_socket(
```
webknjaz: TODO: try setting `os.uname` before this call.

cheroot/server.py
```diff
@@ -1773,20 +1762,93 @@ def error_log(self, msg='', level=20, traceback=False):
...
+        if nodelay and not isinstance(bind_addr, str):
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
 
-        if self.ssl_adapter is not None:
-            self.socket = self.ssl_adapter.bind(self.socket)
+        if ssl_adapter is not None:
+            sock = ssl_adapter.bind(sock)
 
-        host, port = self.bind_addr[:2]
+        host, port = bind_addr[:2]
```
hvenev: I'm not sure these make sense for AF_UNIX. In fact, most of this function should probably be moved to the TCP/IP `bind()`.

cheroot/server.py
```diff
@@ -1773,20 +1762,93 @@ def error_log(self, msg='', level=20, traceback=False):
...
-        if self.nodelay and not isinstance(self.bind_addr, str):
-            self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if nodelay and not isinstance(bind_addr, str):
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
 
-        if self.ssl_adapter is not None:
-            self.socket = self.ssl_adapter.bind(self.socket)
+        if ssl_adapter is not None:
+            sock = ssl_adapter.bind(sock)
```
hvenev: I'm not sure about this, but we might want to use the SSL adapter for systemd-provided sockets. If so, maybe this should be moved to `prepare_socket()`.

cheroot/server.py
```diff
@@ -1773,20 +1762,93 @@ def error_log(self, msg='', level=20, traceback=False):
...
+        prevent_socket_inheritance(sock)
         if not IS_WINDOWS:
             # Windows has different semantics for SO_REUSEADDR,
             # so don't set it.
             # https://msdn.microsoft.com/en-us/library/ms740621(v=vs.85).aspx
-            self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        if self.nodelay and not isinstance(self.bind_addr, str):
-            self.socket.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if nodelay and not isinstance(bind_addr, str):
```
hvenev: Maybe check against `six.string_types` similarly to `prepare()`?

cheroot/server.py
hvenev: Do you mean `umask` or changing the system hostname?

webknjaz: Yeah, but maybe we can grab current umask, create a socket and then unset it. I'll probably not do it for now.

cheroot/server.py
webknjaz: It's a good question. I'd rather check `family` though. I just didn't want to change much.

cheroot/server.py
webknjaz: That's a fair point, I prefer to think about it outside of this PR to keep changes small.

cheroot/server.py
webknjaz: Same, I don't want to touch it now. That's also further plans for refactoring it and turning it into some struct.

cheroot/server.py
```diff
@@ -1660,26 +1660,17 @@ def prepare(self):
...
-            # So everyone can access the socket...
             try:
-                os.chmod(self.bind_addr, 0o777)
-            except Exception:
-                pass
-
-            info = [
-                (socket.AF_UNIX, socket.SOCK_STREAM, 0, '', self.bind_addr)]
+                self.bind_unix_socket(self.bind_addr)
+            except socket.error as serr:
```
jaraco: This error handling is asymmetric to how a socket error is handled for AF_INET*.

cheroot/server.py
```diff
@@ -1798,18 +1860,28 @@ def bind(self, family, type, proto=0):
...
+        return sock
+
+    @staticmethod
+    def bind_socket(socket_, bind_addr):
+        """Bind the socket to given interface."""
+        socket_.bind(bind_addr)
+        return socket_
+
+    @staticmethod
+    def resolve_real_bind_addr(socket_):
```
jaraco: Why not use `sock`, which is used elsewhere in this module for the `socket.socket` instance?

cheroot/server.py
```diff
@@ -1773,20 +1762,93 @@ def error_log(self, msg='', level=20, traceback=False):
...
+        if IS_WINDOWS:
+            """
+            Trying to access socket.AF_UNIX under Windows
+            causes an AttributeError.
+            """
+            raise ValueError(  # or RuntimeError?
+                'AF_UNIX sockets are not supported under Windows.'
+            )
+
+        fs_permissions = 0o777  # TODO: allow changing mode
```
jaraco: I'm still uneasy with this as a default. I think the default should be not to change any permissions from whatever the OS sets.

cheroot/server.py
webknjaz: What do you mean? Moving `socket.close()` inside the method? I hope to generalize this stuff sometime in future.

cheroot/server.py
webknjaz: All servers I can recall have 0o666 as default permissions.
Also, take into account that if the permissions don't have rw, then nobody will be able to use this socket and our goal is to have it accessible by default. See `man -s 7 unix` for details:
```

   Pathname socket ownership and permissions
       In the Linux implementation, pathname sockets honor the permissions of the directory they are in.  Creation of a new
       socket fails if the process does not have write and search (execute) permission on the directory in which the socket
       is created.

       On Linux, connecting to a stream socket object requires write permission on that socket; sending  a  datagram  to  a
       datagram  socket  likewise  requires  write  permission on that socket.  POSIX does not make any statement about the
       effect of the permissions on a socket file, and on some systems (e.g.,  older  BSDs),  the  socket  permissions  are
       ignored.  Portable programs should not rely on this feature for security.

       When creating a new socket, the owner and group of the socket file are set according to the usual rules.  The socket
       file has all permissions enabled, other than those that are turned off by the process umask(2).

       The owner, group, and permissions of a pathname socket can be changed (using chown(2) and chmod(2)).
```

cheroot/server.py
webknjaz: Because it's a function signature, more or less public interface. So I restricted usage of shortcuts inside functions only,

PR #116, opened 2018-09-20
cheroot #42 -- remove unittest.testcase and ddt dependencies from unit tests

jeffvanvoorst:
  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [X] tests/coverage improvement
  - [X] refactoring
  - [ ] other

#42

Tests rely on unittest.testcase and use `ddt`

Remove dependencies on `ddt` and `unittest.testcase`

* **Other information**:
So far, I haven't been able to use the WebTest.TestApp (or maybe it is wsgiproxy.HostProxy) without having a WSGI app loaded.  I am either calling an invalid URL or the server is accessed differently than the previous test code.  The exception is basically that "None doesn't have a start_response".

Need to remove stylistic differences and cleanup files once the changes are agreeable to the cheroot maintainers.

I am not sure how to run the code through CI/automated processes -- maybe I need a branch directly from the cheroot master?

* **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/116)


cheroot/test/conftest.py
```diff
@@ -25,3 +30,76 @@ def wsgi_server_client(wsgi_server):
 def native_server_client(native_server):
     """Create a test client out of given HTTP server."""
     return get_server_client(native_server)
+
+
+@pytest.fixture
+def make_session_for_cheroot(wsgi_server):
+    '''
```
webknjaz: Use PEP257 for docstrings

cheroot/test/conftest.py
```diff
@@ -25,3 +30,76 @@ def wsgi_server_client(wsgi_server):
 def native_server_client(native_server):
     """Create a test client out of given HTTP server."""
     return get_server_client(native_server)
+
+
+@pytest.fixture
+def make_session_for_cheroot(wsgi_server):
```
webknjaz: Can this be useful for end-users of cheroot in the testing of their apps? If yes, then it should go to `cheroot.testing` module.

cheroot/test/test_more.py
```diff
@@ -0,0 +1,57 @@
+import os
+import pytest
+import requests
+import ssl
+from cheroot.ssl.builtin import BuiltinSSLAdapter
+
+def ssl_file(filename):
```
webknjaz: Can you use `trustme` for SSL certificates creation/testing?

jeffvanvoorst: I think that is relatively straightforward to do.

webknjaz: Yeah, please do

webknjaz:
@jeffvanvoorst you can find linter errors in CI @ https://travis-ci.org/cherrypy/cheroot/jobs/431016556#L483
Also, you can run `tox -e pre-commit` for those locally.

webknjaz:
Hint: you can enable tests in your fork here https://travis-ci.com/jeffvanvoorst/cheroot and then you'll be able to restart tests there by yourself if needed :)

.coveragerc
```diff
@@ -0,0 +1,5 @@
+# .coveragerc to control coverage.py
```
webknjaz: We don't need this

cheroot/test/test_errors.py
```diff
@@ -12,7 +13,7 @@
         (
             ('EPROTOTYPE', 'EAGAIN', 'EWOULDBLOCK',
              'WSAEWOULDBLOCK', 'EPIPE'),
-            [91, 11, 32]
+            [32, 10041, 11, 10035] if os.name == 'nt' else [91, 11, 32]
```
webknjaz: This must go to a separate PR.

webknjaz: (and use `system.platform()` instead)

cheroot/test/test_more.py
```diff
@@ -0,0 +1,90 @@
...
+    cert_type = request.param
+    cert_type_to_args = {
+        "client":            ["localhost", "127.0.0.1", "::1"],
+        "client_ip":         ["127.0.0.1", "::1"],
+        "client_wildcard":   ["*.localhost"],
+        "client_wrong_host": ["cherrypy.org", "github.com"],
+    }
+
+    client_cert = None
+    if cert_type == "client_wrong_ca":
```
webknjaz: There's too much logic in the test case, it must go to separate tests, which should be as simple as possible.

cheroot/test/test_more.py
```diff
@@ -0,0 +1,90 @@
...
+    my_session = make_session_for_cheroot(
+        "localhost", prefix="https://", ssl_adapter=SSLAdapter, test_app_client="requests",
+        client_cert=client_cert_files.cert, client_key=client_cert_files.pkey)
+
+    # The pyriform requests adaptor eats the request verify and cert args as it isn't trivial to pass
+    # such args to an instant of WebTest.TestApp.
+    # Therefore, although the verify and cert args are set here, they are effectively ignored.
+    resp = my_session.get('https://localhost/', verify=SSLAdapter.certificate_chain,
+        cert=(client_cert_files.cert, client_cert_files.pkey))
+    if allowed is True:
```
webknjaz: `is True` is redundant. Also, I don't like 2in1 tests

.coveragerc
jeffvanvoorst: not a problem -- my intention was to not have the unit test sources be considered when calculating code coverage.   However, if one looks more at the detailed list  and less at the over all number, then the information is approximately the same either way (implicitly versus explicitly ignoring the test files).

cheroot/test/test_more.py
jeffvanvoorst: Please explain the idea: 2in1 tests.  Is one parameterization enough?  
For example, could go very complex and test all client cert categories, 3 types of client certificate checking (from server point of view), and 2 servers from one test function.  
The other extreme is to have one test for each combination.
Which level of complexity/code sharing are you recommending?

cheroot/test/conftest.py
webknjaz: Well, `pyriform` is should help testing it without starting an HTTP server.

jeffvanvoorst: My understanding is that this current fixture is for starting an HTTP server.

I had to replace certain parts of the TestApp and HostProxy in order to get this to work.  Also there is two sets of requests -- one used by HostProxy and one by Pyriform itself.

jeffvanvoorst: Could probably think about how to add a Pyriform fixture for cherrypy.  Not sure how to add one for cheroot with a different framework.

webknjaz: What do you mean by saying that? cheroot with a different framework?

cheroot/test/conftest.py
```diff
@@ -25,3 +28,94 @@ def wsgi_server_client(wsgi_server):
 def native_server_client(native_server):
     """Create a test client out of given HTTP server."""
     return get_server_client(native_server)
+
+
+@pytest.fixture  # noqa: F811
```
webknjaz: What do you ignore here?

cheroot/testing.py
```diff
@@ -42,7 +52,8 @@ def cheroot_server(server_factory):
         try:
             actual_bind_addr = (interface, bind_port)
             httpserver = server_factory(  # create it
-                bind_addr=actual_bind_addr,
+                actual_bind_addr,
+                conf.pop('wsgi_app', None),
```
webknjaz: please use keyword args.

cheroot/testing.py
```diff
@@ -21,10 +22,19 @@
 ANY_INTERFACE_IPV4 = '0.0.0.0'
 ANY_INTERFACE_IPV6 = '::'
 
+
+def my_crazy_app(environ, start_response):
```
webknjaz: I'm not sure that it belongs here. How can this be useful for testing apps by end-users?

cheroot/test/test_https.py
```diff
@@ -2,146 +2,163 @@
...
+        # WebTest.TestApp.  Therefore, although the verify and cert args are
+        # set here, they are effectively ignored.
+        resp = my_session.get(
+            'https://localhost/',
+            verify=SSLAdapter.certificate_chain,
+            cert=(client_cert_files.cert, client_cert_files.pkey))
+        return resp
+
+    def expected_ssl_failure_status_code(self):
+        if six.PY3:
```
webknjaz: Why is this different?

cheroot/test/test_https.py
```diff
@@ -2,146 +2,163 @@
...
+    client_dir = tmpdir.mkdir('client')
+
+    client_cert_file = client_dir.join('client.cert')
+    client_cert_file.write(client_cert.cert_chain_pems[0].bytes())
+    client_pkey_file = client_dir.join('server_pkey')
+    client_pkey_file.write(client_cert.private_key_pem.bytes())
+    return CertKeyPair(str(client_cert_file), str(client_pkey_file), cert_type)
+
+
+class _TestClientCertValidation:
```
webknjaz: Why do you need a class?

jeffvanvoorst:
Not trying to be obstinate.  The part that I am not understanding is how to use the Pyriform adapter to a real cheroot server to help users.  My guess is that users would want to test their framework code and not the real webserver.  The proxy/test code will be different.

jaraco:
I haven't spent enough time reviewing this to give it a meaningful review. The changes seem reasonable and genuine improvements. Can we say with confidence that this change won't break cherrypy's tests? If so, I'm almost inclined to accept it once @webknjaz is satisfied with the responses to his concerns.

webknjaz:
@jeffvanvoorst this needs remaking because I've factored out the unittest and ddt stuff already so only testing facilitation left to do.

Issue #138, opened 2018-11-09
Google App Engine - ImportError: No module named grp 


zippy84:
I'm trying to run CherryPy in Googles App Engine. Because I only can use Python 2.7 I installed an older version of CherryPy with pip install -t /home/zippy/ecl/ecl_cherrypy/lib "CherryPy==17.4.0". 

And now I get:

```
Traceback (most recent call last):
  File "/home/zippy/ecl/google_appengine/google/appengine/runtime/wsgi.py", line 240, in Handle
    handler = _config_handle.add_wsgi_middleware(self._LoadHandler())
  File "/home/zippy/ecl/google_appengine/google/appengine/runtime/wsgi.py", line 299, in _LoadHandler
    handler, path, err = LoadObject(self._handler)
  File "/home/zippy/ecl/google_appengine/google/appengine/runtime/wsgi.py", line 85, in LoadObject
    obj = __import__(path[0])
  File "/home/zippy/ecl/ecl_cherrypy/ecl.py", line 9, in <module>
    import cherrypy
  File "/home/zippy/ecl/ecl_cherrypy/lib/cherrypy/__init__.py", line 76, in <module>
    from . import _cprequest, _cpserver, _cptree, _cplogging, _cpconfig
  File "/home/zippy/ecl/ecl_cherrypy/lib/cherrypy/_cprequest.py", line 13, in <module>
    from cherrypy import _cpreqbody
  File "/home/zippy/ecl/ecl_cherrypy/lib/cherrypy/_cpreqbody.py", line 135, in <module>
    import cheroot.server
  File "/home/zippy/ecl/ecl_cherrypy/lib/cheroot/server.py", line 102, in <module>
    import grp
  File "/home/zippy/ecl/google_appengine/google/appengine/tools/devappserver2/python/runtime/sandbox.py", line 1095, in load_module
    raise ImportError('No module named %s' % fullname)
ImportError: No module named grp
```
Could you say what's wrong with it?

`python2 -c "import grp"` works fine.

zippy84:
> I think, as a workaround, pinning Cheroot<6.2.0 should temporarily fix the issue for you until it's fixed in upstream.

You are right.

`pip install -t /home/zippy/ecl/ecl_cherrypy/lib "CherryPy==11.0.0"` is not a solution, because the problem is located in Cheroot.

I already find a solution by downgrading Cheroot to 5.8.3.

webknjaz:
[...]

PR #140, opened 2018-11-11
Fix for the use in GAE.

zippy84:
  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] refactoring
  - [ ] other

#138 

`ImportError` on `import grp` and `import pwd`, because they are not available on Google App Engine.

No `ImportError` in Google App Engine :smile:

* **Other information**:

* **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/140)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -99,8 +99,12 @@ IS_WINDOWS = platform.system() == 'Windows'
 
 
 if not IS_WINDOWS:
-    import grp
-    import pwd
+    try:
+        import grp
+        import pwd
+    except ImportError:
+        """Disabled by Google App Engine (GAE)."""
+        pass
     import struct
 
 
@@ -1375,7 +1379,7 @@ class HTTPConnection:
             raise NotImplementedError(
                 'UID/GID lookup can only be done under UNIX-like OS'
             )
-        elif not self.peercreds_resolve_enabled:
+        elif not self.peercreds_resolve_enabled or 'grp' not in globals() or 'pwd' not in globals():
             raise RuntimeError(
                 'UID/GID lookup is disabled within this server'
             )
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -96,14 +96,33 @@ __all__ = ('HTTPRequest', 'HTTPConnection', 'HTTPServer',
 
 
 IS_WINDOWS = platform.system() == 'Windows'
+"""Flag indicating whether the app is running under Windows."""
 
 
-if not IS_WINDOWS:
+IS_GAE = os.getenv('SERVER_SOFTWARE', '').startswith('Google App Engine/')
+"""Flag indicating whether the app is running in GAE env.
+
+Ref:
+https://cloud.google.com/appengine/docs/standard/python/tools
+/using-local-server#detecting_application_runtime_environment
+"""
+
+
+IS_UID_GID_RESOLVABLE = not IS_WINDOWS and not IS_GAE
+"""Indicates whether UID/GID resolution's available under current platform."""
+
+
+if IS_UID_GID_RESOLVABLE:
     try:
         import grp
         import pwd
     except ImportError:
-        """Disabled by Google App Engine (GAE)."""
+        """Unavailable in the current env.
+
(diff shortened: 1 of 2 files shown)


webknjaz:
This pull request **introduces 1 alert** when merging 06d65eb7565effe2bd51e3d94e75ae5e10751c93 into ca85772e6e58a880657a9c26a5352603535517f5 - [view on LGTM.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-1297a26fcd109161558989e4149e19f120c27aa7)

**new alerts:**

* 1 for Unnecessary pass

---

*Comment posted by [LGTM.com](https://lgtm.com)*

cheroot/test/test_server.py
```diff
@@ -186,8 +186,14 @@ def test_peercreds_unix_sock(http_server, unix_sock_file):
...
     assert 'RuntimeError' in testclient.get(PEERCRED_TEXTS_URI)
 
-    httpserver.peercreds_resolve_enabled = True
-    import grp
-    expected_textcreds = os.getlogin(), grp.getgrgid(os.getgid()).gr_name
-    expected_textcreds = '!'.join(map(str, expected_textcreds))
-    assert testclient.get(PEERCRED_TEXTS_URI) == expected_textcreds
+    try:
+        import grp
+    except ImportError:
```
webknjaz: We should use `skipif` marker on the test instead of having a guessing logic inside of it since it's unrelated. You can create a named marker like `non_windows_sock_test` is.

cheroot/server.py
```diff
@@ -1375,7 +1379,7 @@ def resolve_peer_creds(self):  # LRU cached on per-instance basis
             raise NotImplementedError(
                 'UID/GID lookup can only be done under UNIX-like OS'
             )
-        elif not self.peercreds_resolve_enabled:
+        elif not self.peercreds_resolve_enabled or 'grp' not in globals() or 'pwd' not in globals():
```
webknjaz: This violates the linters: https://travis-ci.org/cherrypy/cheroot/jobs/453609030

cheroot/test/test_server.py
zippy84: I'm not familiar with pytest. The whole `test_peercreds_unix_sock` should be skipped, when `grp` is not importable? I don't think so. But this is your decision.

cheroot/server.py
zippy84: Is this better?

```python
try:
    import grp
except ImportError:
    grp = None
```

And then later

```python
if grp is None:
    raise RuntimeError('...')
```
Do you have any reference for such a design pattern? There is no PEP about it.

cheroot/test/test_server.py
webknjaz: Well, it should probably be separated into its own test. Actually, currently, this one is completely disabled, because the test client is unable to HTTP over non-IP connection (yet).

As for `pytest`, you can have a decorator like `@pytest.mark.skipif(<condition_here>, reason="some explanation"`, which would conditionally skip the test.

cheroot/server.py
webknjaz: P.S. you can also add some boolean var with a proper name as a flag for whether grp/pwd are accessible, this tends to make code better readable.

cheroot/test/test_server.py
zippy84: I think you should change it to your own.

cheroot/server.py
zippy84: Too late. Please change it on your own preference, if you don't like the actual changes.

cheroot/test/test_server.py
webknjaz: ok

webknjaz:
@zippy84 I'll merge this after your confirmation that everything looks okay.

zippy84:
Your last changes looks good to me. It works perfect with CherryPy 17.4.0 and Google App Engine 1.9.80.

webknjaz:
I think it's a good balance for optimizations vs. heavy imports

PR #146, opened 2018-12-04
Fix Circle CI builds

duper51:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [X] 📋 refactoring
  - [ ] 💥 other

Fix up the tests failing on Travis.

Travis builds fail due to a PEP compliance error.

They no longer fail.

📋 **Other information**:

CircleCI builds as well as `pre-commit-failing` builds fail due to an underlying issue with Asteroid. They currently have no builds that work with both Python 2.7 and Python 3.7. As such we might need to rework our testing strategy on that front, as it seems they don't have a build for Python 3.7 yet either.

📋 **Checklist**:

  - [X] I have [squashed related commits together][2] after the changes have been approved

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/146)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,14 @@ jobs:
 
     - run: pip install tox tox-pyenv
     - checkout
-    - run: tox -e py27,py34,py35,py36,py37,pypy3 -- -p no:sugar
+    - run:
+        name: Run tests
+        command: tox -e py27,py34,py35,py36,py37,pypy3 -- -p no:sugar
+        # Environment variables for py-cryptography library
+        environment:
+          LDFLAGS: "-L/usr/local/opt/openssl/lib"
+          CPPFLAGS: "-I/usr/local/opt/openssl/include"
+          PKG_CONFIG_PATH: "/usr/local/opt/openssl/lib/pkgconfig"
 
   linux-build:
     parallelism: 4
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ whitelist_externals =
 commands =
     rm -rf .eggs/
     bash -c "if [[ '{env:CIRCLECI:disabled}' == 'disabled' ]]; then pytest --testmon-off {posargs}; fi"
-    bash -c "if [[ '{env:CIRCLECI:disabled}' != 'disabled' ]]; then circleci tests glob **/test/**.py | circleci tests split --split-by=timings | xargs pytest --testmon-off {posargs}; fi"
+    bash -c "if [[ '{env:CIRCLECI:disabled}' != 'disabled' ]]; then circleci tests glob **/test/**.py | circleci tests split --split-by=timings | grep -v '__init__.py' | xargs pytest --testmon-off {posargs}; fi"
     codecov -f coverage.xml -X gcov
 usedevelop = True
 extras = testing


duper51:
Bad news, I had another commit worked in here, I'll deal with that and fix it up.

.pre-commit-config.yaml.failing
```diff
@@ -3,6 +3,7 @@ repos:
   rev: v1.9.1
   hooks:
   - id: pylint
+    language_version: python2.7
```
webknjaz: I'd want it to run it under Python 3.7 or any currently active python.

Could you rewrite it like this? https://github.com/sanitizers/chronographer-github-app/blob/master/.pre-commit-config.yaml#L50-L59
pylint is a special case: it makes not only static but also semi-dynamic checks like checking whether things are imporatable which is impossible from a separate isolated virtualenv pre-commit tool creates.

.github/ISSUE_TEMPLATE.md
```diff
@@ -43,4 +43,3 @@ https://stackoverflow.com/questions/tagged/cheroot+or+cherrypy
 - Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
 
 📋 **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)
-
```
webknjaz: I think this change belongs to a separate PR. We value atomic changes (for both PRs and individual commits) as it makes Git history clean and better usable.

Here's some articles explaining best ways to write commit messages:
  http://chris.beams.io/posts/git-commit/
  http://alistapart.com/article/the-art-of-the-commit
  https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message
  https://github.com/erlang/otp/wiki/writing-good-commit-messages
  https://wiki.openstack.org/wiki/GitCommitMessages
  http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
  https://dev.to/gonedark/when-to-make-a-git-commit

.pre-commit-config.yaml.failing
duper51: I just found that they added compatibility for the 3.7 in pytest 3.0.1, I'm testing to see if it will work back to 2.7. If not, I'll look into using that commit :)

webknjaz: Well, the version of pytest used currently is 4 so it must be ok

.github/ISSUE_TEMPLATE.md
duper51: That got placed in there by a modification by tox. Albeit, I've not worked with the framework very much so I had assumed it wouldn't modify my files unless I told it to. I would agree, moving that elsewhere would be beneficial.

.pre-commit-config.yaml.failing
duper51: That's what I'm curious about. I just noticed it was running on 4, which is way past the [commit](https://docs.pytest.org/en/latest/changelog.html#id553) where they fixed the issue, yet there is still a StopIteration laying around in the source.

.github/ISSUE_TEMPLATE.md
webknjaz: Ah, it's probably done by `pre-commit` tool :) One of the hooks does it.

.pre-commit-config.yaml.failing
webknjaz: But regarding commit messages: you could improve wording there + it looks like you're committing from unconfigured computer. The email you use there doesn't match one bound to your GitHub account, that's why they ain't linked to you.

webknjaz: Oh.. And it doesn't fail hard in travis because it still uses Python 3.6 for linter jobs...

webknjaz: Or not. It's interesting: this fails during tests collection but not in Travis (even under Python 3.7). Can it be Travis CI cache using different version of (indirect?) dependency for some pytest plugin, not pytest itselff?

duper51: Huh, check that out. I'll work on that and push out a fix. I'd really like to have the tests be useful again, especially for my own purposes.

webknjaz: Sure, thanks!

It looks like this line is causing `StopIteration` being raised: https://github.com/pytest-dev/pytest/blob/master/src/_pytest/main.py#L586. Which probably means that object being empty:
```python
>>> next(iter([]))
StopIteration
```

This may be caused by `testmon` plugin if it's active. Or... Maybe it's because CircleCI's CLI command returns empty list (it's used to distribute tests across 4 parallel jobs).

P.S. We mostly rely on Travis CI currently so broken Circle CI is not a major issue.

webknjaz: Oh, there's `--testmon-off` so it's definitely not it, only Circle CI client left to be checked.

webknjaz: It's probably related to env in that docker image used in Circle CI. You can test it locally.

webknjaz:
I've found a stable reproducer:
```console
$ .tox/python/bin/pytest --testmon-off cheroot/test/__init__.py
Test session starts (platform: linux, Python 3.7.0, pytest 4.0.1, pytest-sugar 0.9.2)
cachedir: .pytest_cache
rootdir: /home/wk/src/github/cherrypy/cheroot, inifile: pytest.ini
plugins: testmon-0.9.14, sugar-0.9.2, cov-2.6.0, backports.unittest-mock-1.4
collecting ... 
―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― ERROR collecting  ――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
.tox/python/lib/python3.7/site-packages/_pytest/main.py:586: in _collect
    yield next(m[0].collect())
E   StopIteration

The above exception was the direct cause of the following exception:
.tox/python/lib/python3.7/site-packages/_pytest/runner.py:211: in __init__
    self.result = func()
.tox/python/lib/python3.7/site-packages/_pytest/runner.py:272: in <lambda>
    call = CallInfo(lambda: list(collector.collect()), "collect")
.tox/python/lib/python3.7/site-packages/_pytest/main.py:495: in collect
    for x in self._collect(arg):
E   RuntimeError: generator raised StopIteration
```
Works like this even locally :)

It should probably be reported to `pytest`.

webknjaz:
Reproducible only if the file is named `__init__.py`. Regardless of where it's located.

webknjaz:
@duper51 it's a new and already known bug in pytest: https://github.com/pytest-dev/pytest/issues/4344

Possible workaround for us: we could add `grep -v '__init__.py' |` before `xargs` as a temporary fix.

duper51:
Yeah, that looks like good, also on the title, initially I was fixing Travis, but apparently that was only broken due to the documentation change that was made earlier that was fixed by my `tox` run. I suppose we can just work on Circle for this one.

webknjaz:
Sounds good. I just wanted to see smth clear not just "fixing something mysterious".

duper51:
So I've reverted back a bunch of my earlier commits, and moving forward with the workaround till it's fixed in the `pytest` repo.

duper51:
Looks like it's breaking on a different bug now, related to the docker container not having OpenSSL developer libraries on it, but I can't be sure. I'll look into it.

duper51:
Looks like pip needs to be upgraded on the MacOS docker container. You happen to know what's involved with that?
@webknjaz

duper51:
Okay, we're finally at the end with this one. The tests are now working as expected on all platforms, however, they have presented a bug with one of the tests on pypy3. It seems the unix_socket test will hang when run under pypy3 will hang, which kills the build after 10 minutes. Since it's a separate issue, I'll make a report, but in the meantime I'm satisfied with the changes made here.

Are you happy with these changes @webknjaz?

PR #149, opened 2018-12-19
Make SCRIPT_NAME environ variable optional

AndreLouisCaron:
PEP 333 makes this CGI-style variable optional.  See the the section titled "environ variables":

> The following variables must be present, unless their value would be an empty string, in which case they may be omitted, except as otherwise noted below.

Some WSGI middleware does not set this environ variable.  When combined with `cheroot.wsgi.PathInfoDispatcher`, there is a confusing `KeyError` exception that may not be correctly be translated to an HTTP 500 response.

This change makes the variable optional.

  - [ x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#

A `KeyError` exception is raised when dispatching a request that does not set `SCRIPT_NAME` in the WSGI `environ` dictionary.  Unless the `PathInfoDispatch` is wrapped into some other WSGI middleware, the exception may not be translated to an HTTP 500 error.

The request is routed as though `SCRIPT_NAME` contained an empty string.

📋 **Other information**:

See PEP 333 (and the quote above).

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/149)

### DIFF
--- a/cheroot/test/test_dispatch.py
+++ b/cheroot/test/test_dispatch.py
@@ -1,0 +1,51 @@
+"""Tests for the HTTP server."""
+# -*- coding: utf-8 -*-
+# vim: set fileencoding=utf-8 :
+
+from __future__ import absolute_import, division, print_function
+
+from cheroot.wsgi import PathInfoDispatcher
+
+
+def wsgi_invoke(app, environ):
+    """Serve 1 requeset from a WSGI application."""
+    response = {}
+
+    def start_response(status, headers):
+        response.update({
+            'status': status,
+            'headers': headers,
+        })
+
+    response['body'] = b''.join(
+        app(environ, start_response)
+    )
+
+    return response
+
+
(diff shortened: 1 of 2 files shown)


webknjaz:
@AndreLouisCaron thanks for this patch! :tada:

AndreLouisCaron:
Out of curiosity, what is the usual release cycle?  I'm not in a hurry to get a release, but I'd like to know when I should be able to upgrade to a version with this fix.

Thanks a bunch!

webknjaz:
There's no regular release schedule, we cut releases once we feel like there's something worth releasing or if somebody asks to make it. Our release process is quite well automated so we just bump the version by pushing a new Git tag as per semver. After that, it gets published by Travis CI job.

PR #150, opened 2018-12-20
Merge with skeleton

jaraco:
In an attempt to bring in the changes to the skeleton since the last merge, I'm enacting a new merge that incorporates these changes.

In particular, in addition to the commits from the skeleton, this merge
- eliminates all the compatibility code in setup.py and just relies on the setup parameter for `setuptools_scm`.
- replaces collective.checkdocs with pytest-checkdocs, and runs that with the standard test suite instead of in the 'check' step.
- restores the name of the license file to LICENSE to match the skeleton
- reorders the config in setup.cfg to match skeleton. Removes 'license', which shouldn't be supplied unless the trove classifiers are insufficient.
- changed the 'url' to point to GitHub. I'm only comfortable linking to the docs pages if the docs pages have sufficient links to other resources.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/150)


jaraco:
Gah. Right off the bat, it fails linters because

- the upstream skeleton switched to black style quotes for docs/conf.py.
- the upstream setup.py doesn't conform to pep257 because the module doesn't have a docstring.

jaraco:
I'd blame myself for not knowing the programming language syntax, but as it turns out, the rules are contradictory.

.travis.yml
```diff
@@ -146,17 +146,13 @@ jobs:
...
-        secure: RAfz06AINvz7bfij/YhfkAreRqamgxS8a6jSRNxntYhtJke3ZszUbIDag8+n1I+G5XT2LnMhHqPNR7Plc+AeMz7VXTuy+b81Li5kse20NYlPhd7mBVmTUpXtqYQashV5J39F4qkATBLznCOrMEomM07VTXjO/o2hmQuXniab2Uo=
-      distributions: dists
-      skip_cleanup: true
+    before_script: skip
+    env:
+    - TWINE_USERNAME=jaraco
+    # TWINE_PASSWORD
+    - secure: ahqqwT5dPGZMyOr2NyavsSQAaWgE/RKy0bD0bT55MdXfVh7lpXU6sVupxKJT6/uwS8uaAahvxBjr6mTvTs7pHuUPJ0MVsFb+LnQVUzlqc9Z47tE7nWGn/mYXb+mxN/VBhlUX/Y4rQMxcW6VFqOMSQBPQ2EukIh76nerbvQC1ypA=
+    - TOX_TESTENV_PASSENV="TWINE_USERNAME TWINE_PASSWORD"
+    script: tox -e release
```
webknjaz: I'd like to keep dpl

docs/conf.py
```diff
@@ -3,21 +3,21 @@
 """Configuration of Sphinx documentation generator."""
 
 extensions = [
-    'sphinx.ext.autodoc',
-    'sphinx.ext.extlinks',
-    'sphinx.ext.intersphinx',
-    'jaraco.packaging.sphinx',
-    'rst.linker',
+    "sphinx.ext.autodoc",
```
webknjaz: black is evil :(

setup.cfg
```diff
@@ -1,25 +1,21 @@
...
 
 [metadata]
+license_file = LICENSE
 name = cheroot
-url = https://cheroot.cherrypy.org
+author = CherryPy Team
+author_email = team@cherrypy.org
+description = Highly-optimized, pure-python HTTP server
+long_description = file: README.rst
+url = https://github.com/cherrypy/cheroot
```
webknjaz: Why not docs site?

setup.cfg
```diff
@@ -50,45 +46,33 @@ keywords =
...
 
-# These are required in actual runtime:
 install_requires =
     backports.functools_lru_cache
     six>=1.11.0
     more_itertools>=2.6
 
 [options.extras_require]
-docs =
+testing =
```
webknjaz: Why not keep alphabetic ordering?

setup.py
```diff
@@ -1,136 +1,8 @@
-#! /usr/bin/env python
-"""Cheroot package setuptools installer."""
+#!/usr/bin/env python
```
webknjaz: Please keep that whitespace for better readability.

setup.py
```diff
@@ -1,136 +1,8 @@
...
-# Ref https://github.com/pypa/setuptools/issues/1136
-if 'package_dir' in declarative_setup_params['options']:
-    declarative_setup_params['options']['package_dir'] = {
-        str(k): str(v)
-        for k, v in declarative_setup_params['options']['package_dir'].items()
-    }
-
-setup_params = dict(setup_params, **declarative_setup_params['metadata'])
-setup_params = dict(setup_params, **declarative_setup_params['options'])
+# Project skeleton maintained at https://github.com/jaraco/skeleton
```
webknjaz: I'd like to keep shim up until we drop python 2 here. My goal was to never touch `setup.py` again. Also, it tightens pre-requisites requirements which is unfriendly to users.

setup.py
```diff
@@ -1,136 +1,8 @@
...
 
+import setuptools
 
-__name__ == '__main__' and setuptools.setup(**setup_params)
+if __name__ == "__main__":
```
webknjaz: Let's unblack it.

.travis.yml
webknjaz: Also, script step corresponds to testing. It's logical to put any validation here, but not publishing.

webknjaz: Curious whether we'll be able to use Actions for this once events other than `push` will be available... https://github.com/cherrypy/cheroot/actions

pyproject.toml
```diff
@@ -1,2 +1,3 @@
 [build-system]
-requires = ["setuptools>=30.3", "wheel", "setuptools_scm>=1.15"]
+requires = ["setuptools>=34.4", "wheel", "setuptools_scm>=1.15"]
```
webknjaz: I'm not sure whether we can rely on this solely with the currently default pip versions available to users.

pyproject.toml
```diff
@@ -1,2 +1,3 @@
...
+build-backend = 'setuptools.build_meta'
```
webknjaz: Different quotes style

tox.ini
```diff
@@ -56,9 +55,18 @@ commands = pre-commit run --config .pre-commit-config.yaml.failing --all-files {
 [testenv:setup-check]
 deps =
     twine
-extras =
-    docs
```
webknjaz: I think we should check docs separately from the test suite because this way different types of CI failures are better discoverable and recognizable

setup.cfg
jaraco: I use semantic ordering - placing the more meaningful extras at the beginning, with the skeleton ones first and project-specific extras after. I don't feel strongly about the ordering, though I don't see a lot of value in maintaining alphabetic ordering. Other projects have different ordering preferences (such as shorter names first). For me, stability and consistency across projects is more important.

setup.py
jaraco: Yeah, I'll have to figure this out upstream (skeleton). Some downstream projects are going to demand black style. Others, the other style.

setup.py
jaraco: With wheels published, most users won't be affected by setup.py or its prerequisites at all. Plus, with pip 10, available with most envs, including those defined by this project, the pyproject.toml designates the requirements, so there's no burden on the users.

.travis.yml
jaraco: I disagree. Although the original use-case for tox is to invoke tests, it's actually more useful when thought of as an environment builder and command invoker, not dissimilar to what the CI tools do. By moving the release logic to tox, it makes it possible to cut releases locally using the exact same command as the CI environment does (if as happens from time-to-time the CI environment fails to deploy).

jaraco: I'm not familiar with actions.

pyproject.toml
jaraco: It's not really reliant on it (yet). It still supplies 'setup_requires'. But still, only developers of cheroot are likely to be installing with something besides pip, so they'll all get wheels.

pyproject.toml
jaraco: Yeah, maybe that should be consistent. I'm pretty sure I copied these two different concepts from two separate sources. I can make them consistent in skeleton. [This reference](https://snarky.ca/clarifying-pep-518/) suggests double-quotes for everything.

tox.ini
jaraco: Yes, maybe. I do want to keep things simple so they scale from tiny projects to larger ones. The most important part of this change is that it eliminates support on invoking 'setup.py' (such that only pip does that and so it supplies all dependencies).

jaraco:
The answer to most of the "why" questions - it really comes down to "to avoid unnecessary deviation from other projects". I don't have the energy or time to maintain 100+ projects each with their own approach. My goal with skeleton was to adopt best practices and then share those with the other projects.

It's clear from all of the conflicts that this project has diverged too much from the skeleton for the skeleton to be useful except maybe as a reference. I'm going to resign from maintaining the package and I'll just file tickets when defects arise.

.travis.yml
webknjaz: I'm not talking about tox but Travis' `script` step vs `deploy` step which can also be described in tox of course.

As for CI deployment failures, if tests fail you can manually trigger deployment job only. Also, if it fails because of name collision (because of reruns?) there's a param to ignore such failures.

webknjaz: It's a brand-new toy: https://github.com/features/actions. We have partial access to them. It looks like we can build container-based pipelines on top of this reacting to various GitHub events. And there's some pre-cooked building blocks which is neat.

pyproject.toml
webknjaz: At @ansible we see so much weird retro envs so I just stopped being optimistic about this at some point...

pyproject.toml
webknjaz: Yeah, I think I only saw people using double quotes in toml files so far.

tox.ini
webknjaz: I think some of pip maintainers want `--only-extras` feature for this case... So it still needs to be pulled in from an extra

webknjaz:
Yeah... I see I'll keep supporting more backwards compat things for now.

PR #117, opened 2018-09-30
Non-blocking socket support/better handling of HTTP/1.1 connections

hexaclock:
* **What kind of change does this PR introduce?**
  - [X] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] tests/coverage improvement
  - [ ] refactoring
  - [ ] other

* **What is the related issue number (starting with `#`)**
#91

* **What is the current behavior?** (You can also link to an open issue here)
#91 

* **What is the new behavior (if this is a feature change)?**

* **Other information**:

* **Checklist**:

  - [x] I think the code is well written
  - [ ] I wrote [good commit messages][1]
  - [ ] I have [squashed related commits together][2] after the changes have been approved
  - [ ] Unit tests for the changes exist
  - [ ] Integration tests for the changes exist (if applicable)
  - [X] I used the same coding conventions as the rest of the project
  - [x] The new code doesn't generate linter offenses
  - [ ] Documentation reflects the changes
  - [X] The PR relates to *only* one subject with a clear title
  and description in grammatically correct, complete sentences

[1]: http://chris.beams.io/posts/git-commit/
[2]: https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/117)

Issue #91, opened 2018-05-04
HTTP Keep-Alive Connections not handled correctly


hexaclock:
* **I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
Bug

* **What is the current behavior?**
When the number of open Keep-Alive sessions exceeds CherryPy's thread count specified by `server.thread_pool`, CherryPy queues all new Keep-Alive connections up to the timeout specified by `server.socket_timeout`.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**
```python
# mini_cherrypy_server.py
import cherrypy


class HelloWorld(object):
    @cherrypy.expose
    def index(self):
        print(cherrypy.request.headers)
        return "Hello world!"


if __name__ == '__main__':
    cherrypy.config.update({'server.thread_pool': 5})
    cherrypy.quickstart(HelloWorld())
```

```python
# mini_client.py
import requests
from multiprocessing.dummy import Pool as ThreadPool

urls = ['http://localhost:8080/index'] * 10

def do_request(url):
    sess = requests.Session() # forces a keep-alive session
    resp = sess.get(url)
    print(resp.text)

def main():
    thread_pool = ThreadPool()
    results = thread_pool.map(do_request, urls)
    thread_pool.close()
    thread_pool.join()
    
if __name__ == '__main__':
    main()
```

* **What is the expected behavior?**
Allow threads in the thread pool to answer and process new connections by rotating "sleeping" sessions back into the queue.

* **What is the motivation / use case for changing the behavior?**
Correctly support the HTTP/1.1 spec.

* **Please tell us about your environment:**

- Cheroot version: 6.0.0
- CherryPy version: 13.1.0 (if applicable)
- Python version: 3.5.2 32-bit
- OS: Windows 10
- Browser: [all]

webknjaz:
Thanks, I'll take a look at this when I'll have some time

webknjaz:
@hexaclock I haven't had a chance to look into it, sorry. But likely you're right.
We'll likely need to need smth in https://github.com/cherrypy/cheroot/blob/master/cheroot/workers/threadpool.py, but I'm not sure.
[...]

Issue #69, opened 2017-12-28
Make HTTP server correctly read trailers (Trailer HTTP headers, which come after body)


webknjaz:
* **I'm submitting a ...**
- [X] bug report
- [X] feature request
- [ ] question about the decisions made in the repository

* **Do you want to request a *feature* or report a *bug*?**
It's a bug I noticed during tests rewrite for pytest runner #67 

* **What is the current behavior?**
Unread ending of the request with headers gets read by the next request parser, which recognises it as garbage (because it's not a valid Request-Line)

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**
See `conn_test.py::test_Chunked_Encoding`.

* **What is the expected behavior?**
Payload of the request with trailers is completely read out from the socket, so that it won't affect the next request.

* **What is the motivation / use case for changing the behavior?**
Emm.. My perfectionism, I guess.

* **Please tell us about your environment:**

- Python version: 3.6.2 (but shouldn't version-agnostic)
- Cheroot version: v6.0.0 (and probably all prev versions)
- CherryPy version: N/A
- OS: Gentoo Linux
- Browser: N/A

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

jaraco:
Here's the test failure:

<details>

```
cheroot master $ tox -- --runxfail -k test_Chunked_Encoding                                                                                                               
python develop-inst-noop: /Users/jaraco/code/public/cherrypy/cheroot
[...]

PR #176, opened 2019-02-04
Non-blocking socket support/better handling of HTTP/1.1 connections

jaraco:
  - [X] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#91

This is a re-submit of #117 with some subsequent improvements.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/176)


jaraco:
I've published a pre-release version of this PR in [my devpi repo](https://m.devpi.net/jaraco/dev) as [cheroot-6.5.6.dev28+g9413ed9c](https://m.devpi.net/jaraco/dev/cheroot/6.5.6.dev28+g9413ed9c).

the-allanc:
I don't think this patch even works as is.

Using this basic setup:
```
import cherrypy

class Server:
    @classmethod
    def run(cls):
        config = {
            'global': {
                'server.socket_host': '::',
                'server.socket_port': 56951,
                'server.socket_timeout': 0,
                'server.thread_pool': 2,
            }
        }

        cherrypy.config.update(config)
        cherrypy.engine.start()
        cherrypy.tree.mount(cls(), '/')
        cherrypy.engine.block()

    @cherrypy.expose
    def index(self):
        return "Hello World!"

__name__ == '__main__' and Server.run()
```

And then running this in a separate Python prompt:

```
Python 2.7.12 (default, Nov 12 2018, 14:36:49) 
[GCC 5.4.0 20160609] on linux2
>>> import requests
>>> s = [requests.Session() for i in range(3)]
>>> url = 'http://localhost:56951/'
>>> r0 = s[0].get(url)
>>> r1 = s[1].get(url)
>>> r2 = s[2].get(url)
```

This causes the final line to just hang, which seems to indicate that the keep-alive connections are still occupying threads.

the-allanc:
After looking at this for about a day and a half, I'm not convinced this is the right way of dealing with this issue (even if we corrected the issue). We're conflating multiple things at the same time.

The initial reporting of this issue was to do with some requests encountering a delay of 10 seconds before being dealt with. The ideal way that this should be dealt with is to change the behaviour of the threadpool, so that threads work with *requests*, not *connections*.

But the fix that we have here works around the problem, by focusing on the behaviour we have with keep-alive connections, and encourages the use of non-blocking sockets as a way to address the issue. It doesn't quite bring us request-based worker threads, but it does give us threads with a one-to-many relationship with connections.

This does improve the situation in some conditions, but not without pitfalls:
 - A secondary request coming in via a Keep-Alive session may have to wait if the thread that owns that connection is currently dealing with an incoming request from another connection. This could even happen if there are free threads.
 - It makes thread provisioning much more difficult. If a thread owns a keep-alive connection, then is it ever really "idle"? It has to continuously check the connections that it owns.
 
I don't think we should continue with work on this branch - it's a hack (which is an improvement), but it then starts conflating and confusing issues. I'm going to bring up the separate issues we should focus on in further comments.

jaraco:
Thank you @the-allanc for your critical review and insight. You've got a tighter grasp on this issue than any one else. If you believe this change is inadequate or incorrect, then I'm inclined to agree.

I'm having difficulty disentangling these issues, as you've attempted to do. Our key motivations here are to find a way to robustly handle HTTP 1.1 connections under load, and that using non-blocking sockets would achieve that. cherrypy/cherrypy#1764 was another manifestation of the issue. So we're not wed to supporting non-blocking sockets, but to devising the most robust solution we can afford to implement.

I like your proposal of request-based workers. Would you be willing to draft a proposal?

webknjaz:
I completely agree with @the-allanc's assessments. I just want to add that HTTP/1.1 pipelining seems to be a bit broken: https://github.com/cherrypy/cheroot/issues/69.

the-allanc:
I've created #199 as an alternative fix.

webknjaz:
Closing in favor or #199.

Issue #99, opened 2018-06-12
Extrapolating statistics from Cheroot doesn't work


sanitec:
**I'm submitting a ...**
- [X] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Describe the bug. What is the current behavior?**

Getting the following error when trying to extrapolate statistics from Cheroot:
AttributeError: 'BufferedReader' object has no attribute 'bytes_read'

**What is the motivation / use case for changing the behavior?**
It's broken

**To Reproduce**

1. Configure Cherrypy to use the following plugin:

```python
import logging

import cherrypy

class CherootStatistics(cherrypy.process.plugins.SimplePlugin):
    def __init__(self, bus):
        super(CherootStatistics, self).__init__(bus)
        self._enabled = False

    def main(self):
        if not self._enabled:
            self.bus.log('Enabling Cheroot statistics')
            for key in logging.statistics.keys():
                if 'Cheroot HTTPServer' in key:
                    logging.statistics[key]['Enabled'] = True
                    break
            self._enabled = True
```
2.  Mount cpstats to '/cpstats'
3.  Open localhost:8080/cpstats in browser

**Expected behavior**

To see the beautiful statistics of Cheroot

**Details**

```python-traceback
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/cherrypy/_cprequest.py", line 631, in respond
    self._do_respond(path_info)
  File "/usr/local/lib/python3.6/site-packages/cherrypy/_cprequest.py", line 690, in _do_respond
    response.body = self.handler()
  File "/usr/local/lib/python3.6/site-packages/cherrypy/lib/encoding.py", line 264, in __call__
    ct.params['charset'] = self.find_acceptable_charset()
  File "/usr/local/lib/python3.6/site-packages/cherrypy/lib/encoding.py", line 173, in find_acceptable_charset
    if encoder(self.default_encoding):
  File "/usr/local/lib/python3.6/site-packages/cherrypy/lib/encoding.py", line 114, in encode_string
    for chunk in self.body:
  File "/usr/local/lib/python3.6/site-packages/cherrypy/lib/cpstats.py", line 522, in index
    for title, scalars, collections in self.get_namespaces():
  File "/usr/local/lib/python3.6/site-packages/cherrypy/lib/cpstats.py", line 587, in get_namespaces
    s = extrapolate_statistics(logging.statistics)
[...]

PR #186, opened 2019-03-27
Add support for statistics on Python 3.

jaraco:
  - [X] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#99

Server crashes when statistics are enabled on Python 3.

No crash and statistics are counted. New tests ensure the required interface.

📋 **Other information**:

📋 **Checklist**:

  - [X] Integration tests for the changes exist (if applicable)

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/186)

### DIFF
--- a/cheroot/test/test_makefile.py
+++ b/cheroot/test/test_makefile.py
@@ -1,3 +1,5 @@
+"""self-explanatory."""
+
 from cheroot import makefile
 
 
@@ -5,10 +7,14 @@ __metaclass__ = type
 
 
 class MockSocket:
+    """Mocks a socket."""
+
     def __init__(self):
+        """Initialize."""
         self.messages = []
 
     def recv_into(self, buf):
+        """Simulate recv_into for Python 3."""
         if not self.messages:
             return 0
         msg = self.messages.pop(0)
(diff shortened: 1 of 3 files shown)


cheroot/test/test_makefile.py
```diff
@@ -0,0 +1,52 @@
+"""self-explanatory."""
```
webknjaz: not really...

jaraco: I could write "Tests for the makefile module.", but I feel that's redundant to the name of this module. What would you recommend? Feel free to enhance it.

jaraco: BTW, I wouldn't have bothered writing a docstring at all were it not required by the linters.

webknjaz: Yeah, I have a different policy now. I think it's useful to mention what to expect from the tests inside the module without forcing the reader to actually read tests.

jaraco: That sounds like a maintenance nightmare... trying to keep documentation about tests up-to-date with the code, especially when the documentation isn't adjacent to the code seems like you're just asking for trouble. As much as possible, I try to make code self-documenting and any documentation should expand on or clarify the purpose of a given piece of code and generally not restate what the code describes. Anyway, that's my approach. You're welcome to pursue more aggressive documentation strategies.

webknjaz: I'm thinking more about something answering the WHY question. High-level, mostly not requiring updates over time. What this test suite is for, how it's different from others.

PR #199, opened 2019-05-06
Request based workers

the-allanc:
  - [X] 🐞 bug fix
  - [x] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#91 

❓ **What is the current behavior?**

Worker threads are allocated a connection object, and that thread is occupied by that connection is closed or times out.

This can lead to the server having no spare threads to handle new incoming connections - even if there are no further requests coming over the used connections.

We maintain a connection pool of open connections. Cheroot uses `select` to determine if any of the open connections or the server socket has any incoming data - we then either use an existing connection or accept a new connection, and pass that to the worker thread.

Once the worker thread has processed a request, it will pass the connection back to the connection pool (if it decides that the connection should not be closed). This allow threads to only be occupied by connections that are ready with a request.

📋 **Other information**:

The connection pool uses two settings - a new one to determine how many keep-alive connections to keep open (this only applies to connections which are not currently being handled by a thread).

It also uses the server socket timeout setting to determine when to close an idle keep-alive connection.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/199)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -226,7 +226,10 @@ class ConnectionManager:
 
             conn = self.server.ConnectionClass(self.server, s, mf)
 
-            if not isinstance(self.server.bind_addr, six.string_types):
+            if not isinstance(
+                    self.server.bind_addr,
+                    (six.text_type, six.binary_type),
+            ):
                 # optional values
                 # Until we do DNS lookups, omit REMOTE_HOST
                 if addr is None:  # sometimes this can happen
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -61,16 +61,17 @@ class ConnectionManager:
 
         # Check for too many open connections.
         conns = list(self.connections.items())
-        if self.server.keep_alive_conn_limit is not None:
-            [self.close(conn[0]) for conn in conns[:-self.server.keep_alive_conn_limit]]
-            conns = conns[-self.server.keep_alive_conn_limit:]
+        ka_limit = self.server.keep_alive_conn_limit
+        if ka_limit is not None:
+            [self._close(conn[0]) for conn in conns[:-ka_limit]]
+            conns = conns[-ka_limit:]
 
         # Check for old connections.
         now = time.time()
         for (conn, (ctime, _)) in conns:
             # Oldest connection out of the currently available ones.
             if (ctime + self.server.timeout) < now:
-                self.close(conn)
+                self._close(conn)
             else:
                 break
 
@@ -121,7 +122,7 @@ class ConnectionManager:
         del self.connections[conn]
         return conn
 
-    def close(self, conn):
+    def _close(self, conn):
         del self.connections[conn]
         conn.close()
 
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -1,3 +1,5 @@
+"""Utilities to manage open connections."""
+
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
@@ -47,15 +49,38 @@ else:
 
 
 class ConnectionManager:
+    """Class which manages HTTPConnection objects.
+
+    This is for connections which are being kept-alive for follow-up requests.
+    """
 
     def __init__(self, server):
+        """Initialize ConnectionManager object.
+
+        Args:
+            server (cheroot.server.HTTPServer): web server object
+                that uses this ConnectionManager instance.
+        """
         self.server = server
         self.connections = OrderedDict()
 
     def put(self, conn):
+        """Put idle connection into the ConnectionManager to be managed.
+
+        Args:
+            conn (cheroot.server.HTTPConnection): HTTP connection
+                to be managed.
+        """
         self.connections[conn] = time.time(), conn.rfile.has_data()
 
     def expire(self):
+        """Expire least recently used connections.
+
+        This happens if there are either too many open connections, or if the
+        connections have been timed out.
+
+        This should be called periodically.
+        """
         if not self.connections:
             return
 
@@ -76,7 +101,22 @@ class ConnectionManager:
                 break
 
     def get_conn(self, server_socket):
+        """Return a HTTPConnection object which is ready to be handled.
 
+        A connection returned by this method should be ready for a worker
+        to handle it. If there are no connections ready, None will be
+        returned.
+
+        Any connection returned by this method will need to be `put`
+        back if it should be examined again for another request.
+
+        Args:
+            server_socket: ServerSocket instance to listen to for new
+            connections.
+        Returns:
+            cheroot.server.HTTPConnection instance, or None.
+
+        """
         # Grab file descriptors from sockets, but stop if we find a
         # connection which is already marked as ready.
         socket_dict = {}
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -8,7 +8,6 @@ import os
 import select
 import socket
 import time
-from collections import OrderedDict
 
 from . import errors
 from .makefile import MakeFile
@@ -62,7 +61,7 @@ class ConnectionManager:
                 that uses this ConnectionManager instance.
         """
         self.server = server
-        self.connections = OrderedDict()  # using this as a set
+        self.connections = []
 
     def put(self, conn):
         """Put idle connection into the ConnectionManager to be managed.
@@ -73,7 +72,7 @@ class ConnectionManager:
         """
         conn.last_used = time.time()
         conn.ready_with_data = conn.rfile.has_data()
-        self.connections[conn] = None
+        self.connections.append(conn)
 
     def expire(self):
         """Expire least recently used connections.
@@ -88,7 +87,7 @@ class ConnectionManager:
 
         # Look at the first connection - if it can be closed, then do
         # that, and wait for get_conn to return it.
-        conn = next(iter(self.connections))
+        conn = self.connections[0]
         if conn.closeable:
             return
 
@@ -133,7 +132,7 @@ class ConnectionManager:
 
         # We have a connection ready for use.
         if conn:
-            del self.connections[conn]
+            self.connections.remove(conn)
             return conn
 
         # Will require a select call.
@@ -154,12 +153,11 @@ class ConnectionManager:
                 try:
                     os.fstat(fno)
                 except OSError:
-                    # Socket is invalid, close the connection.
+                    # Socket is invalid, close the connection, insert at
+                    # the front.
+                    self.connections.remove(conn)
+                    self.connections.insert(0, conn)
                     conn.closeable = True
-                else:
-                    # Connection is fine, move to the end so that
-                    # closeable connections appear at the front.
-                    self.connections[fno] = self.connections.pop(fno)
 
             # Wait for the next tick to occur.
             return None
@@ -181,7 +179,7 @@ class ConnectionManager:
         if conn is server_socket:
             return self._from_server_socket(server_socket)
 
-        del self.connections[conn]
+        self.connections.remove(conn)
         return conn
 
     def _from_server_socket(self, server_socket):
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -2,11 +2,15 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import io
+import select
 import six
 import socket
+import time
 
 from . import errors
 from .makefile import MakeFile
+
+from collections import OrderedDict
 
 
 try:
@@ -46,8 +50,48 @@ class ConnectionManager:
 
     def __init__(self, server):
         self.server = server
+        self.connections = OrderedDict()
 
-    def from_server_socket(self, server_socket):
+    def put(self, conn):
+        self.connections[conn] = time.time()
+
+    def expire(self):
+        if not self.connections:
+            return
+
+        # Check for too many open connections.
+        conns = list(self.connections.items())
+        if self.server.keep_alive_conn_limit is not None:
+            [self.close(conn[0]) for conn in conns[:-self.server.keep_alive_conn_limit]]
+            conns = conns[-self.server.keep_alive_conn_limit:]
+
+        # Check for old connections.
+        now = time.time()
+        for (conn, ctime) in conns:
+            # Oldest connection out of the currently available ones.
+            if (ctime + self.server.timeout) < now:
+                self.close(conn)
+            else:
+                break
+
+    def get_conn(self, server_socket):
+        socket_dict = {s.socket.fileno(): s for s in self.connections}
+        socket_dict[server_socket.fileno()] = server_socket
+        rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
+        if not rlist:
+            return None
+        conn = socket_dict[rlist[0]]
+        if conn is server_socket:
+            return self._from_server_socket(server_socket)
+        else:
+            del self.connections[conn]
+            return conn
+
+    def close(self, conn):
+        del self.connections[conn]
+        conn.close()
+
+    def _from_server_socket(self, server_socket):
         try:
             s, addr = server_socket.accept()
             if self.server.stats['Enabled']:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import io
+import os
 import select
 import six
 import socket
@@ -138,10 +139,29 @@ class ConnectionManager:
         # Will require a select call.
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
-        rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
+        try:
+            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
+            # No available socket.
+            if not rlist:
+                return None
+        except OSError:
+            # Mark any connection which no longer appears valid.
+            for fno, conn in list(socket_dict.items()):
+                # If the server socket is invalid, we'll just ignore it and
+                # wait to be shutdown.
+                if fno == ss_fileno:
+                    continue
+                try:
+                    os.fstat(fno)
+                except OSError:
+                    # Socket is invalid, close the connection.
+                    conn.closeable = True
+                else:
+                    # Connection is fine, move to the end so that
+                    # closeable connections appear at the front.
+                    self.connections[fno] = self.connections.pop(fno)
 
-        # No available socket.
-        if not rlist:
+            # Wait for the next tick to occur.
             return None
 
         try:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -130,39 +130,40 @@ class ConnectionManager:
             # No ready connection.
             conn = None
 
+        # We have a connection ready for use.
+        if conn:
+            del self.connections[conn]
+            return conn
+
         # Will require a select call.
-        if not conn:
-            ss_fileno = server_socket.fileno()
-            socket_dict[ss_fileno] = server_socket
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
+        ss_fileno = server_socket.fileno()
+        socket_dict[ss_fileno] = server_socket
+        rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
 
-            # No available socket.
-            if not rlist:
-                return None
+        # No available socket.
+        if not rlist:
+            return None
 
-            try:
-                # See if we have a new connection coming in.
-                rlist.remove(ss_fileno)
-            except ValueError:
-                # No new connection, but reuse existing socket.
-                conn = socket_dict[rlist.pop()]
-            else:
-                conn = server_socket
+        try:
+            # See if we have a new connection coming in.
+            rlist.remove(ss_fileno)
+        except ValueError:
+            # No new connection, but reuse existing socket.
+            conn = socket_dict[rlist.pop()]
+        else:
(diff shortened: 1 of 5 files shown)


cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
+"""Utilities to manage open connections."""
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+import io
+import select
+import six
```
webknjaz: this is a third-party lib, it should go into a separate section

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+import socket
+import time
+
+from . import errors
+from .makefile import MakeFile
+
+from collections import OrderedDict
```
webknjaz: this is an import from stdlib, it should go into the top group of imports

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+try:
```
webknjaz: I feel like this whole try/except cascade should probably go under `_compat` module.

webknjaz:
@the-allanc have you seen https://docs.python.org/3/library/socketserver.html ? Is it applicable to Cheroot project somehow?

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+
+        This happens if there are either too many open connections, or if the
+        connections have been timed out.
+
+        This should be called periodically.
+        """
+        if not self.connections:
+            return
+
+        # Check for too many open connections.
```
webknjaz: It probably makes sense to check for expired connections first. This will allow keeping more of non-expired connections alive.

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        ka_limit = self.server.keep_alive_conn_limit
+        if ka_limit is not None:
+            [self._close(conn[0]) for conn in conns[:-ka_limit]]
+            conns = conns[-ka_limit:]
+
+        # Check for old connections.
+        now = time.time()
+        for (conn, (ctime, _)) in conns:
+            # Oldest connection out of the currently available ones.
+            if (ctime + self.server.timeout) < now:
```
webknjaz: It probably makes sense to add `expired` property to the connection itself which would turn this check into `if conn.expired:`

cheroot/connections.py
webknjaz: Oh, after looking closer: expired connections would go first anyway...

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        if not self.connections:
+            return
+
+        # Check for too many open connections.
+        conns = list(self.connections.items())
+        ka_limit = self.server.keep_alive_conn_limit
+        if ka_limit is not None:
+            [self._close(conn[0]) for conn in conns[:-ka_limit]]
+            conns = conns[-ka_limit:]
```
webknjaz: Maybe
```suggestion
            del conns[:-ka_limit]
```
?

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+
+        # Check for old connections.
+        now = time.time()
+        for (conn, (ctime, _)) in conns:
+            # Oldest connection out of the currently available ones.
+            if (ctime + self.server.timeout) < now:
+                self._close(conn)
```
webknjaz: You close connections but don't delete them from `self.connections`, do you? Wouldn't this cause a memory leak?

webknjaz: Oh, you just remove them inside of that call...

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        self.connections = OrderedDict()
+
+    def put(self, conn):
+        """Put idle connection into the ConnectionManager to be managed.
+
+        Args:
+            conn (cheroot.server.HTTPConnection): HTTP connection
+                to be managed.
+        """
+        self.connections[conn] = time.time(), conn.rfile.has_data()
```
webknjaz: How about having a [Circular buffer](https://en.wikipedia.org/wiki/Circular_buffer) for this?

webknjaz: Maybe inherit `deque`? https://stackoverflow.com/a/4151368/595220

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        # Check for too many open connections.
+        conns = list(self.connections.items())
+        ka_limit = self.server.keep_alive_conn_limit
+        if ka_limit is not None:
+            [self._close(conn[0]) for conn in conns[:-ka_limit]]
+            conns = conns[-ka_limit:]
+
+        # Check for old connections.
+        now = time.time()
+        for (conn, (ctime, _)) in conns:
```
webknjaz: This looks like a perfect use case for [`itertools.takewhile()`](https://docs.python.org/3/library/itertools.html#itertools.takewhile)

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Any connection returned by this method will need to be `put`
+        back if it should be examined again for another request.
+
+        Args:
+            server_socket: ServerSocket instance to listen to for new
+            connections.
+        Returns:
+            cheroot.server.HTTPConnection instance, or None.
+
```
webknjaz:
```suggestion
```

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        A connection returned by this method should be ready for a worker
+        to handle it. If there are no connections ready, None will be
+        returned.
+
+        Any connection returned by this method will need to be `put`
+        back if it should be examined again for another request.
+
+        Args:
+            server_socket: ServerSocket instance to listen to for new
```
webknjaz:
```suggestion
            server_socket (socket.socket): ServerSocket instance to listen to for new
```

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+            socket_dict[conn.socket.fileno()] = conn
+        else:
+            # No ready connection.
+            conn = None
+
+        # Will require a select call.
+        if not conn:
+            ss_fileno = server_socket.fileno()
+            socket_dict[ss_fileno] = server_socket
+            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
```
webknjaz: Is this run from the main thread?

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+        for conn, (tstamp, has_data) in self.connections.items():
+            if has_data:
+                break
+            socket_dict[conn.socket.fileno()] = conn
+        else:
+            # No ready connection.
+            conn = None
+
+        # Will require a select call.
+        if not conn:
```
webknjaz: You could revert the guard expression to be `if conn:` and return early. This would improve readability by reducing the indentation of the whole code block following this line.

cheroot/connections.py
```diff
@@ -0,0 +1,254 @@
...
+                    s.settimeout(self.server.timeout)
+
+            conn = self.server.ConnectionClass(self.server, s, mf)
+
+            if not isinstance(self.server.bind_addr, six.string_types):
+                # optional values
+                # Until we do DNS lookups, omit REMOTE_HOST
+                if addr is None:  # sometimes this can happen
+                    # figure out if AF_INET or AF_INET6.
+                    if len(s.getsockname()) == 2:
```
webknjaz: How about:
```suggestion
                    addr = '::', 0
                    if len(s.getsockname()) == 2:
                        addr[0] = '0.0.0.0'
```

cheroot/server.py
```diff
@@ -1978,99 +1947,18 @@ def resolve_real_bind_addr(socket_):
...
+        if not self.ready:
+            return
 
+        conn = self.connections.get_conn(self.socket)
+        if conn:
             try:
                 self.requests.put(conn)
             except queue.Full:
                 # Just drop the conn. TODO: write 503 back?
                 conn.close()
```
webknjaz: Should this happen through the manager?

cheroot/connections.py
the-allanc: Yes.

cheroot/connections.py
the-allanc: I only moved this existing code from `server.py` to `connections.py`, so I don't want to get involved with making changes to that code (at least, not in this pull request).

cheroot/connections.py
the-allanc: No comment on that - I just moved the code from `server.py` to `connections.py`.

cheroot/connections.py
the-allanc: And I did initially have no whitespace there, but then [this CI job](https://travis-ci.org/cherrypy/cheroot/jobs/529425494) failed because of it.

Is there some tool that we can use to deal with auto-format these things prior to commit?

cheroot/server.py
the-allanc: The manager is responsible for providing a connection, but the server is then responsible for getting a worker thread to work on it. So no.

cheroot/connections.py
the-allanc: I've changed this as suggested.

cheroot/connections.py
the-allanc: Fixed.

cheroot/connections.py
the-allanc: I've changed the behaviour of expire, so we don't loop like this any more.

cheroot/connections.py
the-allanc: I did have this in earlier implementations. The problem is that we want a queue when checking if connections are still valid, but we want a hash-based lookup when we want to remove the connection object from the manager.

cheroot/connections.py
the-allanc: This code doesn't exist any more.

cheroot/connections.py
the-allanc: I've done something like that in the most recent set of changes.

the-allanc:
> @the-allanc have you seen https://docs.python.org/3/library/socketserver.html ? Is it applicable to Cheroot project somehow?

I use the term "server socket" as a relic of my days programming in Java -[ServerSocket is a class](https://docs.oracle.com/javase/7/docs/api/java/net/ServerSocket.html) which is used for listening for and accepting incoming connections.

If some other term is appropriate, then please change it. It's not related to the SocketServer class in Python.

cheroot/connections.py
the-allanc: Done.

cheroot/connections.py
the-allanc: Done.

the-allanc:
I've changed the code a bit in a few ways:

1. We store properties on the `HTTPConnection` object to indicate if it has data ready on the socket, if it's closeable and when it can be expired/
2. We now only expire one connection at a time (the "oldest" one), and let the main Cheroot loop "tick" and to then process that connection.
3. WorkerThreads are now responsible for closing connections, rather than the main Cheroot thread (which is how it used to work previously).

cheroot/connections.py
jaraco: Yes. I suggest using [pipx](https://pypi.org/project/pipx) to install [pre-commit](https://pypi.org/project/pre-commit), then `pre-commit install`. This will run pre-commit before committing code. Alternately, for a lighter touch, just run `tox -e pre-commit`, which will run pre-commit manually. Sometimes, pre-commit will automatically edit the files. Other times, it can only raise an error requiring manual intervention.

webknjaz:
@the-allanc FTR I'm having an HTTP parser migration in my mind (https://github.com/cherrypy/cheroot/issues/201). How well will this patch play with that?

cheroot/connections.py
webknjaz: @the-allanc It would be nice if you could add a comment about this decision to the source code as well.

webknjaz:
@the-allanc have you seen my question @ https://github.com/cherrypy/cheroot/pull/199#issuecomment-490317311?

cheroot/connections.py
webknjaz: Yes, but since you're already moving it plz put it under _compat and just import this func from there making this module cleaner.

webknjaz: I'm not asking you to change it but I think it doesn't belong here.

cheroot/connections.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def __init__(self, server):
+        """Initialize ConnectionManager object.
+
+        Args:
+            server (cheroot.server.HTTPServer): web server object
+                that uses this ConnectionManager instance.
+        """
+        self.server = server
+        self.connections = OrderedDict()  # using this as a set
```
webknjaz: If you use it as a set why not use set? If you're using it for more, put a comment here explaining why

cheroot/makefile.py
```diff
@@ -235,6 +235,7 @@ def readline(self, size=-1):
                         break
                     buf.write(data)
                 return buf.getvalue()
+
```
webknjaz: I'm pretty sure this change is unrelated.

webknjaz:
I believe that @mar10 will be interested in testing this out as well.

the-allanc:
> @the-allanc FTR I'm having an HTTP parser migration in my mind (#201). How well will this patch play with that?

Had a quick look at it, I think generally it should work. Some minor changes required, like the connection manager needing a different way to determine if there's data in the buffer.

> @the-allanc have you seen my question @ [#199 (comment)](https://github.com/cherrypy/cheroot/pull/199#issuecomment-490317311)?

[Yes.](https://github.com/cherrypy/cheroot/pull/199#issuecomment-491367392)

the-allanc:
After testing with my previous changes, it seems that `OrderedDict` isn't thread-safe as I originally thought - I think this would explain some weird errors we were seeing. So after some thought, I've decided just to use a basic list to maintain the collection of connections.

@jaraco - do you think we can get another dev release to try out?

jaraco:
I made [this new release](https://m.devpi.net/jaraco/dev/cheroot/6.5.6.dev21+ge81bd245). I don't understand why it's `dev21` whereas the previous release was `dev28`.

jaraco:
Aha - because dev28 was part of a different prerelease for a different branch (9413ed9c).

webknjaz:
> because dev28 was part of a different prerelease

It's also tricky with history rewrites: since it's a distance to the last tag in chain squashing with force push will have a similar effect...

webknjaz:
@the-allanc why isn't socketserver module applicable for Cheroot? It's in stdlib and I think I saw Werkzeug using it. It seemed to me that if there's something as high-level in the stdlib it might not be worth it to have an own low-level networking implemented: does it have any downsides?

webknjaz:
@the-allanc what's not thread-safe with OrderedDict? Weren't you only manipulating it from a single thread?

cheroot/connections.py
```diff
@@ -133,7 +132,7 @@ def get_conn(self, server_socket):
 
         # We have a connection ready for use.
         if conn:
-            del self.connections[conn]
+            self.connections.remove(conn)
```
webknjaz: Removal of an item from a list in Python is expensive in general. Especially, from its beginning: all following elements are shifted in memory. So OrderedDict was a more effective solution.

cheroot/connections.py
```diff
@@ -154,12 +153,11 @@ def get_conn(self, server_socket):
                 try:
                     os.fstat(fno)
                 except OSError:
-                    # Socket is invalid, close the connection.
+                    # Socket is invalid, close the connection, insert at
+                    # the front.
+                    self.connections.remove(conn)
+                    self.connections.insert(0, conn)
```
webknjaz: Inserting things into the very beginning will shift the *entire* array in memory.

webknjaz: (https://stackoverflow.com/a/36529893/595220)

jaraco:
I'm inclined to accept this even with a potentially performance-sensitive data structure like a list. The number of connections will typically be on the order of tens or hundreds, so seems unlikely to be performance sensitive. If the code is working in the production environment where the-allanc is piloting this code, I'd consider that sufficient to validate it as performance-safe in most production environments.... and we can optimize after.

I do think a [collections.deque](https://docs.python.org/3.7/library/collections.html#collections.deque) may be what @webknjaz has in mind for a more performant structure. If that's easy enough to use in place of a list without complicating the semantics, it's probably a good choice.

webknjaz:
@jaraco even if I stop blocking this, the CIs (AppVeyor and Travis) should be made green first.

webknjaz:
I looked again. It's acceptable now but still needs CIs to be green before merging.

the-allanc:
> @the-allanc why isn't socketserver module applicable for Cheroot? It's in stdlib and I think I saw Werkzeug using it. It seemed to me that if there's something as high-level in the stdlib it might not be worth it to have an own low-level networking implemented: does it have any downsides?

Let me clarify - I explained why I used the term "server socket" in [this comment](https://github.com/cherrypy/cheroot/pull/199#issuecomment-491367392). It's not directly related to the `socketserver` module. I could have called it `incoming_connection_socket`, and we wouldn't be having this conversation.

Now if you want to use the `socketserver` module, that's your choice. It's been there since Python 1.5 (and possibly earlier), but CherryPy was developed after that and yet chose to ignore its existence. Maybe there were good reasons for that. But moving to that module is something that can be considered outside this merge request.

cheroot/connections.py
the-allanc: > Weren't you only manipulating it from a single thread?
No. The main thread was cycling through open connections, but worker threads were also putting back non-closed connections.

cheroot/connections.py
the-allanc: This isn't an action we would expect to be performing often.

the-allanc:
>> I do think a [collections.deque](https://docs.python.org/3.7/library/collections.html#collections.deque) may be what @webknjaz has in mind for a more performant structure. If that's easy enough to use in place of a list without complicating the semantics, it's probably a good choice.

[From the documentation](https://docs.python.org/2/library/collections.html#collections.deque) on the page for `collections.deque`:
> Indexed access is O(1) at both ends but slows to O(n) in the middle. For fast random access, use lists instead.

We do both - adding to the end of the structure happens when reusable connections are put back, popping from the start of the structure happens when a connection needs to be expired, but connections are also removed from the middle of the structure itself.

I'm agnostic regarding what structure is used (`list` or `deque`), but I feel it's worth pointing out that it's by no means clear that a `deque` is better than a `list` for this code.

webknjaz:
@the-allanc AFAIU `deque` is more efficient for removing items from the start of the collection because this operation doesn't require shifting each and every item in the memory.

webknjaz:
As for the `socketserver`, I noticed it in the recent Werkzeug versions and was just curious. It's definitely unrelated to this PR :)

mar10:
Hi, 
it seems that this PR is 95% complete and seems to bring major improvements to speed in some scenarios. Are there any estimations, when you can merge it?

jaraco:
The main blocker here is figuring out why the tests are failing and getting them passing. If they're failing in the master, then we need to either fix those failures or mark them as xfail and file a separate ticket to fix those failures (with a preference being to fix them properly in order to maintain coverage). Then, merge the passing master into these proposed changes and assert stability.

If instead the test failures are due to the proposed changes, someone needs to investigate why and what needs to happen to the test suite to allow tests to pass without degrading coverage.

webknjaz:
(triggered CIs)

webknjaz:
pre-commit check, which has been failing earlier is no longer red.

webknjaz:
It looks like `cheroot.test.test_ssl.test_http_over_https_error[::-builtin]` is persistently failing on master in AppVeyor under Python 2.7 only.

webknjaz:
Alright, I kinda patched that..

webknjaz:
Going to rebase this branch.

webknjaz:
Looks like this PR makes some of TLS tests unstable under Windows: https://ci.appveyor.com/project/CherryPy/cheroot/build/job/bghdcuv66ri0o1pk

webknjaz:
@the-allanc I've rebased this branch and solved the conflicts.

h-vetinari:
Any update on this? :)

cheroot/server.py
```diff
@@ -1255,30 +1260,26 @@ def __init__(self, server, sock, makefile=MakeFile):
...
-                    return
+            req = self.RequestHandlerClass(self.server, self)
+            req.parse_request()
+            if self.server.stats['Enabled']:
+                self.requests_seen += 1
+            if not req.ready:
+                # Something went wrong in the parsing (and the server has
+                # probably already made a simple_response). Return and
+                # let the conn close.
+                return
```
webknjaz: I don't recall if I asked this before but it feels like this should return `False` explicitly. WDYT?

the-allanc: I agree, I've made changes for it.

jaraco:
I've tested this branch on my Windows box and the Python 2.7 errors happen on master same as on the branch, so I believe this branch is safe to merge.

webknjaz:
Looks like this change introduced the following regressions:
* https://github.com/cherrypy/cheroot/issues/249
* https://github.com/cherrypy/cheroot/issues/263

jaraco:
@the-allanc As I imagine you're aware, this change has caused some fairly serious regressions in the stability of cheroot (#249, #263). I suspect they'll be solved by the same investigation. I've spent some time in #263 writing a regression test that captures the failure and expected outcome. Will you have time to investigate these issues in the next few weeks? If not, I may take a stab at it, but if neither of us can make progress on it in fairly short order, I'm going to have to revert this feature.

PR #204, opened 2019-05-22
Fix race condition in threadpool shrink code.

the-allanc:
  - [X] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

This was previously reported as a [CherryPy bug](https://github.com/cherrypy/cherrypy/issues/1206).

📋 **Other information**:

There was a previously submitted fix in the original ticket. I've transplanted the idea behind it into cheroot, and changed the counter to a deque object instead (as I suspect modifying a counter on an object is not an atomic operation).

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/204)

### DIFF
--- a/cheroot/workers/threadpool.py
+++ b/cheroot/workers/threadpool.py
@@ -4,6 +4,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
+import collections
 import threading
 import time
 import socket
@@ -155,6 +156,7 @@ class ThreadPool:
         self._queue = queue.Queue(maxsize=accepted_queue_size)
         self._queue_put_timeout = accepted_queue_timeout
         self.get = self._queue.get
+        self._pending_shutdowns = collections.deque()
 
     def start(self):
         """Start the pool of threads."""
@@ -170,7 +172,8 @@ class ThreadPool:
     @property
     def idle(self):  # noqa: D401; irrelevant for properties
         """Number of worker threads which are idle. Read-only."""
-        return len([t for t in self._threads if t.conn is None])
+        idles = len([t for t in self._threads if t.conn is None])
+        return max(idles - len(self._pending_shutdowns), 0)
 
     def put(self, obj):
         """Put request into queue.
@@ -180,8 +183,16 @@ class ThreadPool:
                 waiting to be processed
         """
         self._queue.put(obj, block=True, timeout=self._queue_put_timeout)
-        if obj is _SHUTDOWNREQUEST:
-            return
+
+    def _clear_dead_threads(self):
+        # Remove any dead threads from our list
+        for t in self._threads:
+            if not t.isAlive():
+                self._threads.remove(t)
+                try:
+                    self._pending_shutdowns.popleft()
+                except IndexError:
+                    pass
 
     def grow(self, amount):
         """Spawn new worker threads (not above self.max)."""
@@ -208,10 +219,10 @@ class ThreadPool:
         """Kill off worker threads (not below self.min)."""
         # Grow/shrink the pool if necessary.
         # Remove any dead threads from our list
-        for t in self._threads:
-            if not t.is_alive():
-                self._threads.remove(t)
-                amount -= 1
+        amount -= len(self._pending_shutdowns)
+        self._clear_dead_threads()
+        if amount <= 0:
+            return
 
         # calculate the number of threads above the minimum
         n_extra = max(len(self._threads) - self.min, 0)
@@ -223,6 +234,7 @@ class ThreadPool:
         # to remove. As each request is processed by a worker, that worker
         # will terminate and be culled from the list.
         for n in range(n_to_remove):
+            self._pending_shutdowns.append(None)
             self._queue.put(_SHUTDOWNREQUEST)
 
     def stop(self, timeout=5):
--- a/cheroot/workers/threadpool.py
+++ b/cheroot/workers/threadpool.py
@@ -186,7 +186,7 @@ class ThreadPool:
 
     def _clear_dead_threads(self):
         # Remove any dead threads from our list
-        for t in [t for t in self._threads if not t.isAlive()]:
+        for t in [t for t in self._threads if not t.is_alive()]:
             self._threads.remove(t)
             try:
                 self._pending_shutdowns.popleft()
--- a/cheroot/workers/threadpool.py
+++ b/cheroot/workers/threadpool.py
@@ -186,13 +186,12 @@ class ThreadPool:
 
     def _clear_dead_threads(self):
         # Remove any dead threads from our list
-        for t in self._threads:
-            if not t.isAlive():
-                self._threads.remove(t)
-                try:
-                    self._pending_shutdowns.popleft()
-                except IndexError:
-                    pass
+        for t in [t for t in self._threads if not t.isAlive()]:
+            self._threads.remove(t)
+            try:
+                self._pending_shutdowns.popleft()
+            except IndexError:
+                pass
 
     def grow(self, amount):
         """Spawn new worker threads (not above self.max)."""
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ v6.6.0
 - Revisit :pr:`85` under :pr:`221`. Now
   ``backports.functools_lru_cache`` is only
   required on Python 3.2 and earlier.
+- :cp-issue:`1206` via :pr:`204`: Fix race condition in
+  threadpool shrink code.
 
 v6.5.8
 ======


the-allanc:
To test this, I've been using a basic CherryPy app (with cherrypy_dynpool installed):

```import cherrypy
import cherrypy_dynpool
import time

class HelloWorld(object):
    @cherrypy.expose
    def index(self):
        time.sleep(0.5)
        return "Yo World!"


def main():
    cherrypy.engine.threadpool_monitor = cherrypy_dynpool.ThreadPoolMonitor(cherrypy.engine)
    cherrypy.engine.threadpool_monitor.subscribe()

    cherrypy.config.update({
        'server.thread_pool': 5,
        'server.thread_pool_max': 600,
        'server.thread_pool_minspare': 2,
        'server.thread_pool_maxspare': 40,
        'server.thread_pool_frequency': 4,
        'server.thread_pool_shrink_frequency': 10,
        'server.thread_pool_log_frequency': 3,
    })

    app = cherrypy.tree.mount(HelloWorld(), '/', config={})
    cherrypy.engine.start()
    cherrypy.engine.threadpool_monitor.configure(
        thread_pool=cherrypy.server.httpserver.requests,
        logger=cherrypy.engine.log
    )
    cherrypy.engine.block()


if __name__ == '__main__':
    main()
```
I then use ApacheBench to fire multiple requests at the server: `ab -n 6000 -c 300 http://localhost:8080/`

On my machine, both versions of cheroot (before and after my changes) fail to fulfill all 6000 requests (they normally manage around 5800 requests). But in the code before my changes, the logs look like this at the end:

```
[22/May/2019:22:53:34] ENGINE Thread pool: [current=64/idle=64/queue=0]
[22/May/2019:22:53:41] ENGINE Thread pool: [current=64/idle=64/queue=0] Shrinking by 59
[22/May/2019:22:53:43] ENGINE Thread pool: [current=33/idle=33/queue=23]
[22/May/2019:22:55:10] ENGINE Thread pool: [current=33/idle=33/queue=112]
```
This is the result of ApacheBench finishing running, and then cheroot being in a state where threads are apparently idle, but they are no longer handling any requests. The last line is me attempting to re-run ApacheBench against the same server, but no requests are being processed at all.

Repeating that against cheroot with my changes in place:

```
[22/May/2019:22:59:31] ENGINE Thread pool: [current=41/idle=41/queue=0] Shrinking by 36
[22/May/2019:22:59:32] ENGINE Thread pool: [current=41/idle=5/queue=0]
[22/May/2019:22:59:41] ENGINE Thread pool: [current=41/idle=5/queue=0] Shrinking by 2
[22/May/2019:22:59:41] ENGINE Thread pool: [current=21/idle=5/queue=0]
[22/May/2019:22:59:51] ENGINE Thread pool: [current=21/idle=5/queue=0] Shrinking by 2
[22/May/2019:22:59:53] ENGINE Thread pool: [current=12/idle=5/queue=0]
[22/May/2019:23:00:01] ENGINE Thread pool: [current=12/idle=5/queue=0] Shrinking by 2
[22/May/2019:23:00:02] ENGINE Thread pool: [current=7/idle=5/queue=0]
[22/May/2019:23:00:11] ENGINE Thread pool: [current=7/idle=5/queue=0] Shrinking by 2
[22/May/2019:23:00:11] ENGINE Thread pool: [current=5/idle=5/queue=0]
```

The thread pool successfully shrinks down to the desired size, and you can re-run ApacheBench against the server too. Note that there's nothing stuck in the queue, either.

cheroot/workers/threadpool.py
```diff
@@ -180,8 +183,16 @@ def put(self, obj):
                 waiting to be processed
         """
         self._queue.put(obj, block=True, timeout=self._queue_put_timeout)
-        if obj is _SHUTDOWNREQUEST:
-            return
+
+    def _clear_dead_threads(self):
+        # Remove any dead threads from our list
+        for t in self._threads:
+            if not t.isAlive():
```
webknjaz: Could you please change this to a "guard expression" so that the nesting level would be lower?
```suggestion
            if t.isAlive():
                continue
```

cheroot/workers/threadpool.py
```diff
@@ -180,8 +183,16 @@ def put(self, obj):
                 waiting to be processed
         """
         self._queue.put(obj, block=True, timeout=self._queue_put_timeout)
-        if obj is _SHUTDOWNREQUEST:
-            return
+
+    def _clear_dead_threads(self):
+        # Remove any dead threads from our list
+        for t in self._threads:
```
webknjaz: Alternatively, just filter out stuff right here:
```suggestion
        for t in (t for t in self._threads if t.isAlive()):
```

cheroot/workers/threadpool.py
```diff
@@ -208,10 +219,10 @@ def shrink(self, amount):
         """Kill off worker threads (not below self.min)."""
         # Grow/shrink the pool if necessary.
         # Remove any dead threads from our list
-        for t in self._threads:
-            if not t.isAlive():
-                self._threads.remove(t)
-                amount -= 1
+        amount -= len(self._pending_shutdowns)
+        self._clear_dead_threads()
```
webknjaz: Maybe it's easier to return the amount of cleared threads and use it in that substitution instead? It seems a more consistent way to go...

cheroot/workers/threadpool.py
```diff
@@ -223,6 +234,7 @@ def shrink(self, amount):
         # to remove. As each request is processed by a worker, that worker
         # will terminate and be culled from the list.
         for n in range(n_to_remove):
+            self._pending_shutdowns.append(None)
```
webknjaz: You probably wanted to use the constant here:
```suggestion
            self._pending_shutdowns.append(_SHUTDOWNREQUEST)
```

cheroot/workers/threadpool.py
```diff
@@ -223,6 +234,7 @@ def shrink(self, amount):
         # to remove. As each request is processed by a worker, that worker
         # will terminate and be culled from the list.
         for n in range(n_to_remove):
```
webknjaz: Even better: you can move it out of the loop:
```suggestion
        self._pending_shutdowns.extend((_SHUTDOWNREQUEST, ) * n_to_remove)
        for n in range(n_to_remove):
```

webknjaz:
@the-allanc plz sync this branch with master: i've fixed linter errors there

webknjaz:
I've manually triggered canceled stages in travis

cheroot/workers/threadpool.py
the-allanc: I used your later suggestion.

cheroot/workers/threadpool.py
the-allanc: Well, you meant `if not t.isAlive()`, but I've incorporated that change.

cheroot/workers/threadpool.py
the-allanc: But we're not specifically interested in how many threads are dead at this point, we're only interested how many threads are scheduled to be shutdown. And that figure can also include threads which have already terminated.

cheroot/workers/threadpool.py
the-allanc: No, the object we store in `_pending_shutdowns` is irrelevant. That constant is important in the requests queue.

cheroot/workers/threadpool.py
the-allanc: If it would mean that the loop was no longer necessary, then I would. But it doesn't, so I think it's better to keep the loop, as it reads easier.

the-allanc:
> @the-allanc plz sync this branch with master: i've fixed linter errors there
Done.

cheroot/workers/threadpool.py
webknjaz: Maybe have another named constant for this? I don't like it being a magic value with no meaning.

webknjaz:
Apparently, this PR somehow breaks tests under Python 2.7 under Windows.

webknjaz:
@the-allanc In general, I have no objections except we need AppVeyor CI to be green first.

cheroot/workers/threadpool.py
```diff
@@ -180,8 +183,15 @@ def put(self, obj):
                 waiting to be processed
         """
         self._queue.put(obj, block=True, timeout=self._queue_put_timeout)
-        if obj is _SHUTDOWNREQUEST:
-            return
+
+    def _clear_dead_threads(self):
+        # Remove any dead threads from our list
+        for t in [t for t in self._threads if not t.isAlive()]:
```
webknjaz: `isAlive()` is deprecated: https://travis-ci.org/cherrypy/cheroot/jobs/536951698#L440
```suggestion
        for t in [t for t in self._threads if not t.is_alive()]:
```

jaraco:
This looks good to me. I don't know when I'll have time to give it proper attention, so happy to move forward once the tests are stable/addressed.

cheroot/workers/threadpool.py
webknjaz: @the-allanc WDYT?

cheroot/workers/threadpool.py
```diff
@@ -180,8 +183,15 @@ def put(self, obj):
...
         """
         self._queue.put(obj, block=True, timeout=self._queue_put_timeout)
-        if obj is _SHUTDOWNREQUEST:
-            return
+
+    def _clear_dead_threads(self):
+        # Remove any dead threads from our list
+        for t in [t for t in self._threads if not t.is_alive()]:
+            self._threads.remove(t)
+            try:
```
webknjaz: This is a perfect case for contextlib usage
```suggestion
            with contextlib.suppress(IndexError):
```

webknjaz:
I've fixed AppVeyor on master so it's probably worth rebasing this PR.

cheroot/workers/threadpool.py
the-allanc: It's not a magic value, it's a _meaningless_ value. Naming a constant for it and then not referring to that named constant anywhere else would be more confusing.

I needed something to act as a thread-safe counter, and using a deque seemed to be the most straight-forward option.

cheroot/workers/threadpool.py
the-allanc: Fine with me.

PR #211, opened 2019-07-10
Fixed 'pip install cheroot' on Python 2

marc1n:
  - [x] 🐞 bug fix

On Python 2 "pip install cheroot" ends with error:
ERROR: more-itertools requires Python '>=3.4' but the running Python is 2.7.15

The reason: more_itertools package from version 6 does not support Python 2.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/211)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -63,7 +63,8 @@ setup_requires =
 install_requires =
     backports.functools_lru_cache
     six>=1.11.0
-    more_itertools>=2.6
+    more_itertools>=2.6<6; python_version<'3'
+    more_itertools>=2.6; python_version>='3'
 
 [options.extras_require]
 docs =


marc1n:
There is a problem with test 'test_tls_client_auth[2-False-localhost-builtin]'. I think the test's failure is not related with my change.

setup.cfg
```diff
@@ -63,7 +63,8 @@ setup_requires =
 install_requires =
     backports.functools_lru_cache
     six>=1.11.0
-    more_itertools>=2.6
+    more_itertools>=2.6<6; python_version<'3'
```
webknjaz: The comma is missing
```suggestion
    more_itertools>=2.6,<6; python_version<'3'
```

marc1n:
$ pip --version
pip 18.0 from /home/.../lib/python2.7/site-packages/pip (python 2.7)

marc1n:
I've updated pip to newest version (19.1.1) and issue is still valid.

webknjaz:
Please share the log of that.

webknjaz:
I'm pretty sure you're hitting https://github.com/erikrose/more-itertools/issues/272#issuecomment-462721259.
Could you please share an [asciinema](https://asciinema.org/) recording of the whole process with the Pip version check?

marc1n:
I've found the problem exists if I use my local pypi mirror.
If I 'pip install' from public pypi.org that everything is OK.

marc1n:
Installing from local pypi:
```
$ pip install more-itertools
Looking in indexes: http://pypi.local
Collecting more-itertools
  Downloading http://devpi.local/root/pypi/%2Bf/8bb/43d1f51ecef60/more-itertools-7.1.0.tar.gz (71kB)
     |████████████████████████████████| 71kB 66.9MB/s 
ERROR: more-itertools requires Python '>=3.4' but the running Python is 2.7.15
```

webknjaz:
So the problem is that you don't have a compatible version in your index. Just add proper dists there and you should be fine.

marc1n:
It seems that this is problem of dev-server package (https://github.com/devpi/devpi/issues/511) fixed in 4.7.0 version.

webknjaz:
makes sense

marc1n:
Anyway, thanks for support!

webknjaz:
You're welcome :)

Issue #203, opened 2019-05-18
cheroot on python 3.x is broken in MacPorts due to backports.functools.lru_cache missing


mrdomino:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [X] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**

cheroot on Python 3.x seems to be broken in MacPorts at the moment due to backports.functools_lru_cache only being packaged for Python 2.7 there. It seems based on inspection that their current general approach to backports is to only package them for versions where the functionality is missing. (FYI, I'm not connected with MacPorts; I just use it.)

❓ **What is the motivation / use case for changing the behavior?**

I'd hope to be able to install cheroot from MacPorts.

💡 **To Reproduce**
1. Run `port install py37-cheroot`
2. Create `foo.py` with contents:
```py
__requires__ = 'cheroot'
from pkg_resources import load_entry_point
```
3. Run `python foo.py`

(I first encountered this issue with [fava](https://github.com/beancount/fava); the above is an attempt to minimize the failing case. Importing cheroot directly from the console does not have this issue.)

💡 **Expected behavior**

The script completes successfully.

📋 **Details**

```
Traceback (most recent call last):
  File "foo.py", line 4, in <module>
    from pkg_resources import load_entry_point
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3241, in <module>
    @_call_aside
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3225, in _call_aside
    f(*args, **kwargs)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pkg_resources/__init__.py", line 3254, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    ws.require(__requires__)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/pkg_resources/__init__.py", line 900, in require
    needed = self.resolve(parse_requirements(requirements))
[...]

Issue #181, opened 2019-03-09
cheroot 6.5.4 traceback for python 3.6 complaining about backports.functools_lru_cache requirement not being met


williamh:
❓ **I'm submitting a ...**
- [X] 🐞 bug report

🐞 **Describe the bug. What is the current behavior?**

cheroot crashes with a traceback on python 3.6 on Gentoo Linux, but it shouldn't because the function it needs is in the standard library.

💡 **To Reproduce**

1. Run '...'
2. Make request '....'
3. See error
```python-traceback
whubbs@whubbs1 ~ $ cheroot
Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.6/cheroot", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 3126, in <module>
    @_call_aside
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 3110, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 3139, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 581, in _build_master
    ws.require(__requires__)
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 898, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/usr/lib64/python3.6/site-packages/pkg_resources/__init__.py", line 784, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'backports.functools_lru_cache' distribution was not found and is required by cheroot
whubbs@whubbs1 ~ $ python
Python 3.6.5 (default, Jul 24 2018, 10:46:29)
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from functools import lru_cache
>>>
whubbs@whubbs1 ~ $
```
💡 **Expected behavior**

No crash, because as shown, the function is available.

📋 **Details**

See above.

📋 **Environment**
- Cheroot version: 6.5.4
- Python version: 3.6.5
- OS: Linux

📋 **Additional context**

I am a packager for Gentoo Linux working on upgrading cheroot/cherrypy for us.

williamh:
This pull request is related:
[...]

PR #221, opened 2019-09-02
Only require backports.functools_lru_cache for python < 3.3

dcermak:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

None

cheroot requires the `backports.functools_lru_cache` module even for python versions that ship `functools.lru_cache`

cheroot only requires `backports.functools_lru_cache` for python < 3.3

📋 **Other information**:

📋 **Checklist**:

  - [-] Unit tests for the changes exist
  - [-] Integration tests for the changes exist (if applicable)
  - [-] I used the same coding conventions as the rest of the project
  - [-] The new code doesn't generate linter offenses
  - [-] Documentation reflects the changes
  - [-] The PR relates to *only* one subject with a clear title

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/221)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+v6.6.0
+======
+
+- Revisit :pr:`85` under :pr:`221`. Now ``backports.unittest_mock``
+  is only required on Python 3.2 and earlier.
+
 v6.5.6
 ======
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,8 +1,9 @@
 v6.6.0
 ======
 
-- Revisit :pr:`85` under :pr:`221`. Now ``backports.unittest_mock``
-  is only required on Python 3.2 and earlier.
+- Revisit :pr:`85` under :pr:`221`. Now
+  ``backports.functools_lru_cache`` is only
+  required on Python 3.2 and earlier.
 
 v6.5.8
 ======
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-    backports.functools_lru_cache
+    backports.functools_lru_cache ; python_version < '3.3'
     six>=1.11.0
     more_itertools>=2.6
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,7 +61,7 @@ setup_requires =
 
 # These are required in actual runtime:
 install_requires =
-    backports.functools_lru_cache ; python_version < '3.3'
+    backports.functools_lru_cache; python_version < '3.3'
     six>=1.11.0
     more_itertools>=2.6
 


smithfarm:
@webknjaz Was the "wontfix" label added intentionally?

webknjaz:
@smithfarm it's not broken. You didn't install the dependency.

webknjaz:
I'm leaving this up to @jaraco to decide whether to merge this. It's a duplicate of what he blocked two times already.

smithfarm:
> @smithfarm it's not broken. You didn't install the dependency.

But the dependency is not needed for cheroot to work on recent Python 3, so why have it?

webknjaz:
@smithfarm it's a fallback in runtime. But `pkg_resources` seems to disagree. That dep is noop under Python 3.

webknjaz:
Anyway, I'm almost sure that you didn't use Pip to install Cheroot which means that you must blame your OS packager.

webknjaz:
It's a duplicate of https://github.com/cherrypy/cheroot/issues/181 and https://github.com/cherrypy/cheroot/issues/203

smithfarm:
> Anyway, I'm almost sure that you didn't use Pip to install Cheroot which means that you must blame your OS packager.

Right, I am trying to keep this stuff running in the OS.

If this fix helps the OS maintainers and does not hurt folks using pip, I don't understand why it's being blocked?

webknjaz:
https://github.com/cherrypy/cheroot/commit/ea8acd0786cfb7435011f018897fbfd0f9759415

webknjaz:
Feel free to read all the discussions in linked tickets. But the bottom line is that we only support Pip and separate ourselves from anything that is related to OS cases because it's something that somebody else who is interested in this should maintain.

dcermak:
> But the bottom line is that we only support Pip and separate ourselves from anything that is related to OS cases because it's something that somebody else who is interested in this should maintain.

I fully understand that.

What I don't understand is why you insist on keeping `backports.functools_lru_cache` as a direct dependency for Python versions that ship this module in their standard library? Yes, the installation is not broken with pip (because it just downloads the package but never uses it) and yes we can work around this dependency when creating a rpm package, but that's not the point of this pull request. The point is that you **do not need** `backports.functools_lru_cache` as `lru_cache` is part of `functools` since Python 3.2. So why require it then?

jaraco:
That reversion commit mentions #85.

CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+v6.6.0
+======
+
+- Revisit :pr:`85` under :pr:`221`. Now ``backports.unittest_mock``
```
dcermak:
```suggestion
- Revisit :pr:`85` under :pr:`221`. Now ``backports.functools_lru_cache``
```

jaraco: Added in 2f6199ae

jaraco:
> So why require it then?

Reading up on #85, which goes into detail, the tl;dr is that it's simpler for this project (and all other projects) to unconditionally declare these compatibility libraries and in fact, the backports guidance specifically advises projects to [unconditionally _use_ the backport](https://pypi.org/project/backports/) (see "What if the feature is present"). Cheroot doesn't do that, which is what gives the impression that it's an optional dependency. It's the declared dependencies that are authoritative, so if downstream packagers aren't honoring the declared dependencies, that's a bug in their system.

So let's say we accept this proposal. Does that make it easier for downstream packagers? It arguably makes it more difficult, as now packagers have different dependency trees depending on the target Python version.

Still, since cheroot is conditionally importing this dependency, I guess it's fine to accept to make it easier on the packagers.

dcermak:
> Does that make it easier for downstream packagers?

Thank you for merging this, it actually makes our job easier in opensuse.

PR #224, opened 2019-09-14
Feature/extract open url once

jaraco:
  - [ ] 🐞 bug fix
  - [X] 🐣 feature
  - [ ] 📋 docs update
  - [X] 📋 refactoring
  - [ ] 💥 other

In master, I've separated the concerns of retrying an openURL call from the essential operation of an openURL call. This PR goes a step further and replaces most of the "retry" handling with a proven `retry_call` operation. Re-using this pattern has the advantage of declaring the more abstract goal (of retrying a call) and giving the technique more exposure and re-use.

This change has one backward-incompatible detail - the `raise_subcls` and `ssl_context` can no longer be supplied as a positional parameter, and `None` is no longer a valid value for a degenerate 'raise_subcls` value.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/224)

### DIFF
--- a/cheroot/test/webtest.py
+++ b/cheroot/test/webtest.py
@@ -491,8 +491,6 @@ def openURL(
     a socket.error regardless that they were are subclass of a
     socket.error and therefore not considered for a connection retry.
     """
-    headers = cleanHeaders(headers, method, body, host, port)
-
     # Trying 10 times is simply in case of socket errors.
     # Normal case--it should run once.
     for trial in range(10):
@@ -514,6 +512,8 @@ def _open_url_once(
         body, headers, host, http_conn, method,
         port, protocol, ssl_context, url,
 ):
+    headers = cleanHeaders(headers, method, body, host, port)
+
     # Allow http_conn to be a class or an instance
     if hasattr(http_conn, 'host'):
         conn = http_conn
--- a/cheroot/test/webtest.py
+++ b/cheroot/test/webtest.py
@@ -26,11 +26,13 @@ import os
 import json
 import unittest
 import warnings
+import functools
 
-from six.moves import range, http_client, map, urllib_parse
+from six.moves import http_client, map, urllib_parse
 import six
 
 from more_itertools.more import always_iterable
+import jaraco.functools
 
 
 def interface(host):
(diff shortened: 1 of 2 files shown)


cheroot/test/webtest.py
```diff
@@ -478,42 +477,38 @@ def shb(response):
...
+        retries=9,
+        cleanup=on_exception,
+        trap=socket.error,
+    )
 
 
 def _open_url_once(
-        body, headers, host, http_conn, method,
-        port, protocol, ssl_context, url,
+    url, headers=None, method='GET', body=None,
```
webknjaz: pylint requires this to have more indentation. this is so it's not aligned with the function body.

cheroot/test/webtest.py
```diff
@@ -478,42 +477,38 @@ def shb(response):
...
-        except socket.error as e:
-            if raise_subcls is not None and isinstance(e, raise_subcls):
-                raise
-            else:
-                time.sleep(0.5)
-                if trial == 9:
-                    raise
+    opener = functools.partial(_open_url_once, *args, **kwargs)
+
+    def on_exception():
```
webknjaz: I'd rather have the exponential backoff implemented here...

webknjaz:
I'm not sure how I feel about this.
Honestly, I can't wait till we can deprecate this but it'd require us to roll out a proper pytest fixtures based solution first...

Other than that, I don't have strong opinions here.

P.S. It looks like Python 2 can't parse a kw with a default val in between args and kwargs: https://travis-ci.org/cherrypy/cheroot/jobs/584998920#L331

cheroot/test/webtest.py
jaraco: That's not unreasonable, but it's a feature request outside the scope of this change. This change will make it easier, however, to plug in different back off techniques. Good news is, someone has already written [BackoffDelay](https://tempora.readthedocs.io/en/latest/index.html#tempora.timing.BackoffDelay) for just this purpose.

cheroot/test/webtest.py
jaraco: Linters seem to be passing for me and in CI. If the already nitpicking linters aren't complaining, I'm not going to worry about it.

jaraco:
The test failures look like spurious ones, so proceeding.

jaraco:
Updated the changelog in 5739b6d4.

Issue #214, opened 2019-07-25
Use an API token for uploading to PyPI from Travis CI instead of account creds


webknjaz:
It is now possible to create an API token at https://pypi.org/manage/account/token/. Travis CI autoreleases should switch to that.

❓ **I'm submitting a ...**
- [ ] 🐣 infra improvement

❓ **Do you want to request a *feature* or report a *bug*?**

N/A

❓ **What is the current behavior?**

Account creds used.

💡 **If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

N/A

💡 **What is the expected behavior?**

Token auth used for uploading to PyPI.

❓ **What is the motivation / use case for changing the behavior?**

Security

📋 **Please tell us about your environment:**

N/A

📋 **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

Ref: https://discuss.python.org/t/pypi-security-work-multifactor-auth-progress-help-needed/1042/31

Use `@token` for a username and the token itself for a password when using Twine: https://github.com/pypa/warehouse/issues/994#issuecomment-512634222

jaraco:
I tried doing this but once again ran into the [data too large issue](https://github.com/travis-ci/travis.rb/issues/221) when trying to encrypt the PyPI token for a password field. That means cheroot can't use token-based auth until its travis instance gets a larger key or this project moves away from DPL or DPL provides another mechanism by which to transmit the password (travis-ci/dpl#1101).

jaraco:
FYI, I have installed a TWINE_PASSWORD for cheroot in Travis, so if you wish to switch the release technique to the [one in jaraco/skeleton](https://github.com/jaraco/skeleton/blob/4baec898fdfef6da27653d21fdf223da10b13342/tox.ini#L23-L36), this ticket can be closed.

webknjaz:
@jaraco I don't actually like the idea of having a global var because of the security implications, also I want this to be in the Travis' deployment provider because it looks like a good semantical fit. In the future, I'm thinking of using https://github.com/pypa/gh-action-pypi-publish in a workflow triggered by the `deployment` event but that'll need to wait until GitHub fixes the bug on their side preventing `deployment` workflows from starting.

[...]

PR #226, opened 2019-09-26
Feature/pep517 release

jaraco:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [X] 💥 other

The release process is embedded in a Travis config, so it's difficult to run manually.

In addition to providing for a manual release process (for when the automated process is failing, such as it is now), this technique leverages pep517 to build the package, supporting isolated builds and best practices for packaging.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/226)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -226,8 +226,8 @@ jobs:
     <<: *python_3_7_mixture
     install:
     - python -m pip install tox tox-venv
-    script:
-    - python -m tox -e release
+    env:
+      TOXENV: release
 
 cache:
   pip: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -224,17 +224,10 @@ jobs:
   - <<: *stage_deploy
     if: tag IS present
     <<: *python_3_7_mixture
-    install: skip
-    script: skip
-    deploy:
-      provider: pypi
-      on:
-        all_branches: true
-      user: jaraco
-      password:
-        secure: RAfz06AINvz7bfij/YhfkAreRqamgxS8a6jSRNxntYhtJke3ZszUbIDag8+n1I+G5XT2LnMhHqPNR7Plc+AeMz7VXTuy+b81Li5kse20NYlPhd7mBVmTUpXtqYQashV5J39F4qkATBLznCOrMEomM07VTXjO/o2hmQuXniab2Uo=
-      distributions: dists
-      skip_cleanup: true
+    install:
+    - python -m pip install tox tox-venv
+    script:
+    - python -m tox -e release
 
 cache:
   pip: true
--- a/tox.ini
+++ b/tox.ini
@@ -80,3 +80,18 @@ commands =
     python -m setup checkdocs check --metadata --restructuredtext --strict --verbose
     python -m twine check .tox/dist/*
     python -m setuptools_scm ls
+
+[testenv:release]
+skip_install = True
+deps =
+    pep517>=0.5
+    twine>=1.13
+    path.py
+passenv =
+    TWINE_PASSWORD
+setenv =
+    TWINE_USERNAME = {env:TWINE_USERNAME:__token__}
+commands =
+    python -c "import path; path.Path('dist').rmtree_p()"
+    python -m pep517.build .
+    python -m twine upload dist/*
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,3 @@
 [build-system]
-requires = ["setuptools>=30.3", "wheel", "setuptools_scm>=1.15"]
+requires = ["setuptools>=34.4", "wheel", "setuptools_scm>=1.15"]
+build-backend = "setuptools.build_meta"


jaraco:
This change also fixes #214.

.travis.yml
```diff
@@ -224,17 +224,10 @@ jobs:
...
-    deploy:
-      provider: pypi
-      on:
-        all_branches: true
-      user: jaraco
-      password:
-        secure: RAfz06AINvz7bfij/YhfkAreRqamgxS8a6jSRNxntYhtJke3ZszUbIDag8+n1I+G5XT2LnMhHqPNR7Plc+AeMz7VXTuy+b81Li5kse20NYlPhd7mBVmTUpXtqYQashV5J39F4qkATBLznCOrMEomM07VTXjO/o2hmQuXniab2Uo=
-      distributions: dists
-      skip_cleanup: true
+    install:
```
webknjaz: This is not needed, just add `tox-venv` to `tox.ini` and set
```suggestion
    env:
      TOXENV: release
```

.travis.yml
```diff
@@ -224,17 +224,10 @@ jobs:
...
+    - python -m pip install tox tox-venv
+    script:
```
webknjaz: This won't be necessary too.

tox.ini
```diff
@@ -80,3 +80,18 @@ commands =
...
+    pep517>=0.5
+    twine>=1.13
+    path.py
+passenv =
+    TWINE_PASSWORD
+setenv =
+    TWINE_USERNAME = {env:TWINE_USERNAME:__token__}
+commands =
+    python -c "import path; path.Path('dist').rmtree_p()"
+    python -m pep517.build .
```
webknjaz: I usually prefer having separate build and publish envs. Then we'd do smth like `TOXENV=build-dists,setup-check,release`.

webknjaz:
Now that I released https://github.com/marketplace/actions/pypi-publish, I'd want to start using it. But this would not wait on Travis CI out of the box (yet doable) so I'd probably set a "manual" trigger which would be either a "Release" button on the Checks page (I have a working app for that) or publishing a release @ https://github.com/cherrypy/cheroot/releases/new.

.travis.yml
jaraco: I'm not sure I want to invoke `tox -m tox --notest` or `*python_ssl_openssl_version` or `python -m OpenSSL.debug` during `install`, which is why I chose to override it.

Also, I'm not sure about adding `tox-venv` to `tox.ini`. There may be bootstrapping issues. If `tox` needs a later version or needs to install `tox-venv`, then it will create a `.tox/.tox` env in which to install those packages. If `tox-venv` isn't already installed (or is), that may affect how `.tox/.tox` is created. The only technique I trust is to install `tox-venv` alongside `tox` prior to invoking `.tox`. I'm open to experimenting with other techniques, but I don't want to adopt those risks as part of this change.

.travis.yml
jaraco: Done in 31150f6a. I'm slightly reluctant to use an environment variable for a one-off where a command-line switch will do, as the latter provides a nice example for a user as to what to do to release, whereas setting an environment variable for the duration of a command is harder and varies by environments (shell, OS, etc).

.travis.yml
webknjaz: Okay, so let's change this to installing `tox-venv` and invoking `tox` with `--notest` because `--notest` allows us to clearly separate envs bootstrap (hence the install step) from actually running main commands (script).
Using `env` allows us to unify interactions with tox across the config file.

.travis.yml
```diff
@@ -224,17 +224,10 @@ jobs:
...
-      provider: pypi
-      on:
-        all_branches: true
-      user: jaraco
-      password:
-        secure: RAfz06AINvz7bfij/YhfkAreRqamgxS8a6jSRNxntYhtJke3ZszUbIDag8+n1I+G5XT2LnMhHqPNR7Plc+AeMz7VXTuy+b81Li5kse20NYlPhd7mBVmTUpXtqYQashV5J39F4qkATBLznCOrMEomM07VTXjO/o2hmQuXniab2Uo=
-      distributions: dists
-      skip_cleanup: true
+    install:
+    - python -m pip install tox tox-venv
```
webknjaz:
```suggestion
    - python -m pip install tox tox-venv
    - python -m tox --notest
```

tox.ini
jaraco: Three steps for one artifact? I dunno. That sounds overly complicated. Then you'd have to ensure that the artifacts from each invocation are shared across the three steps... and a manual release would then require the same steps (or at least the build/release steps). I can imagine a few scenarios where it might be useful to have separate steps, but 99% of the use-cases are going to be to invoke the suite of commands, so I'd prefer to keep it simple. And in that 1% of the cases where it's necessary to separate the concerns, one could use `tox --notest` and run the commands incrementally. Why not just keep it simple and concise?

webknjaz: I think it does too much. I often want a command that gives me dists what I can then inspect with other tools. How about something like this https://github.com/pypa/twine/blob/5cb1867/tox.ini#L53-L61 then?

jaraco:
> Now that I released https://github.com/marketplace/actions/pypi-publish, I'd want to start using it. But this would not wait on Travis CI out of the box (yet doable) so I'd probably set a "manual" trigger which would be either a "Release" button on the Checks page (I have a working app for that) or publishing a release @ https://github.com/cherrypy/cheroot/releases/new.

All fine and good. I'm happy to adopt any low-friction workflow you'd like to implement... but it needs to work and have low-friction fallbacks when the automated process fails. Some features of the release flows in the projects I support:

- web browser not required to cut a release (just tag and push)
- command-line not required to cut a release (just tag in a browser)
- one-step release process (just tag)
- easy fallback when automation fails (force deploy in CI if blocked by [rare] flaky tests or run `tox -e release` if that doesn't work).

I want this project to have similar characteristics.

tox.ini
webknjaz: Also, ideally in the future, I'd like to try out a better testing strategy: first build the artifacts, then pip-install from there and completely avoid lib imports from the repo.

.travis.yml
jaraco: Thanks but no thanks. I'm not going to add a command and complication when the default process does the same thing. Better to rely on tox to do what it does best, rather than try to micro-manage it. This change seems counter to best-practices... and I can hardly imagine the benefit.

jaraco:
I'm testing this technique out now with the [7.0 release](https://travis-ci.org/cherrypy/cheroot/builds/590139398).

webknjaz:
Yeah, that could be done. GitHub has a `deployment` event which can be produced via API or a browser app with a button, and workflows can be triggered with that event too.

.travis.yml
webknjaz: I consider it best practice. Seen it used in a lot of places. tox itself uses `--notest` as it was designed for such separation: https://github.com/tox-dev/azure-pipelines-template/blob/master/run-tox-env.yml#L118

.travis.yml
jaraco: Oh, I see. By running `--notest`, you get bootstrapping of the environment(s) before running their commands. That's interesting and does make sense. I'm reluctant to add complication, but I can see the benefit there. Let's explore that after validating this simpler approach.

webknjaz: Yep, that was my point :)

webknjaz: Oh, this discussion got distributed over comments... https://github.com/cherrypy/cheroot/pull/226#discussion_r328815933

tox.ini
jaraco: > I'd like to try out a better testing strategy

Yes, maybe. I'm heavily invested in the test-at-source methodology because it's faster. It's one of the reasons I prefer Python to (your language here). I do acknowledge there's weakness in testing against the source instead of the built artifact... but my instinct there would be to provide additional testing for the build artifact rather than encumber the development process with two sources of truth (source and installed code).

webknjaz: Yeah, I think that for the development purposes it's more comfortable to rely on editable installs. And yes, I do agree that we need more artifact testing in CI than locally. I've been meaning to implement this for a while.
Here's a nice article that Paul published this topic last month: https://blog.ganssle.io/articles/2019/08/test-as-installed.html

jaraco:
[success](https://travis-ci.org/cherrypy/cheroot/jobs/590139419).

PR #243, opened 2019-10-17
Provide TLS-related details via WSGI environment interface

mxii-ca:
  - [x] 🐞 bug fix
  - [x] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Resolves cherrypy/cherrypy#910

The WSGI environment is missing many of the [PEP 3333] [SSL variables].

More [PEP 3333] [SSL variables] are added to the WSGI environment when using the built-in adapter.

This also fixes the assumption that DN components only appear once. (ex. OU)

📋 **Other information**:

Tests added to verify the SSL env generation.

📋 **Checklist**:

[PEP 3333]: https://www.python.org/dev/peps/pep-3333/
[SSL variables]: http://www.modssl.org/docs/2.8/ssl_reference.html#ToC25

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/243)


webknjaz:
Thanks for the PR, @mxii-ca! It'll take some time for us to review this, feel free to ping me or @jaraco if we fail to reply in a week or two.

cheroot/test/test_ssl.py
```diff
@@ -30,6 +30,7 @@
     _get_conn_data,
     _probe_ipv6_sock,
 )
+from ..wsgi import Gateway_10
```
jaraco: I feel uneasy about this change. What is the meaning of this change? Can you confirm this change doesn't affect other tests in this module?

cheroot/ssl/builtin.py
```diff
@@ -102,12 +136,42 @@ def __init__(
         if self.ciphers is not None:
             self.context.set_ciphers(ciphers)
 
+        self._server_env = self.env_cert_dict(
+            'SSL_SERVER', self.parse_cert(certificate, private_key),
+        )
+        if self._server_env:
```
webknjaz: I prefer having "reverse clauses". By doing
```suggestion
        if not self._server_env:
            return
```
you can reduce nesting and improve readability. Please do this here and below.

cheroot/ssl/builtin.py
```diff
@@ -102,12 +136,42 @@ def __init__(
...
+            cert = None
+            with open(certificate, mode='rt') as f:
+                cert = f.read()
+
+            # strip off any keys by only taking the first certificate
+            if all(x in cert for x in {ssl.PEM_HEADER, ssl.PEM_FOOTER}):
+                cert = cert.split(ssl.PEM_HEADER, 1)[-1]
```
webknjaz: Are you trying to do just:
```suggestion
                cert_start = cert.find(ssl.PEM_HEADER)
                cert_end = cert.find(ssl.PEM_FOOTER) + len(ssl.PEM_FOOTER)
                cert = cert[cert_start:cert_end]
```
Splitting and recreating reads cryptic...

cheroot/test/test_ssl.py
webknjaz: That's exactly why I postponed this review... While it looks like the author knows what they're doing, it's more complicated than it looks.

cheroot/ssl/builtin.py
```diff
@@ -102,12 +136,42 @@ def __init__(
...
+    @staticmethod
+    def sni_callback(sock, sni, context):
+        """Handle the SNI callback to tag the socket with the SNI."""
+        sock.sni = sni
+        return None
+
     def wrap(self, sock):
         """Wrap and return the given socket, plus WSGI environ entries."""
+        # Python 3.7+
+        # set the SNU callback here incase the user supplies their own
```
webknjaz:
```suggestion
        # set the SNI callback here incase the user supplies their own
```

cheroot/ssl/builtin.py
```diff
@@ -172,22 +235,155 @@ def get_environ(self, sock):
             'HTTPS': 'on',
             'SSL_PROTOCOL': cipher[1],
             'SSL_CIPHER': cipher[0],
-            # SSL_VERSION_INTERFACE     string  The mod_ssl program version
-            # SSL_VERSION_LIBRARY   string  The OpenSSL program version
+            'SSL_CIPHER_EXPORT': '',
+            'SSL_CIPHER_USEKEYSIZE': cipher[2],
+            'SSL_VERSION_INTERFACE': 'Python %s' % sys.version,
```
webknjaz: This should probably include versions of Cheroot (also available as a server attribute https://github.com/cherrypy/cheroot/blob/38c6338/cheroot/server.py#L1519-L1520) and OpenSSL backend too...

cheroot/ssl/builtin.py
```diff
@@ -172,22 +235,155 @@ def get_environ(self, sock):
...
+            'SSL_VERSION_LIBRARY': ssl.OPENSSL_VERSION,
+            'SSL_CLIENT_VERIFY': 'NONE',
         }
 
+        # Python 3.3+
+        if hasattr(sock, 'compression'):
```
webknjaz: I think it's cleaner to do
```suggestion
        with contextlib.suppress(AttributeError):
```

cheroot/ssl/builtin.py
```diff
@@ -172,22 +235,155 @@ def get_environ(self, sock):
...
+            compression = sock.compression()
+            if compression is not None:
+                ssl_environ['SSL_COMPRESS_METHOD'] = compression
+
+        # Python 3.6+
+        if hasattr(sock, 'session'):
+            ssl_environ['SSL_SESSION_ID'] = sock.session.id.hex()
+        if hasattr(sock.context, 'get_ciphers'):
+            for cip in sock.context.get_ciphers():
+                if cip['name'] == cipher[0] and cip['protocol'] == cipher[1]:
```
webknjaz: it may read easier as
```python
target_cipher = cipher[:2]
```
```suggestion
                if target_cipher == (cip['name'], cip['protocol']):
```

cheroot/ssl/builtin.py
```diff
@@ -172,22 +235,155 @@ def get_environ(self, sock):
...
+        # - SSL_CLIENT_CERT_CHAIN_n
+        # - SRP_USER
+        # - SRP_USERINFO
 
         return ssl_environ
 
+    def parse_cert(self, path, private_key=False):
+        """Parse a certificate."""
+        # Python 3+ Unix, Python 3.5+ Windows
+        if private_key is not False:
```
webknjaz: In Python, it is common to use `None` for arg defaults, not `bool` as this is counterintuitive. 
```suggestion
        if private_key is not None:
```

cheroot/ssl/builtin.py
```diff
@@ -172,22 +235,155 @@ def get_environ(self, sock):
...
+            try:
```
webknjaz: I'd use `contextlib.suppress` if you need to ignore an exception. This also needs a comment explaining such a decision.

Also, that's a huge nested block, let's save some spaces by factoring it out into a private helper method.

cheroot/test/test_ssl.py
```diff
@@ -71,11 +72,12 @@
 )
 
 
-class HelloWorldGateway(Gateway):
+class HelloWorldGateway(Gateway_10):
     """Gateway responding with Hello World to root URI."""
 
     def respond(self):
         """Respond with dummy content via HTTP."""
+        type(self).env = self.get_environ()
```
webknjaz: Why are you doing this?

cheroot/test/test_ssl.py
mxii-ca: sorry, I was lazy in adding tests... while it shouldn't have effected them, I've split it off into a separate fixture so it doesn't introduce scope into the existing tests

cheroot/ssl/builtin.py
mxii-ca: done

cheroot/ssl/builtin.py
mxii-ca: updated to use find

cheroot/ssl/builtin.py
mxii-ca: fixed

cheroot/ssl/builtin.py
mxii-ca: I added cheroot but left out openssl since it is already in the environment variable `'SSL_VERSION_LIBRARY'` immediately below

cheroot/ssl/builtin.py
mxii-ca: agreed. changed

cheroot/ssl/builtin.py
mxii-ca: done

cheroot/ssl/builtin.py
mxii-ca: it was an il-fated attempt to detect parameter passing without the typical `UNSET = object()` 

none of the code calls it without the parameter anyway so I just removed the default

cheroot/ssl/builtin.py
mxii-ca: done

I also refined the exceptions ignored

cheroot/test/test_ssl.py
mxii-ca: it was a shortcut to pass the environment back to the test... i've since changed it to pickle it through the response body

cheroot/ssl/builtin.py
mxii-ca: btw, contextlib.suppress was only introduced in Python 3.4 and not backported to 2.7

cheroot/ssl/builtin.py
```diff
@@ -196,12 +420,24 @@ def env_dn_dict(self, env_prefix, cert_value):
...
-                    env['%s_%s' % (env_prefix, attr_code)] = val
+                    dn_attrs.setdefault(attr_code, [])
+                    dn_attrs[attr_code].append(val)
+
+        env = {
+            env_prefix: ','.join(dn),
+        }
+        for attr_code, values in dn_attrs.items():
+            env['%s_%s' % (env_prefix, attr_code)] = ','.join(values)
+            if len(values) > 1:
```
webknjaz: Plz use "if not...: continue" style in order to reduce the following block indentation

cheroot/ssl/builtin.py
webknjaz: FTR it's a good tone to accept suggested changes via the GitHub UI button because it automatically credits the co-author.

cheroot/ssl/builtin.py
webknjaz: Interesting. We need to check that we don't use it in the other places, then...

cheroot/ssl/builtin.py
```diff
@@ -23,17 +27,32 @@
     except ImportError:
         DEFAULT_BUFFER_SIZE = -1
 
+try:
+    from contextlib import suppress
+except ImportError:
+    # suppress was added in Python 3.4
```
webknjaz: I think that the proper place for the shim would be the `.._compat` module.

cheroot/ssl/builtin.py
mxii-ca: sorry, I wasn't aware of that. thanks for letting me know

cheroot/ssl/builtin.py
mxii-ca: moved

cheroot/ssl/builtin.py
mxii-ca: changed

cheroot/ssl/builtin.py
```diff
@@ -67,19 +70,51 @@ class BuiltinSSLAdapter(Adapter):
...
+        # - OCSP
+        # - caIssuers
+        # - crlDistributionPoints
     }
 
+    # from mod_ssl/pkg.sslmod/ssl_engine_vars.c ssl_var_lookup_ssl_cert_dn_rec
     CERT_KEY_TO_LDAP_CODE = {
         'countryName': 'C',
-        'stateOrProvinceName': 'ST',
+        'stateOrProvinceName': 'ST',  # or 'SP' (SSLeay)
```
webknjaz: Does this comment imply that the current implementation is imperfect?
I'm not sure how helpful it is in this state. Mind putting more details here? Maybe a `# TODO: ` or a `# Note: ` with the explanation of consequences of this.

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
+        # KLUDGE: using an undocumented, private, test method to parse a cert
+        # unfortunately, it is the only built-in way without a connection
+        # as a private, undocumented method, it may change at any time
+        # so be tolerant of *any* possible errors it may raise
+        with suppress(Exception):
+            return ssl._ssl._test_decode_cert(cert)
+
+        return {}
+
+    def env_cert_dict(self, env_prefix, parsed_cert):
```
webknjaz: Since this is a method, it's name should refer to an action rather than a state. Hence it should start with a verb.

Would this be an accurate suggestion?
```suggestion
    def make_env_cert_dict(self, env_prefix, parsed_cert):
```

Any alternatives? (`create_`/`convert_`/`generate_`?)
The goal is to have a semantically relevant identifier.

webknjaz: Oh, and since it's only used from within the instance and isn't a part of the interface contract, it's a good idea to make it private:

```suggestion
    def _make_env_cert_dict(self, env_prefix, parsed_cert):
```

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
+                    client, do_handshake_on_connect=True,
+                    server_side=False,
+                ).getpeercert()
+            finally:
+                thread.join()
+        finally:
+            client.close()
+            server.close()
+
+    def parse_cert(self, cert, private_key):
```
webknjaz: Make it private?

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
+        # not supplied by the Python standard library (as of 3.8)
+        # - SSL_SESSION_RESUMED
+        # - SSL_SECURE_RENEG
+        # - SSL_CLIENT_CERT_CHAIN_n
+        # - SRP_USER
+        # - SRP_USERINFO
 
         return ssl_environ
 
+    def loopback_for_cert(self, cert, private_key):
```
webknjaz: Make it private?

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
+        # mod_ssl 2.1+; Python 3.2+
+        if 'notBefore' in parsed_cert:
+            remain = ssl.cert_time_to_seconds(parsed_cert['notAfter'])
+            remain -= ssl.cert_time_to_seconds(parsed_cert['notBefore'])
+            remain /= 60 * 60 * 24
+            env['%s_V_REMAIN' % (env_prefix,)] = str(int(remain))
+
+        return env
+
+    def env_san_dict(self, env_prefix, cert_value):
```
webknjaz: This name should have a verb.

cheroot/ssl/builtin.py
webknjaz: `self` is only used once. This method could be just a helper function.

webknjaz: This would allow us to unit-test it w/o need to mock any side-effects.

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
+                env.update(self.env_dn_dict(key, value))
+            else:
+                env[key] = str(value)
+
+        # mod_ssl 2.1+; Python 3.2+
+        if 'notBefore' in parsed_cert:
+            remain = ssl.cert_time_to_seconds(parsed_cert['notAfter'])
+            remain -= ssl.cert_time_to_seconds(parsed_cert['notBefore'])
+            remain /= 60 * 60 * 24
+            env['%s_V_REMAIN' % (env_prefix,)] = str(int(remain))
```
webknjaz: Why should this be in hours?

cheroot/_compat.py
```diff
@@ -15,6 +15,21 @@
...
+    from contextlib import suppress
+except ImportError:
+    from contextlib import contextmanager
+
+    @contextmanager
+    def suppress(*exceptions):
+        """Return a context manager that suppresses the `exceptions`."""
+        try:
+            yield
+        except Exception as e:
```
webknjaz: I'm pretty sure that you can do
```suggestion
        except exceptions as e:
```
and drop the rest.

cheroot/ssl/builtin.py
```diff
@@ -102,12 +137,44 @@ def __init__(
...
+            return
+        cert_end += len(ssl.PEM_FOOTER)
+        self._server_env['SSL_SERVER_CERT'] = cert[cert_start:cert_end]
+
     def bind(self, sock):
         """Wrap and return the given socket."""
         return super(BuiltinSSLAdapter, self).bind(sock)
 
+    @staticmethod
+    def sni_callback(sock, sni, context):
```
webknjaz: Let's move this to a separate function too. I don't see any need for keeping it in the class.

cheroot/ssl/builtin.py
```diff
@@ -102,12 +137,44 @@ def __init__(
...
+        """Handle the SNI callback to tag the socket with the SNI."""
+        sock.sni = sni
+        return None
```
webknjaz: This is redundant. `None` is a default return value anyway.

cheroot/ssl/builtin.py
```diff
@@ -102,12 +137,44 @@ def __init__(
...
+
     def wrap(self, sock):
         """Wrap and return the given socket, plus WSGI environ entries."""
+        # Python 3.7+
+        # set the SNI callback here incase the user supplies their own
+        # but don't override their callback
+        with suppress(AttributeError):
+            if ssl.HAS_SNI and self.context.sni_callback is None:
+                self.context.sni_callback = self.sni_callback
```
webknjaz: Looks like this should go to the place where the context is declared, in the `__init__()`

cheroot/ssl/builtin.py
```diff
@@ -173,22 +239,164 @@ def get_environ(self, sock):
...
             'SSL_CIPHER': cipher[0],
-            # SSL_VERSION_INTERFACE     string  The mod_ssl program version
-            # SSL_VERSION_LIBRARY   string  The OpenSSL program version
+            'SSL_CIPHER_EXPORT': '',
+            'SSL_CIPHER_USEKEYSIZE': cipher[2],
+            'SSL_VERSION_INTERFACE': '%s Python/%s' % (
+                HTTPServer.version, sys.version,
+            ),
+            'SSL_VERSION_LIBRARY': ssl.OPENSSL_VERSION,
+            'SSL_CLIENT_VERIFY': 'NONE',
```
webknjaz: What does this mean?

cheroot/test/test_ssl.py
```diff
@@ -95,11 +97,34 @@ def respond(self):
...
+        env = self.get_environ()
+        env.pop('wsgi.errors')
+        env.pop('wsgi.input')
+
+        req = self.req
+        req_uri = bton(req.uri)
+        if req_uri == '/':
+            req.status = b'200 OK'
+            req.ensure_headers_sent()
+            req.write(pickle.dumps(env))
```
webknjaz: Why didn't you just use JSON?

cheroot/test/test_ssl.py
```diff
@@ -95,11 +97,34 @@ def respond(self):
...
+            return
+        return super(WSGIEnvironmentGateway, self).respond()
+
+
+def make_tls_http_server(
+    bind_addr, ssl_adapter, request,
+    gateway=HelloWorldGateway,
```
webknjaz: This is not necessary.

cheroot/test/test_ssl.py
```diff
@@ -375,6 +409,108 @@ def test_tls_client_auth(
...
+        ssl.CERT_REQUIRED,  # server should validate if client cert CA is OK
+    ),
+)
+def test_ssl_env(
+    mocker,
+    tls_wsgi_server, adapter_type,
+    ca, tls_verify_mode, tls_certificate,
+    tls_certificate_chain_pem_path,
+    tls_certificate_private_key_pem_path,
+    tls_ca_certificate_pem_path,
```
webknjaz: plz indent this block one more tab

cheroot/test/test_ssl.py
```diff
@@ -375,6 +409,108 @@ def test_tls_client_auth(
...
+        # pyOpenSSL generates the env before the handshake completes
+        if adapter_type == 'pyopenssl':
+            return
+
+        for key in {'SSL_PROTOCOL', 'SSL_CIPHER'}:
+            assert key in env
+
+        # client certificate env
+        if tls_verify_mode == ssl.CERT_NONE:
+            assert 'SSL_CLIENT_VERIFY' in env
```
webknjaz: move this outside of the if-block to avoid the duplication

cheroot/test/test_ssl.py
```diff
@@ -375,6 +409,108 @@ def test_tls_client_auth(
...
+        )
+        if PY34 and resp.status_code != 200:
+            pytest.xfail(
+                'Python 3.4 has problems with verifying client certs',
+            )
+
+        env = pickle.loads(resp.content)
+
+        # hard coded env
+        assert 'wsgi.url_scheme' in env
```
webknjaz: This line is unnecessary. I'll fail on the next check anyway.

cheroot/test/test_ssl.py
```diff
@@ -95,11 +97,34 @@ def respond(self):
         return super(HelloWorldGateway, self).respond()
 
 
-def make_tls_http_server(bind_addr, ssl_adapter, request):
+class WSGIEnvironmentGateway(Gateway_10):
```
webknjaz: Why don't you just add this logic into `HelloWorldGateway` under a different URL (like `/env`).

cheroot/test/test_ssl.py
```diff
@@ -375,6 +409,108 @@ def test_tls_client_auth(
...
+            pytest.xfail(
+                'Python 3.4 has problems with verifying client certs',
+            )
+
+        env = pickle.loads(resp.content)
+
+        # hard coded env
+        assert 'wsgi.url_scheme' in env
+        assert env['wsgi.url_scheme'] == 'https'
+        assert 'HTTPS' in env
```
webknjaz: Same here

cheroot/test/test_ssl.py
webknjaz: This is unnecessary. This check is already embedded into the next assertion line.

cheroot/test/test_ssl.py
```diff
@@ -375,6 +409,108 @@ def test_tls_client_auth(
...
+
+        # client certificate env
+        if tls_verify_mode == ssl.CERT_NONE:
+            assert 'SSL_CLIENT_VERIFY' in env
+            assert env['SSL_CLIENT_VERIFY'] == 'NONE'
+        else:
+            assert 'SSL_CLIENT_VERIFY' in env
+            assert env['SSL_CLIENT_VERIFY'] == 'SUCCESS'
+
+            assert 'SSL_CLIENT_CERT' in env
```
webknjaz: Checking for keys is unnecessary in tests when we always expect them.

cheroot/ssl/builtin.py
mxii-ca: added a `# NOTE:`

cheroot/ssl/builtin.py
mxii-ca: private and no longer on the class

cheroot/ssl/builtin.py
mxii-ca: private and no longer on the class

cheroot/ssl/builtin.py
mxii-ca: changed to `_make_env_san_dict`

cheroot/ssl/builtin.py
mxii-ca: pulled off the class along with `loopback_for_cert`

cheroot/ssl/builtin.py
mxii-ca: it should be in days as per the Apache mod_ssl documentation:
> SSL_CLIENT_V_REMAIN | string | Number of days until client's certificate expires

i've added a comment to note the timescale

cheroot/ssl/builtin.py
mxii-ca: moved

cheroot/ssl/builtin.py
mxii-ca: changed into a comment to note the intention

cheroot/ssl/builtin.py
mxii-ca: comment added

cheroot/test/test_ssl.py
mxii-ca: no particular reason... changed

cheroot/test/test_ssl.py
mxii-ca: done

cheroot/test/test_ssl.py
mxii-ca: removed

cheroot/test/test_ssl.py
mxii-ca: merged

cheroot/test/test_ssl.py
mxii-ca: removed

cheroot/test/test_ssl.py
mxii-ca: removed

cheroot/test/test_ssl.py
mxii-ca: removed

cheroot/ssl/builtin.py
mxii-ca: moved to a property setter with a comment explanation

cheroot/_compat.py
```diff
@@ -15,6 +15,20 @@
 except ImportError:
     IS_ABOVE_OPENSSL10 = None
 
+# contextlib.suppress was added in Python 3.4
```
webknjaz: I've cherry-picked this shim to master.

mxii-ca: I've rebased this MR

mxii-ca:
@jaraco @webknjaz are there any other changes needed for this MR?

webknjaz:
Yeah... Done is better than perfect, I guess :)

webknjaz:
Thanks, @mxii-ca!

webknjaz:
Released under [v8.3.0](https://pypi.org/project/cheroot/8.3.0/).

Issue #106, opened 2018-07-19
Python int too large to convert to C ssize_t


Dobatymo:
**I'm submitting a ...**
- [X] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Describe the bug. What is the current behavior?**
I found this in my error logs
```
ENGINE OverflowError('Python int too large to convert to C ssize_t',)
Traceback (most recent call last):
  File "C:\Program Files\Python35\lib\site-packages\cheroot\server.py", line 1152, in communicate
    req.respond()
  File "C:\Program Files\Python35\lib\site-packages\cheroot\server.py", line 974, in respond
    self.server.gateway(self).respond()
  File "C:\Program Files\Python35\lib\site-packages\cheroot\wsgi.py", line 136, in respond
    self.write(chunk)
  File "C:\Program Files\Python35\lib\site-packages\cheroot\wsgi.py", line 215, in write
    self.req.send_headers()
  File "C:\Program Files\Python35\lib\site-packages\cheroot\server.py", line 1083, in send_headers
    self.rfile.read(remaining)
  File "C:\Program Files\Python35\lib\site-packages\cheroot\server.py", line 315, in read
    data = self.rfile.read(size)
  File "C:\Program Files\Python35\lib\_pyio.py", line 1000, in read
    return self._read_unlocked(size)
  File "C:\Program Files\Python35\lib\_pyio.py", line 1040, in _read_unlocked
    chunk = self.raw.read(wanted)
OverflowError: Python int too large to convert to C ssize_t
```

**To Reproduce**
I don't know.

**Expected behavior**
This exception should be handled gracefully.

**Environment**
- Cheroot version: 6.3.2.post0 (my log entry is 3 weeks old, but i don't think I updated since then...)
- Python version: Python 3.5.4 (v3.5.4:3f56838, Aug  8 2017, 02:17:05) [MSC v.1900 64 bit (AMD64)] on win32
- OS: Windows Server 2012 R2 x64

webknjaz:
@Dobatymo how big was the payload of HTTP request? Any useful info with regard to this?

webknjaz:
Also, in what way are you running Cheroot? Please include at least a small snippet.

Dobatymo:
I am hosting a simple flask application with cherrypy. The actual cherrpy related code is just a few lines. Tomorrow I can check if I can corelate my other logs to find any info about that request.

Dobatymo:
[...]

PR #247, opened 2019-11-08
Add a test that demonstrates an OverflowError

calebrob6:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#106

The current test server fixture in `test_core.py` set a reasonable default for `max_request_body_size`. There doesn't exist tests that check for behavior as a result of malicious Content-Length headers.

I have removed the setting of `max_request_body_size` in the test server fixtures to reflect the default behavior of servers. I have added a test case for requests with large Content-Length headers that is marked as xfail.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/247)

### DIFF
--- a/cheroot/test/test_core.py
+++ b/cheroot/test/test_core.py
@@ -18,6 +18,7 @@ from cheroot.test import helper
 HTTP_BAD_REQUEST = 400
 HTTP_LENGTH_REQUIRED = 411
 HTTP_NOT_FOUND = 404
+HTTP_REQUEST_ENTITY_TOO_LARGE = 413
 HTTP_OK = 200
 HTTP_VERSION_NOT_SUPPORTED = 505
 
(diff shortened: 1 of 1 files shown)


cheroot/test/test_core.py
```diff
@@ -404,7 +422,6 @@ def testing_server_close(wsgi_server_client):
     """Attach a WSGI app to the given server and preconfigure it."""
     wsgi_server = wsgi_server_client.server_instance
     wsgi_server.wsgi_app = CloseController()
-    wsgi_server.max_request_body_size = 30000000
```
webknjaz: Please don't change unrelated fixtures, this affects other tests.

cheroot/test/test_core.py
```diff
@@ -81,7 +82,6 @@ def testing_server(wsgi_server_client):
     """Attach a WSGI app to the given server and preconfigure it."""
     wsgi_server = wsgi_server_client.server_instance
     wsgi_server.wsgi_app = HelloController()
-    wsgi_server.max_request_body_size = 30000000
```
webknjaz: Rather than changing a fixture that a lot of tests use, please create a separate fixture if your test needs a different config.

calebrob6: My reasoning here is that max_request_body_size is 0 by default, therefore the tests that use this fixture will not be covering a typical case. Furthermore, it doesn't seem like any of the tests in test_core deal with max_request_body_size.

(I'm happy to create a new fixture though.)

cheroot/test/test_core.py
```diff
@@ -262,6 +262,24 @@ def test_content_length_required(test_client):
     assert actual_status == HTTP_LENGTH_REQUIRED
 
 
+@pytest.mark.xfail(reason='https://github.com/cherrypy/cheroot/issues/106')
+def test_large_request(test_client):
+    """Test GET query with maliciously large Content-Length."""
+    # If the server's max_request_body_size is not set (i.e. is set to 0)
+    # then this will result in an Overflow error in the server.
```
webknjaz: It's better to spell it exactly as it happens:
```suggestion
    # then this will result in an `OverflowError: Python int too large to convert to C ssize_t`
    # in the server.
```

webknjaz:
It would be useful to also demonstrate a `MemoryError` too.

cheroot/test/test_core.py
webknjaz: I think at least one of those tests needs it. But the point is that unrelated changes should be submitted via separate PRs.

PR #248, opened 2019-11-13
Fix parsing of the `--bind` option for abstract UNIX sockets

cyraxjoe:
  - [X] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [X] 📋 tests/coverage improvement
  - [ ] 📋 refactoring
  - [ ] 💥 other

It wasn't reported.

The `--bind` option wasn't properly parsed for abstract unix socket.
Something like `cheroot --bind @mysocket` would immediately fail with something like:
```python-traceback
$ python -m cheroot 'myapp.wsgi' --bind @MYSOCK
Keyboard Interrupt: shutting down
  File "/nix/store/nzb7x6kij7mk07m88jwv8m1hhyjzjmb1-python3-3.8.0/lib/python3.8/runpy.py", line 192, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/nix/store/nzb7x6kij7mk07m88jwv8m1hhyjzjmb1-python3-3.8.0/lib/python3.8/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/nix/store/ij4qzqw81lnas4f59g05l3amkh928y4l-python3.8-cheroot-8.2.1/lib/python3.8/site-packages/cheroot/__main__.py", line 6, in <module>
    main()
  File "/nix/store/ij4qzqw81lnas4f59g05l3amkh928y4l-python3.8-cheroot-8.2.1/lib/python3.8/site-packages/cheroot/cli.py", line 234, in main
    raw_args._wsgi_app.server(raw_args).safe_start()
  File "/nix/store/ij4qzqw81lnas4f59g05l3amkh928y4l-python3.8-cheroot-8.2.1/lib/python3.8/site-packages/cheroot/server.py", line 1686, in safe_start
    self.start()
  File "/nix/store/ij4qzqw81lnas4f59g05l3amkh928y4l-python3.8-cheroot-8.2.1/lib/python3.8/site-packages/cheroot/server.py", line 1795, in start
    self.prepare()
  File "/nix/store/ij4qzqw81lnas4f59g05l3amkh928y4l-python3.8-cheroot-8.2.1/lib/python3.8/site-packages/cheroot/server.py", line 1728, in prepare
    info = socket.getaddrinfo(
  File "/nix/store/nzb7x6kij7mk07m88jwv8m1hhyjzjmb1-python3-3.8.0/lib/python3.8/socket.py", line 914, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
OSError: [Errno 16] Device or resource busy
```
Because the server would try to connect using TCP/IP at `MYSOCK` port `None`, given the faulty parsing of the CLI.

The `--bind` option now works for abstract UNIX sockets.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/248)

### DIFF
--- a/cheroot/cli.py
+++ b/cheroot/cli.py
@@ -134,6 +134,7 @@ def parse_wsgi_bind_location(bind_addr_string):
     # with value: "<value>" and port: None
     if bind_addr_string.startswith('@'):
         return AbstractSocket(bind_addr_string[1:])
+
     # try and match for an IP/hostname and port
     match = six.moves.urllib.parse.urlparse('//{}'.format(bind_addr_string))
     try:
@@ -143,6 +144,7 @@ def parse_wsgi_bind_location(bind_addr_string):
             return TCPSocket(addr, port)
     except ValueError:
         pass
+
     # else, assume a UNIX socket path
     return UnixSocket(path=bind_addr_string)
 
--- a/cheroot/cli.py
+++ b/cheroot/cli.py
@@ -64,9 +64,9 @@ class UnixSocket(BindLocation):
 class AbstractSocket(BindLocation):
     """AbstractSocket."""
 
-    def __init__(self, addr):
+    def __init__(self, abstract_socket):
         """Initialize."""
-        self.bind_addr = '\0{}'.format(self.abstract_socket)
+        self.bind_addr = '\0{}'.format(abstract_socket)
 
 
 class Application:
@@ -128,6 +128,12 @@ class GatewayYo:
 
 def parse_wsgi_bind_location(bind_addr_string):
     """Convert bind address string to a BindLocation."""
+    # if the string begins with an @ symbol, use an abstract socket,
+    # this is the first condition to verify, otherwise the urlparse
+    # validation would detect //@<value> as a valid url with a hostname
+    # with value: "<value>" and port: None
+    if bind_addr_string.startswith('@'):
+        return AbstractSocket(bind_addr_string[1:])
     # try and match for an IP/hostname and port
     match = six.moves.urllib.parse.urlparse('//{}'.format(bind_addr_string))
     try:
@@ -137,11 +143,7 @@ def parse_wsgi_bind_location(bind_addr_string):
             return TCPSocket(addr, port)
     except ValueError:
         pass
-
     # else, assume a UNIX socket path
-    # if the string begins with an @ symbol, use an abstract socket
-    if bind_addr_string.startswith('@'):
-        return AbstractSocket(bind_addr_string[1:])
     return UnixSocket(path=bind_addr_string)
 
 
--- a/cheroot/cli.py
+++ b/cheroot/cli.py
@@ -25,7 +25,7 @@ import argparse
 from importlib import import_module
 import os
 import sys
-import contextlib
+import inspect
 
 import six
 
@@ -49,6 +49,7 @@ class TCPSocket(BindLocation):
         Args:
             address (str): Host name or IP address
             port (int): TCP port number
+
         """
         self.bind_addr = address, port
 
@@ -77,12 +78,12 @@ class Application:
         """Read WSGI app/Gateway path string and import application module."""
         mod_path, _, app_path = full_path.partition(':')
         app = getattr(import_module(mod_path), app_path or 'application')
-
-        with contextlib.suppress(TypeError):
-            if issubclass(app, server.Gateway):
-                return GatewayYo(app)
-
-        return cls(app)
+        # verify that `app` is a class before the issubclass verification,
+        # otherwise a TypeError will be rised.
+        if inspect.isclass(app) and issubclass(app, server.Gateway):
+            return GatewayYo(app)
+        else:
+            return cls(app)
 
     def __init__(self, wsgi_app):
         """Initialize."""
--- a/cheroot/cli.py
+++ b/cheroot/cli.py
@@ -25,12 +25,12 @@ import argparse
 from importlib import import_module
 import os
 import sys
-import inspect
 
 import six
 
 from . import server
 from . import wsgi
+from ._compat import suppress
 
 
 __metaclass__ = type
@@ -78,12 +78,12 @@ class Application:
         """Read WSGI app/Gateway path string and import application module."""
         mod_path, _, app_path = full_path.partition(':')
         app = getattr(import_module(mod_path), app_path or 'application')
-        # verify that `app` is a class before the issubclass verification,
-        # otherwise a TypeError will be rised.
-        if inspect.isclass(app) and issubclass(app, server.Gateway):
-            return GatewayYo(app)
-        else:
-            return cls(app)
+        # suppress the `TypeError` exception, just in case `app` is not a class
+        with suppress(TypeError):
+            if issubclass(app, server.Gateway):
+                return GatewayYo(app)
+
+        return cls(app)
 
     def __init__(self, wsgi_app):
         """Initialize."""
--- a/cheroot/test/test_cli.py
+++ b/cheroot/test/test_cli.py
@@ -74,7 +74,6 @@ def test_Aplication_resolve(
         full_path = None
         for p in pkg_name.split('.'):
             full_path = p if full_path is None else '.'.join((full_path, p))
-            print('Full path is {}'.format(full_path))
             monkeypatch.setitem(sys.modules, full_path, mocked_app)
     else:
         monkeypatch.setitem(sys.modules, pkg_name, mocked_app)
--- a/cheroot/test/test_cli.py
+++ b/cheroot/test/test_cli.py
@@ -3,6 +3,7 @@
 # vim: set fileencoding=utf-8 :
 import sys
 
+import six
 import pytest
 
 from cheroot.cli import (
@@ -65,6 +66,17 @@ def test_Aplication_resolve(
     wsgi_app_spec, pkg_name, app_method, mocked_app,
 ):
     """Check the wsgi application name conversion."""
-    monkeypatch.setitem(sys.modules, pkg_name, mocked_app)
+    if six.PY2:
+        # python2 requires the previous namespaces to be part of sys.modules
+        #   (e.g. for 'a.b.c' we need to insert 'a', 'a.b' and 'a.b.c')
+        # otherwise it fails, we're setting the same instance on each level,
+        # we don't really care about those, just the last one.
+        full_path = None
+        for p in pkg_name.split('.'):
+            full_path = p if full_path is None else '.'.join((full_path, p))
+            print('Full path is {}'.format(full_path))
+            monkeypatch.setitem(sys.modules, full_path, mocked_app)
+    else:
+        monkeypatch.setitem(sys.modules, pkg_name, mocked_app)
     expected_app = getattr(mocked_app, app_method)
     assert Application.resolve(wsgi_app_spec).wsgi_app == expected_app
--- a/cheroot/test/test_cli.py
+++ b/cheroot/test/test_cli.py
@@ -3,18 +3,23 @@ from cheroot.cli import (
     parse_wsgi_bind_addr
 )
 
+
 def test_parse_wsgi_bind_location_for_tcpip():
     assert parse_wsgi_bind_addr('192.168.1.1:80') == ('192.168.1.1', 80)
     assert parse_wsgi_bind_addr('[::1]:8000') == ('::1', 8000)
 
+
 def test_parse_wsgi_bind_location_for_unix_socket():
     assert parse_wsgi_bind_addr('/tmp/cheroot.sock') == '/tmp/cheroot.sock'
+
 
 def test_parse_wsgi_bind_addr_for_abstract_unix_socket():
     assert parse_wsgi_bind_addr('@cheroot') == '\0cheroot'
 
+
 def test_Aplication_resolve():
     import sys
+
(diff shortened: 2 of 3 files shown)


cheroot/test/test_cli.py
```diff
@@ -0,0 +1,37 @@
+from cheroot.cli import (
+    Application,
+    parse_wsgi_bind_addr
+)
+
+
+def test_parse_wsgi_bind_location_for_tcpip():
```
webknjaz: Could you plz use `@pytest.mark.parametrize` here?

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,37 @@
...
+    import sys
+
+    class WSGIAppMock:
+        def application(self):
+            pass
+
+        def main(self):
+            pass
+    try:
+        wsgi_app_mock = WSGIAppMock()
```
webknjaz: Plz make a fixture providing this.

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,37 @@
...
+        sys.modules['mypkg.wsgi'] = wsgi_app_mock
```
webknjaz: I think it's best to rely on the builtin `monkeypatch` fixture provided by pytest.

cheroot/cli.py
```diff
@@ -137,11 +143,7 @@ def parse_wsgi_bind_location(bind_addr_string):
             return TCPSocket(addr, port)
     except ValueError:
         pass
-
```
webknjaz: I'd keep this empty line in place.

webknjaz:
Hey Joel!

The change looks okay, but we need linters to pass and please use pytest's way of organizing tests as I outlined in the comments.

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,37 @@
...
+        app = Application.resolve('mypkg.wsgi')
```
webknjaz: Please use `parametrize` for this as well.

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,37 @@
...
+def test_parse_wsgi_bind_location_for_unix_socket():
+    assert parse_wsgi_bind_addr('/tmp/cheroot.sock') == '/tmp/cheroot.sock'
+
+
+def test_parse_wsgi_bind_addr_for_abstract_unix_socket():
+    assert parse_wsgi_bind_addr('@cheroot') == '\0cheroot'
+
+
+def test_Aplication_resolve():
+    import sys
```
webknjaz: Imports like this belong to the top of the module. But maybe you could even avoid it with the use of `monkeypatch` fixture.

webknjaz:
P.S. Running `tox -e pre-commit` does some minor bits of autoformatting automatically.

cyraxjoe:
Hey @webknjaz!

Thank you for your detailed review! 

I'll make the changes later in the day. :+1:

cyraxjoe:
:disappointed:  the idea of using ``sys.modules`` as a way to mock the module loading does not work in python 2. ``import_module`` is implemented differently, it uses ``__import__`` without verifying `sys.modules`.

cyraxjoe:
Note to self: mark the PR as `WIP ` to avoid triggering all the verification unnecessarily. :sweat_smile:

cheroot/cli.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, address, port):
         Args:
             address (str): Host name or IP address
             port (int): TCP port number
+
```
webknjaz: Why is this necessary?

cyraxjoe: I went full _follow the linter_, and that missing blank line caused error `D413` from this list: http://www.pydocstyle.org/en/latest/error_codes.html

webknjaz: Interesting. I guess we'll have to switch this to proper sphinx-recognized params anyway. But that's later.

cheroot/cli.py
```diff
@@ -77,12 +78,12 @@ def resolve(cls, full_path):
...
         app = getattr(import_module(mod_path), app_path or 'application')
-
-        with contextlib.suppress(TypeError):
-            if issubclass(app, server.Gateway):
-                return GatewayYo(app)
-
-        return cls(app)
+        # verify that `app` is a class before the issubclass verification,
+        # otherwise a TypeError will be rised.
+        if inspect.isclass(app) and issubclass(app, server.Gateway):
```
webknjaz: EAFP is more natural to Python. If you worry about the Py2 compat, I could cherry-pick this https://github.com/cherrypy/cheroot/pull/243/files#diff-544070cbdc8d4e10b3faf72ba309abc6R18-R31 shim to master.

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+from cheroot.cli import (
+    Application,
+    parse_wsgi_bind_addr,
+)
+
+
+@pytest.mark.parametrize(
+    'raw_bind_addr, expected_bind_addr', (
+        ('192.168.1.1:80', ('192.168.1.1', 80)),
+        ('[::1]:8000', ('::1', 8000)),
```
webknjaz: This reminds me that we should probably also test catch-all addrs like
```suggestion
        ('[::1]:8000', ('::1', 8000)),
        ('[::]:8000', ('::', 8000)),
        ('12345', (12345)),
```

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+    ),
+)
+def test_parse_wsgi_bind_addr_for_tcpip(raw_bind_addr, expected_bind_addr):
+    """Check the parsing of the --bind option for TCP/IP addresses."""
+    assert parse_wsgi_bind_addr(raw_bind_addr) == expected_bind_addr
```
webknjaz: I'd merge these all similar tests into `parametrize` of the first one. Just make sure to use `pytest.param(..., ..., id=)` (http://doc.pytest.org/en/latest/example/parametrize.html#different-options-for-test-ids / https://docs.pytest.org/en/latest/reference.html#pytest-param)

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+        """Empty custom method (callable) inside the mocked WSGI app.
+
+        It has an empty body because we are expecting to verify that
+        the same method is return no the actual execution of it.
+        """
+
+
+@pytest.mark.parametrize(
+    'wsgi_app_spec, pkg_name, app_method, mocked_app', (
+        ('mypkg.wsgi', 'mypkg.wsgi', 'application', WSGIAppMock()),
```
webknjaz: `WSGIAppMock()` is evaluated during the test module import time. Better use a fixture.

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+    """Check the parsing of the --bind option for UNIX Sockets."""
+    assert parse_wsgi_bind_addr('/tmp/cheroot.sock') == '/tmp/cheroot.sock'
+
+
+def test_parse_wsgi_bind_addr_for_abstract_unix_socket():
+    """Check the parsing of the --bind option for Abstract UNIX Sockets."""
+    assert parse_wsgi_bind_addr('@cheroot') == '\0cheroot'
+
+
+class WSGIAppMock:
```
webknjaz: This is how you make a fixture with pytest:
```suggestion
@pytest.fixture
def wsgi_app():
    """Return a WSGI app stub."""
    class WSGIAppMock:
	    """Mock of a wsgi module."""

	     def application(self):
	        """Empty application method.
	         Default method to be called when no specific callable
	        is defined in the wsgi application identifier.
	         It has an empty body because we are expecting to verify that
	        the same method is return no the actual execution of it.
	        """

	     def main(self):
	        """Empty custom method (callable) inside the mocked WSGI app.
	         It has an empty body because we are expecting to verify that
	        the same method is return no the actual execution of it.
	        """

	return WSGIAppMock()
```

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+@pytest.mark.parametrize(
+    'wsgi_app_spec, pkg_name, app_method, mocked_app', (
+        ('mypkg.wsgi', 'mypkg.wsgi', 'application', WSGIAppMock()),
+        ('mypkg.wsgi:application', 'mypkg.wsgi', 'application', WSGIAppMock()),
+        ('mypkg.wsgi:main', 'mypkg.wsgi', 'main', WSGIAppMock()),
+    ),
+)
+def test_Aplication_resolve(
+    monkeypatch,
+    wsgi_app_spec, pkg_name, app_method, mocked_app,
```
webknjaz: To use the fixture from https://github.com/cherrypy/cheroot/pull/248/files#r347003745:
```suggestion
    wsgi_app_spec, pkg_name, app_method, wsgi_app,
```

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,81 @@
...
+):
+    """Check the wsgi application name conversion."""
+    if six.PY2:
```
webknjaz: This whole monkeypatching can be also moved to the fixture so that only testing remains in the test function: https://github.com/cherrypy/cheroot/pull/248/files#r347003745.

cheroot/test/test_core.py
```diff
@@ -282,8 +282,8 @@ def test_content_length_required(test_client):
 def test_large_request(test_client_with_defaults):
     """Test GET query with maliciously large Content-Length."""
     # If the server's max_request_body_size is not set (i.e. is set to 0)
-    # then this will result in an `OverflowError: Python int too large to convert to C ssize_t`
-    # in the server.
+    # then this will result in an `OverflowError: Python int too large to
+    # convert to C `ssize_t` in the server.
```
webknjaz: Wow, I probably missed this in the previous PR. I'll commit it to master directly because it's completely unrelated to this PR.

cheroot/cli.py
cyraxjoe: Yeah the original approach was a plain `try/except TypeError`  and ignore with a comment on why are we ignoring that exception, but I end up preferring the LBFL in this case. But it seem that the `suppress` context manager is used a lot in the other PR. If that's going to be included, then yeah I could use that here and be done by leaving this fragment as it was.

cheroot/test/test_cli.py
cyraxjoe: Yeah... I was just avoiding doing the extra work of thinking of future possible uses cases and all the possible values that urllib may support.

cheroot/test/test_cli.py
cyraxjoe: BTW that unpacking syntax doesn't work on Python 2.

cheroot/test/test_cli.py
cyraxjoe: The mock instance is directly related to the other params. I did liked the idea of passing it along the others params to express the direct relationship.

For example a future param `('a.b:foo', 'a.b', 'foo')` will break the test, because the mock doesn't define a `foo` method. This is not obvious via the implicit mock injection of pytest.

But I understand if you have a strong opinion on going over the pytest mock injection route.

cheroot/test/test_cli.py
webknjaz: Wow, I guess I'm too addicted to py3 :)

cheroot/test/test_cli.py
webknjaz: In this case, you'd have to deal with this on the fixture level. This is a pytest way.

cheroot/test/test_core.py
webknjaz: This one's in master already.

cheroot/cli.py
webknjaz: It's in master already.

cheroot/test/test_cli.py
```diff
@@ -14,68 +14,77 @@
...
     'raw_bind_addr, expected_bind_addr', (
+        # tcp/ip
         ('192.168.1.1:80', ('192.168.1.1', 80)),
+        # ipv6 ips has to be enclosed in brakets when specified in url form
         ('[::1]:8000', ('::1', 8000)),
+        ('localhost:5000', ('localhost', 5000)),
+        # this is a valid input, but foo gets discarted
+        ('foo@bar:5000', ('bar', 5000)),
+        ('foo', ('foo', None)),
+        ('123456789', ('123456789', None)),
```
webknjaz: How about `:8080`?

webknjaz: and catch-all addresses like `0.0.0.0`/`::`?

cheroot/test/test_cli.py
```diff
@@ -0,0 +1,90 @@
...
+        ('/tmp/cheroot.sock', '/tmp/cheroot.sock'),
+        ('/tmp/some-random-file-name', '/tmp/some-random-file-name'),
+        # abstract sockets
+        ('@cheroot', '\0cheroot'),
+    ),
+)
+def test_parse_wsgi_bind_addr(raw_bind_addr, expected_bind_addr):
+    """Check the parsing of the --bind option.
+
+    Verify some of the supported addresses and the excpected return value.
```
webknjaz: It'd be nice to also have a negative test with invalid data. But that could be done separately.

Issue #190, opened 2019-04-17
Dynamic resizing of cheroot Threadpool 


ppipada:
I'm submitting a ...
- [ ] 🐞 bug report
- [X] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

Currently the Threadpool implementation in cheroot provides a grow and shrink method which does not seem to be called anywhere inside cheroot by default. this also means that the "max" value passed in ThreadPool init is not used.
This results in a static number of threads being spawned at init time with no dynamism in response to number of requests handled by the server. As a defensive measure we generally have to init a sufficiently large number of threads (with an appopriate request_queue_size configuration). 

🐣 **Describe the solution you'd like**
ThreadPool should handle the the min and max num threads values. 
The grow and shrink decisions can be taken at request start time or can be done as a busy loop. Given that cheroot is a excellent production ready, pure python WSGI server, dynamic load handling should come baked in to it rather than an external package.   

📋 **Describe alternatives you've considered**
There is a cherrypy plugin that seems to be written explicitly for doing this: https://pypi.org/project/cherrypy-dynpool/
- This assumes cherrypy usage and does not consider Cheroot as standalone WSGI server. I would like to use cheroot as a WSGI server independent of what WSGI APP it serves (E.g: Cheroot + Flask) 
- This is an external plugin (relying on one more package "dynpool") for building dynamism into a WSGI Server. 

📋 **Additional context**
Cherrypy has a open issue that mentions this as a root cause:
https://github.com/cherrypy/cherrypy/issues/1120

webknjaz:
Hi, I think we're not opposed to this but somebody has to write the code and submit a PR. You are welcome to do so :)

webknjaz:
This may be related to the refactoring @the-allanc is doing now.

the-allanc:
> This may be related to the refactoring @the-allanc is doing now.

It's not really. I think the question is whether we bring in cherrypy_dynpool into the cheroot codebase.

webknjaz:
@the-allanc what I meant is that your change with having a manager entity should make it may easier..

dtantsur:
Hi! I'd like to revive this since it seems to be a serious problem for scalability (you're either limited to a small number of threads or have to run a large number of threads even when nothing is happening).

Which approach would you find acceptable @webknjaz? My quick idea is to update the ThreadPool's `put` and add three new variables:
[...]

PR #256, opened 2019-12-18
WIP: Add a threadpool monitor that can resize the pool at certain frequency

ppipada:
  - [ ] 🐞 bug fix
  - [x] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#190 

https://github.com/cherrypy/cheroot/issues/190

Add a monitor for cheroot threadpool that can resize the pool at a certain predefined frequency. 
As of now the monitor doesn't get invoked unless configure_monitor method of threadpool class doesnt get called. I have not integrated the configure call with `HTTPServer` as of now. 

📋 **Other information**:

I am creating this as an initial PR to get early feedback on the nature of the change and any pointers on code structure, best practices etc. Once the structure of the change is discussed and finalized, I can proceed to further cleanup this PR as necessary. 
One major thing pending here is tests for the `ThreadPool` itself. The pool resizer tests are present.
~~`dynpool.py` and corresponding change has been directly taken from~~
~~https://github.com/tabo/dynpool/blob/master/dynpool.py~~
~~I have copied this file here to avoid an external dependency, but we can take a call on this.~~
Adds [dynpool](https://github.com/tabo/dynpool) as external dependency.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/256)


.gitignore
```diff
@@ -179,3 +179,6 @@ ENV/
 
 # pip wheel dist info
 pip-wheel-metadata/
+
+# vscode
+.vscode/
```
webknjaz: This certainly does not belong to this PR. https://julien.danjou.info/properly-managing-your-gitignore/.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+        function,
+        args=None,
+        kwargs=None,
+        logger=None,
+    ):
+        """Initialize a BackgroundTask."""
+        super(BackgroundTask, self).__init__()
+        self.interval = interval
+        self.function = function
+        self.args = args if args else []
```
webknjaz: Checks for defaults must be `if not None`.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+        self.kwargs = kwargs if kwargs else {}
+        self.running = False
+        self.log = logger or (lambda msg: None)
```
webknjaz: Not sure why this is needed.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+            if not self.running:
+                return
+            try:
+                self.function(*self.args, **self.kwargs)
+            except Exception as e:
+                self.log(
+                    'Error in background task thread function {}:{}.'.format(
+                        self.function,
+                        e,
+                    ), )
```
webknjaz: The point of having a trailing comma is that it's at the very end of the string. It looks weird otherwise.
```suggestion
                    ),
                )
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+                # Quit on first error to avoid massive logs.
+                raise
+
+
+class Monitor(object):
```
webknjaz: Instead of inheriting from `object` explicitly, `__metaclass__ = type` must be used in the beginning of the module. It's implemented this way across all modules consistently.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+    """Monitor a BackgroundTask."""
+
+    callback = None
+    """The function to call at intervals."""
+
+    frequency = 60
+    """The time in seconds between callback runs."""
+
+    thread = None
+    """A :class:`BackgroundTask` thread.
```
webknjaz:
```suggestion
    """A :class:`BackgroundTask` thread."""
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+    """
```
webknjaz: I'm pretty sure there's no reason to keep this on a separate line. Plus you're missing an empty line here.
```suggestion

```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+        self.frequency = frequency
+        self.thread = None
+        self.name = name or self.__class__.__name__
+        self.log = logger or (lambda msg: None)
+
+    def start(self):
+        """Start our callback in its own background thread."""
+        if self.frequency > 0:
+            threadname = self.name
+            if self.thread is None:
```
webknjaz: Please invert the clauses and interrupt the method early. This will save unnecessary nesting. Linear flow is easier to read.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+            else:
+                self.log(
+                    'Monitor thread {} already started.'.format(threadname), )
+
+    start.priority = 70
+
+    def stop(self):
+        """Stop our callback's background task thread."""
+        if self.thread is None:
+            self.log('No thread running for {}.'.format(self.name))
```
webknjaz: Interrupting early allows dedenting the following block
```suggestion
            self.log('No thread running for {}.'.format(self.name))
            return
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,155 @@
...
+                    logger=self.log,
+                )
+                self.thread.setName(threadname)
+                self.thread.start()
+                self.log('Started monitor thread {}.'.format(threadname))
+            else:
+                self.log(
+                    'Monitor thread {} already started.'.format(threadname), )
+
+    start.priority = 70
```
webknjaz: What is the priority doing here?

cheroot/workers/threadpool.py
```diff
@@ -3,7 +3,6 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-
```
webknjaz: Please don't add completely unrelated changes. This makes it harder to do reviews.

cheroot/workers/threadpool.py
```diff
@@ -52,7 +52,6 @@ class WorkerThread(threading.Thread):
     ready = False
     """A simple flag for the calling server to know when this thread
     has begun polling the Queue."""
-
```
webknjaz: unrelated + violates the style.

cheroot/workers/threadpool.py
```diff
@@ -147,8 +146,13 @@ class ThreadPool:
     """
 
     def __init__(
-            self, server, min=10, max=-1, accepted_queue_size=-1,
-            accepted_queue_timeout=10,
+        self,
```
webknjaz: The indentation is broken. It shouldn't match the method body nesting.

cheroot/workers/threadpool.py
```diff
@@ -170,6 +174,29 @@ def __init__(
         self._queue_put_timeout = accepted_queue_timeout
         self.get = self._queue.get
         self._pending_shutdowns = collections.deque()
+        self.log = logger or (lambda msg: None)
+        self.monitor = None
+        self.monitor_configured = False
+
+    def configure_monitor(
+        self,
```
webknjaz: pylint will be unhappy with this.

setup.cfg
```diff
@@ -96,6 +96,8 @@ testing =
     coverage==4.5.3
     # send test coverage to codecov.io
     codecov==2.0.15
+    # use mock
+    backports.unittest_mock
```
webknjaz: This shouldn't be used. Pytest way is to use fixtures and there's already `pytest-mock` integrated, it must be used for consistency.

.gitignore
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Intention: 
Wanted a logger to be passed to the monitor and use that. 
If input logger is not provided add a none logger. 
Added this way as CherryPy seemed to be doing it this way. 

Question: Any specific way I should handle this?

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/monitor.py
ppipada: Removed.

cheroot/workers/threadpool.py
ppipada: Seems like this was either done by precommit hook or flake8 autoformatter used for satisfying precommit hook. Will check again.

cheroot/workers/threadpool.py
ppipada: Same as above.
Seems like this was either done by precommit hook or flake8 autoformatter used for satisfying precommit hook. Will check again.

cheroot/workers/threadpool.py
ppipada: Again seems like flake8 autoformatter. Do we have any specific autoformatter that we recommend or use?

cheroot/workers/threadpool.py
ppipada: Can you please elaborate? any specific thing/pylint error to look out for? When I run pylint on current code there seems to be lot of W,C,R issues to be present.

setup.cfg
ppipada: Done.

cheroot/workers/threadpool.py
ppipada: Added line.

cheroot/workers/threadpool.py
ppipada: removed line

ppipada:
> Honestly, I'm still unsure to which extent we need to change Cheroot. It starts looking like feature creep. Also, it seems like you're trying to bring in things that better belong to CherryPy's abstraction level.

This assumes cherrypy usage and does not consider Cheroot as standalone WSGI server. I would like to use cheroot as a WSGI server independent of what WSGI APP it serves (E.g: Cheroot + Flask).
Also, given that ThreadPool in cheroot takes min and max as params, shouldn't the consumer assume that the pool manages its own threads and respects these values?

webknjaz:
> shouldn't the consumer assume that the pool manages its own threads and respects these values?

Well, you're turning a simple web-server into the load-balancer. So I'm still unsure if that complexity should be a part of Cheroot or remain a separate wrapper.

cheroot/workers/dynpool.py
```diff
@@ -0,0 +1,277 @@
+# -*- coding: utf-8 -*-
+"""Dynamic pool.
+
+`dynpool <https://tabo.pe/projects/dynpool/>`_ is a Python
+library that handles the growing and shrinking of a pool
+of resources depending on usage patterns,
+written by `Gustavo Picón <https://tabo.pe/>`_ and
+licensed under the Apache License 2.0.
```
webknjaz: I don't know if Apache 2.0 is compatible with BSD-3-Clause. You may need to acquire the permission for vendoring from @tabo.

webknjaz:
Also, I didn't realize that you're vendoring `dynpool`, not `cherrypy-dynpool`. So since it doesn't depend on CherryPy I don't see any point in vendoring at all. We shouldn't introduce more maintenance burden here.

cheroot/workers/threadpool.py
```diff
@@ -147,8 +148,8 @@ class ThreadPool:
     """
 
     def __init__(
-            self, server, min=10, max=-1, accepted_queue_size=-1,
```
webknjaz: This indentation change is incorrect.

cheroot/workers/monitor.py
webknjaz: As you may notice, there's no standard way to use logging in Cheroot. I'd be interested in adding logging support across the whole Cheroot in a consistent manner. But please don't add it in this PR as it's unrelated to its main goal and debating on how to implement it properly will block the whole effort so this has to be separate.

cheroot/workers/threadpool.py
```diff
@@ -170,6 +171,24 @@ def __init__(
         self._queue_put_timeout = accepted_queue_timeout
         self.get = self._queue.get
         self._pending_shutdowns = collections.deque()
+        self.log = logger or (lambda msg: None)
+        self.monitor = None
+        self.monitor_configured = False
+
+    def configure_monitor(
+        self, thread_pool, monitor_freq=5, minspare=5, maxspare=15,
```
webknjaz:
```suggestion
            self, thread_pool, monitor_freq=5, minspare=5, maxspare=15,
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,154 @@
...
+        self.log = logger or (lambda msg: None)
+        # default to daemonic
+        self.daemon = True
+
+    def cancel(self):
+        """Cancel the task."""
+        self.running = False
+
+    def run(self):
+        """Run - Invoke function every interval seconds."""
```
webknjaz:
```suggestion
        """Invoke the callable with the given interval."""
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,154 @@
...
+    ):
+        """Initialize a BackgroundTask."""
+        super(BackgroundTask, self).__init__()
+        self.interval = interval
+        self.function = function
+        self.args = args if args is not None else []
+        self.kwargs = kwargs if kwargs is not None else {}
+        self.running = False
+        self.log = logger or (lambda msg: None)
+        # default to daemonic
```
webknjaz: This is too obvious to make it to a comment. Comments must explain why and not rephrase the code exactly as it reads.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,154 @@
...
+
+class BackgroundTask(threading.Thread):
+    """A subclass of threading.Thread whose run() method repeats.
+
+    Use this class for most repeating tasks. It uses time.sleep() to wait
+    for each interval, which isn't very responsive; that is, even if you call
+    self.cancel(), you'll have to wait until the sleep() call finishes before
+    the thread stops. To compensate, it defaults to being daemonic, which means
+    it won't delay stopping the whole process.
+    """
```
webknjaz: It'd be useful to explain args with Sphinx syntax following this example: https://github.com/cherrypy/cheroot/blob/85a190f/cheroot/server.py#L431-L441

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,154 @@
...
+            name=name,
+            logger=logger,
+        )
+
+    def run(self):
+        """Run the monitor."""
+        self._run()
+
+    def configure(self, thread_pool, minspare, maxspare, shrinkfreq, logfreq):
+        """Configure the pool resizer."""
```
webknjaz: It'd be useful to explain args with Sphinx syntax following this example: https://github.com/cherrypy/cheroot/blob/85a190f/cheroot/server.py#L1413-L1418

cheroot/workers/monitor.py
ppipada: Got it. Removing logging from here.

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: Done.

cheroot/workers/threadpool.py
ppipada: Done.

cheroot/workers/dynpool.py
ppipada: Removed dynpool vendoring.

cheroot/workers/threadpool.py
ppipada: Done.

cheroot/workers/threadpool.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: removed

ppipada:
> Well, you're turning a simple web-server into the load-balancer. So I'm still unsure if that complexity should be a part of Cheroot or remain a separate wrapper.

Managing threads in the cheroot threadpool could be done in following ways:
- Write an external wrapper in an external repository similar to cherrypy-dynpool.
This would leave the current threadpool as is i.e it provides interfaces for its growth and shrink, but leaves the decision on when and how to do this to the consumer. 
This would also mean that the consumers of cheroot have to know that pool doesnt manage its threads and would have to look out for such a wrapper 
- Write a separate monitor package in the cheroot repo. 
This option is slightly different from the current PR. 
It would provide a monitor thread as a plugin with threadpool not knowing about it. 
Threadpool would continue providing the grow and shrink interfaces but would be unaware of its usage. 
The monitor thread would need to be integrated at a higher level i.e either at the server or some controller/manager entity. 
It would still leave the decision on whether to use it to user discretion but provide an option of using such a thing within cheroot itself via simple configuration.
- Threadpool manages its own threads. (This PR).
This abstracts the management of threads from consumers of the pool.
The consumers provide appropriate configuration to pool itself without knowing about how the pool manages its threads. 
I have gone with this option as a threaded webserver not managing its threads dynamically seems unintuitive to me. 

I would still argue for cheroot going with either option 2 or 3. 
Any case, once we take a decision on how to proceed, I can work on this PR accordingly.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def cancel(self):
+        """Cancel the task."""
+        self.running = False
+
+    def run(self):
+        """Invoke the callable with the given interval."""
+        self.running = True
+        while self.running:
+            time.sleep(self.interval)
```
webknjaz: This inserts a delay before the first invocation. I think it'd be more intuitive to wait after it's called. Also, this would remove the need to do the following check.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+        self.stop()
+        self.start()
+
+
+class ThreadPoolMonitor(Monitor):
+    """ThreadPoolMonitor for dynamic resizing."""
+
+    def __init__(self, frequency, name=None):
+        """Initialize ThreadPoolMonitor."""
+        self._run = lambda: None
```
webknjaz: Not sure if a dummy callable is correct semantically. It looks like just `None` could be better.

webknjaz:
@ppipada I would like to know what @jaraco thinks so let's wait for him to reply before we proceed.

jaraco:
Thanks for putting together a proposed solution to the lack of thread-pool dynamism in a cheroot threadpool.

Of the three options you present, Option 1 seems like the least intrusive and also aligns with the pattern of CherryPy. On the other hand, as you indicate, this behavior is fairly integral to the behavior (after all, cheroot provides the grow/shrink hooks). The main reason I'd be reluctant to accept an integrated solution is that I've seen services fall over due to over-provisioning of threads or other unintentional dynamic pool behaviors, so I'd be cautious about enabling those by default or encouraging use without understanding the implications.

Have you considered what impact these changes would have on a CherryPy client (with or without CherryPy-dynpool)? What would happen if they enabled dynamic pool management in both cheroot and cherrypy? Would you expect CherryPy users to transition to the cheroot-backed dynamic pool or for someone to maintain both dynamic pools?

ppipada:
Thanks @jaraco and @webknjaz for your replies.
Sorry for the late reply. I was on a vacation and hence couldn't reply earlier. 

Few comments below:
>> Have you considered what impact these changes would have on a CherryPy client (with or without CherryPy-dynpool)? What would happen if they enabled dynamic pool management in both cheroot and cherrypy? Would you expect CherryPy users to transition to the cheroot-backed dynamic pool or for someone to maintain both dynamic pools?

Enabling/Configuring the threadpool is an "explicit" action on the threadpool.
As mentioned in the description of PR, I have not done the explicit configure integration in HTTPServer or any controller class as of now. This is majorly because I am not sure what would be a correct configure entrypoint for the threadpool or if this is needed.

Given that this is an explicit action, there is no impact on existing clients as it is be disabled by default. Clients could choose to keep on using existing cherrypy-dynpool or migrate to this solution depending on migration effort. This PR would enable clients not using cherrypy-dynpool or those using cheroot directly (without cherrypy) to explicitly use this threadpool via a configuration step.

Would it be better if this configuration step is done in a different manner than in this PR? 
This PR does a direct instantiation of the ThreadPoolMonitor object in ThreadPool itself. I understand that this reduces pluggabality for the monitor.
ThreadPool could just declare the required interface for monitor objects and the current solution could be one of the available Monitors. (I would really like to see if a n-pending-requests based monitor can be done at some point.)

cheroot/workers/monitor.py
ppipada: This callable is bound dynamically after the configure step. 
the monitor run method calls this directly i.e irrespective of configured or not; hence the callable.

cheroot/workers/monitor.py
ppipada: Not sure if I understand fully. Are you suggesting we do the following:
`
    def run(self):
        """Invoke the callable with the given interval."""
        self.running = True
        while self.running:
            # Any error in the calling function will be raised immediately
            self.function(*self.args, **self.kwargs)
            time.sleep(self.interval)
`

webknjaz: Yes, that seems to be more responsive.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+        self.name = name or self.__class__.__name__
+
+    def start(self):
+        """Start our callback in its own background thread."""
+        if self.frequency <= 0:
+            return
+        if self.thread is not None:
+            return
+        self.thread = BackgroundTask(self.frequency, self.callback)
+        self.thread.setName(self.name)
```
webknjaz: I think it's reasonable to move it inside `BackgroundTask.__init__()` because the initializer should produce a ready-to-use object rather than requiring extra init steps.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+        """Initialize Monitor."""
+        self.callback = callback
+        self.frequency = frequency
+        self.thread = None
+        self.name = name or self.__class__.__name__
+
+    def start(self):
+        """Start our callback in its own background thread."""
+        if self.frequency <= 0:
+            return
```
webknjaz: This should be 
```suggestion
            raise RuntimeError('The frequency must be a positive number')
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+        if self.thread is not None:
+            return
```
webknjaz: Same here — it's not something that should be ignored.
```suggestion
            raise RuntimeError('Only one background task can be monitored')
```

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,131 @@
...
+
+    callback = None
+    """The function to call at intervals."""
+
+    frequency = 60
+    """The time in seconds between callback runs."""
+
+    thread = None
+    """A :class:`BackgroundTask` thread."""
+    def __init__(self, callback, frequency=60, name=None):
```
webknjaz:
```suggestion

    def __init__(self, callback, frequency=60, name=None):
```

ppipada: Done

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
ppipada: Done

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,135 @@
...
+    the thread stops. To compensate, it defaults to being daemonic, which means
+    it won't delay stopping the whole process.
+
+    :param int interval: time interval between invocation of input function
+    :param function: callable that needs to be invoked
+    :param args: list of arguments to pass to the function
+    :param kwargs: dictionary of keyword args to pass to the function
+    """
+
+    def __init__(self, interval, function, name=None, args=None, kwargs=None):
```
webknjaz: So I have two thoughts about this:
1) `function` arg is tightly bound to `args` and `kwargs`, so putting `name` in the middle seems counterintuitive from the API perspective. So maybe do something like this:
```suggestion
    def __init__(
        self, interval,
        function, args=None, kwargs=None,
        name=None,
    ):
```
2) On the second thought, maybe it's a better API if this would only accept a callable. This way, end-users would have to pass a `functools.partial()` instance or `lambda`, or just some wrapper. This may look cleaner.

cheroot/workers/monitor.py
```diff
@@ -0,0 +1,135 @@
...
+        """Stop our callback's background task thread."""
+        if self.thread is None:
+            return
+        if self.thread is not threading.currentThread():
+            self.thread.cancel()
+            if not self.thread.daemon:
+                self.thread.join()
+        self.thread = None
+
+    def graceful(self):
```
webknjaz: Why is this called graceful and how is it supposed to be used?

cheroot/workers/monitor.py
webknjaz: I don't see any place where any args/kwargs are passed — it doesn't seem like they are needed

webknjaz:
> Would it be better if this configuration step is done in a different manner than in this PR?
> This PR does a direct instantiation of the ThreadPoolMonitor object in ThreadPool itself. I understand that this reduces pluggabality for the monitor.

I don't yet have an opinion on this so meanwhile trying to reduce unnecessary code and look for more obvious mistakes.

cheroot/workers/monitor.py
ppipada: 1. This class is derived from threading.Thread whose init looks like: 
```
class Thread(group=None, target=None, name=None, args=(), kwargs=None, daemon=None)
```
Hence the name between target and its args in this class too. 
2. as of now args and kwargs are not used, but given that this is a generic BT class and can be used for any func callable, isn't it better to have the args and kwargs support?

cheroot/workers/monitor.py
ppipada: This is in line with cherryPy states for a background activity. It denotes a graceful restart/reload activity. Not used here as of now, but kept it to be consistent. 
https://github.com/cherrypy/cherrypy/blob/e77f0830754b1eb86ee53d82d10e2c76fe31cae6/cherrypy/process/plugins.py#L577

ppipada:
@webknjaz @jaraco How do we want to take this forward? 
Is there anything specific pending from my side?

cheroot/workers/monitor.py
webknjaz: The purpose of having the server in a separate library is so that we don't copy-over unnecessary code from CherryPy keeping it separate. Not sure if this one is needed here. But I'll be happy to hear @jaraco's opinion too...

webknjaz:
@ppipada nothing specific atm. This change requires very careful consideration.

webknjaz:
recheck

Issue #252, opened 2019-11-28
Dropping support for Python 2 and Python 3 below 3.5


webknjaz:
🕒 It's time.

I'd like to implement dropping the support on Dec 31st, 2019 because I still hope to find some time to improve a few things that I'd like to make it into py2-supported version.

❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [X] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Is your feature request related to a problem? Please describe.**

Maintenance burden. Inability to use new language features.

🐣 **Describe the solution you'd like**

Drop support for old Python versions.

📋 **Describe alternatives you've considered**

Keep suffering?

📋 **Additional context**

https://python3statement.org

jaraco:
There's nothing to stop someone (especially maintainers) from contributing to cheroot prior to the 3.x-only release. We can even create a `maint/8.x` branch to mirror what CherryPy has for LTS.

webknjaz:
Current status: I want to do this after #262 and #243. And maybe #256 if we can decide to go for it.

webknjaz:
TODO: drop cryptography deprecation warning ignores from `pytest.ini` once this is done.

Safihre:
Or, just drop 2.7.. The core usage of cheroot is still cherrypy, since cherrypy dropped support, why bother many months more? :)

jaraco:
> Or, just drop 2.7.. The core usage of cheroot is still cherrypy, since cherrypy dropped support, why bother many months more? :)

Good question. One case is that older versions of CherryPy still can and do pick up the latest cheroot.

I'd like to drop support for sunset Pythons (3.5, 2.7) ASAP, but I want cheroot to be stable. There were a couple of regressions identified with recent releases and I want to be sure those are addressed. Looking at the changelog, it does look as if several regressions were fixed. I'll scan the issues and see if there are any outstanding we need to address (or roll back).

> Current status: I want to do this after #262 ~and #243~. And maybe #256 if we can decide to go for it.

At one point, getting these features in for Python 2 support had merit, but it feels like that time has passed. Let's freeze feature development for 8.x and focus on getting 9.x out for Python 3.6+.

webknjaz:
[...]

PR #262, opened 2020-01-22
WIP: Defer HTTP parsing to Hyper (h11 project)

### LINKED ISSUE #201: Defer HTTP/1.1 parsing to Hyper (h11 project)
I suggest merging our HTTP implementation with Hyper/H11: https://github.com/python-hyper/h11 / https://h11.readthedocs.io/.

This would contribute to having a well-made parser shared with others.

I've also declared the intent @ https://github.com/python-hyper/h11/issues/84

❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [X] 🐣 feature request

Currently, CherryPy team has to maintain its own HTTP parser implementation which doesn't scale.

🐣 **Describe the solution you'd like**

We should contribute to one well-designed upstream parser implementation instead of maintaining our own thing. Shared responsibility is the key to having a healthy open source project :)

📋 **Describe alternatives you've considered**

N/A

📋 **Additional context**

Hyper is a bring-your-own-I/O set of principles encouraging users to rely on protocol implementations which do not hit the network. It also potentially offers testability improvements.

It'll also prevent problems like https://github.com/cherrypy/cheroot/pull/74

ian-otto:
  - [ ] 🐞 bug fix
  - [X] 🐣 feature
  - [ ] 📋 docs update
  - [X] 📋 refactoring
  - [ ] 💥 other

Resolves #201 

Currently, the Cheroot project has its own HTTP parser and logic.

This PR will defer the primary HTTP parser to the h11 project's implementation.

📋 **Other information**:

Depends on: https://github.com/python-hyper/h11/pull/99

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/262)


duper51:
Something of note, is I'm not sure whether I should anticipate supporting Python 2.7 specifically in this PR as it appears there is a discussion of its deprecation (#252)

cheroot/server.py
```diff
@@ -1146,8 +1196,8 @@ def send_headers(self):
                 pass
             else:
                 needs_chunked = (
```
njsmith: You don't need to handle this yourself anymore – if you don't have a `Content-Length`, then h11 will automatically choose between HTTP/1.1-style chunked encoding and HTTP/1.0-style connection-termination for you based on the capabilities of the peer, and add/remove `Transfer-Encoding: chunked` and `Connection: close` headers as appropriate.

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+        req_line = self._next_event()
+        if isinstance(req_line, h11.Request):
+            self.started_request = True
+            self.uri = req_line.target
+            scheme, netloc, path, query, fragment = urllib.parse.urlsplit(self.uri)
+            self.path = path
+            self.qs = query
+            self.method = req_line.method
+            self.request_protocol = b"HTTP/%s" % req_line.http_version
+            self.response_protocol = b"HTTP/%s" % req_line.http_version
```
njsmith: This is your problem – you should always use HTTP/1.1 in the response. That works fine even if the client is speaking HTTP/1.0, because it's forwards/backwards compatible.

cheroot/server.py
```diff
@@ -469,14 +523,14 @@ def _fetch(self):
         except ValueError:
             raise ValueError(
                 'Bad chunked transfer size: {chunk_size!r}'.
-                format(chunk_size=chunk_size),
-            )
+                    format(chunk_size=chunk_size),
```
webknjaz: Please don't change the formatting here. It was good originally, just like I intended — with closing bracket matching the initial block indentation.

cheroot/server.py
```diff
@@ -469,14 +523,14 @@ def _fetch(self):
...
+                    )
 
         if chunk_size <= 0:
             self.closed = True
             return
 
-#            if line: chunk_extension = line[0]
+        #            if line: chunk_extension = line[0]
```
webknjaz: Please don't change formatting because it adds unnecessary noise to the PR. It's easier to review if only related bits show up in the diff.

cheroot/server.py
```diff
@@ -490,7 +544,7 @@ def _fetch(self):
             raise ValueError(
                 "Bad chunked transfer coding (expected '\\r\\n', "
                 'got ' + repr(crlf) + ')',
-            )
+                )
```
webknjaz: (same here — plz don't reformat)

cheroot/server.py
```diff
@@ -838,9 +888,9 @@ def read_request_line(self):
             # invalid URIs without raising errors
             # https://tools.ietf.org/html/rfc7230#section-5.3.3
             invalid_path = (
-                _authority != uri
-                or not _port
-                or any((_scheme, _path, _qs, _fragment))
+                    _authority != uri
```
webknjaz: Why did you decide to do this?

cheroot/server.py
```diff
@@ -340,8 +341,65 @@ def __next__(self):
 
     next = __next__
 
+class RFile:
+    def __init__(self, rfile):
+        self.rfile = rfile
+
+    def read(self, size=None):
```
webknjaz: Could you please use ABCMeta and mark this as an `@abstractmethod`? (and all the other methods too, of course)

cheroot/server.py
```diff
@@ -340,8 +341,65 @@ def __next__(self):
...
+        return event.data
+
+    def readlines(self, sizehint=None):
+        if self.conn.their_state != h11.SEND_BODY:
+            return b""
+        lines = []
+        line = self.rfile.readline()
+        while line:
+            event = self.conn.next_event()
+            if event is h11.NEED_DATA:
```
webknjaz: plz reverse this check in order to reduce the following block indentation
```suggestion
            if event is not h11.NEED_DATA:
                continue
```

cheroot/server.py
```diff
@@ -340,8 +341,65 @@ def __next__(self):
...
+                line = self.rfile.readline()
+                self.conn.receive_data(line)
+                event = self.conn.next_event()
+                assert isinstance(event, h11.Data)
```
webknjaz: What is the purpose of this assertion?

cheroot/server.py
```diff
@@ -340,8 +341,65 @@ def __next__(self):
...
+        self.conn = conn
+
+    def read(self, size=None):
+        if self.conn.their_state != h11.SEND_BODY or size == 0:
+            return b""
+        event = self.conn.next_event()
+        if event is h11.NEED_DATA:
+            self.conn.receive_data(self.rfile.read(size))
+            event = self.conn.next_event()
+        assert isinstance(event, h11.Data)
```
webknjaz: Is this some sort of an exit-check for ensuring contracts?

webknjaz:
> I'm not sure whether I should anticipate supporting Python 2.7 specifically in this PR as it appears there is a discussion of its deprecation

@duper51 I wanted to deprecate Python 2.7 after the HTTP parser replacement so that the last py2-compatible release would get the changes and then we'd start dropping support for this. Will probably merge #243 too before that. And maybe even #256 if we agree on it. That said, this PR should also be tested against CherryPy 17 which is the last major version to support Python 2.

@njsmith thanks a lot for your comments here!

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
         self.conn.wfile.write(EMPTY.join(buf))
 
 
+class HyperHTTPRequest(HTTPRequest):
+    """An HTTP Request (and response).
+
+    A single HTTP connection may consist of multiple request/response pairs.
+    """
+
+    server = None
```
webknjaz: Do these attrs really need to be copied over from that base class?

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+
+    header_reader = HeaderReader()
+    """
+    A HeaderReader instance or compatible reader.
+    """
+
+    def __init__(self, server, conn, proxy_mode=False, strict_mode=True):
+        """Initialize HTTP request container instance.
+
+        Args:
```
webknjaz: Please use native Sphinx params syntax. I've already started converting some places that use this one: https://github.com/cherrypy/cheroot/blob/ee55bb1/cheroot/server.py#L1420-L1423

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+        self.close_connection = self.__class__.close_connection
+        self.chunked_read = False
+        self.chunked_write = self.__class__.chunked_write
+        self.proxy_mode = proxy_mode
+        self.strict_mode = strict_mode
+
+    def _next_event(self):
+        while True:
+            event = self.h_conn.next_event()
+            if event is h11.NEED_DATA:
```
webknjaz: This could be simplified as
```suggestion
            if event is not h11.NEED_DATA:
                return event
```
This would allow to dedent the whole following block and drop `continue`

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+        self.outheaders = []
+        self.sent_headers = False
+        self.close_connection = self.__class__.close_connection
+        self.chunked_read = False
+        self.chunked_write = self.__class__.chunked_write
+        self.proxy_mode = proxy_mode
+        self.strict_mode = strict_mode
+
+    def _next_event(self):
+        while True:
```
webknjaz: Alternatively, this could be rewritten as
```suggestion
        event = self.h_conn.next_event()
        while event is not h11.NEED_DATA:
            if self.h_conn.they_are_waiting_for_100_continue:
                go_ahead = h11.InformationalResponse(status_code=100)
                bytes_out = self.h_conn.send(go_ahead)
                self.conn.wfile.write(bytes_out)

            self.h_conn.receive_data(self.conn.rfile.readline())
            event = self.h_conn.next_event()

        return event
```

does this look cleaner?

cheroot/server.py
duper51: We have some tests that I believe check for HTTP/1.1 in the response which is why I left it that way. In my testing at home I have done this to some success!

cheroot/server.py
duper51: Sorry about that, the IDE I'm using likes to be a bit overzealous and auto-format entire files, I'll pull out these changes in an upcoming commit.

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+                continue
+            return event
+
+    def parse_request(self):
+        """Parse the next HTTP request start-line and message-headers."""
+        req_line = self._next_event()
+        if isinstance(req_line, h11.Request):
+            self.started_request = True
+            self.uri = req_line.target
+            scheme, netloc, path, query, fragment = urllib.parse.urlsplit(self.uri)
```
webknjaz: Make sure to test this against Python 3.9. I recall they fixed RFC compat which isn't exactly backward-compatible. `'host:port'` now puts `host` into the scheme part, not host as before.

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+            self.path = path
+            self.qs = query
+            self.method = req_line.method
+            self.request_protocol = b"HTTP/%s" % req_line.http_version
```
webknjaz: I'm pretty sure that %-formatting doesn't work well with byte strings... Or does it?

cheroot/server.py
webknjaz: I think you could save a few lines by doing the assignments directly
```suggestion
            _scheme, _netloc, self.path, self.qs, _fragment = urllib.parse.urlsplit(self.uri)
```

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+        self.status = ''
+        self.outheaders = []
+        self.sent_headers = False
+        self.close_connection = self.__class__.close_connection
+        self.chunked_read = False
+        self.chunked_write = self.__class__.chunked_write
+        self.proxy_mode = proxy_mode
+        self.strict_mode = strict_mode
+
+    def _next_event(self):
```
webknjaz: This name doesn't add value to the semantics of its purpose. Method names should start with verbs because methods/functions ultimately represent actions.
Also, it seems to be called once, "next" in the name kinda suggests that it's supposed to be called multiple times which is misleading.

Let's think of something better here. What does it do? To me, it looks like it `await`s for some event to happen and also communicates some data/state to the other end.

So maybe something like `_communicate_100`? But it's probably too much implementation details also it's not the primary purpose of this method, is it?

How about
```suggestion
    def _await_request_line_bytes(self):
```
?

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+    """
+
+    server = None
+    """The HTTPServer object which is receiving this request."""
+
+    conn = None
+    """The HTTPConnection object on which this request connected."""
+
+    inheaders = {}
+    """A dict of request headers."""
```
webknjaz: Should this be a dict, though? I seem to recall that there could be header duplicates...

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+            self.started_request = True
+            self.uri = req_line.target
+            scheme, netloc, path, query, fragment = urllib.parse.urlsplit(self.uri)
+            self.path = path
+            self.qs = query
+            self.method = req_line.method
+            self.request_protocol = b"HTTP/%s" % req_line.http_version
+            self.response_protocol = b"HTTP/%s" % req_line.http_version
+            # TODO: oneliner-ify this
+            self.inheaders = {}
```
webknjaz: Does this work?
```suggestion
            self.inheaders = dict(req_line.headers)
```

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+            # TODO
+            raise NotImplementedError("Only expecting Request object here")
+
+    def respond(self):
+        """Call the gateway and write its iterable output."""
+        mrbs = self.server.max_request_body_size
+        if not self.chunked_read:
+            cl = int(self.inheaders.get(b'Content-Length', 0))
+            if mrbs and mrbs < cl:
+                if not self.sent_headers:
```
webknjaz: How about reversing this?
```suggestion
                if self.sent_headers:
                    return
```

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+                        'The entity sent with the request exceeds the '
+                        'maximum allowed bytes.',
+                    )
+                return
+        self.rfile = HyperRFile(self.conn, self.h_conn)
+
+        self.server.gateway(self).respond()
+        self.ready and self.ensure_headers_sent()
+
+    def simple_response(self, status, msg=''):
```
webknjaz: Usually, it's bad to use mutable defaults. Use `None` and assign the default in the beginning of the method
```suggestion
    def simple_response(self, status, msg=None):
```

cheroot/server.py
```diff
@@ -1204,6 +1254,190 @@ def send_headers(self):
...
+            ))
+
+        if b'server' not in hkeys:
+            self.outheaders.append((
+                b'server',
+                self.server.server_name.encode('ISO-8859-1'),
+            ))
+
+        res = h11.Response(status_code=status, headers=self.outheaders,
+                           http_version=self.response_protocol, reason=self.status[3:])
```
webknjaz: Maybe?
```suggestion
                           http_version=self.response_protocol, reason=self.status[4:])
```

cheroot/server.py
webknjaz: Thanks. I just wanted to make some early review and ended up checking each diff chunk for whether it's some formatting change or is it actually useful...

cheroot/server.py
```diff
@@ -614,10 +668,6 @@ def read_trailer_lines(self):
 
             yield line
 
-    def close(self):
-        """Release resources allocated for ``rfile``."""
-        self.rfile.close()
-
 
 class HTTPRequest:
```
webknjaz: Are you keeping this temporary while refactoring?

duper51:
So I've been peeking over the code reviews and I agree with most, if not all of them. I do have a few questions in regards to a few of them however.

There are a few instances where I'm attempting to take things from h11 (a lot of byte-strings), and shim them into the format that the old `HTTPRequest` class used, which increases the complexity and I believe a few of the errors I've made throughout. 

@webknjaz Would you like me to attempt to make this change backwards compatible with the existing class, or throw that out in favor of increasing interoperability with h11? Note that some of the expectations (such as sending an HTTP 1.0 request and expecting an HTTP 1.0 response) are already going to be violated.

webknjaz:
@duper51 I think I'm okay with not keeping compat. But just make sure that it doesn't affect CherryPy (it shouldn't but still).

webknjaz:
FTR this document basically confirms Nathaniel's advice to always respond with HTTP/1.1: https://www.rfc-editor.org/rfc/rfc2145.html#section-2.3

webknjaz:
@duper51 master got updated, time to rebase.

cheroot/test/test_core.py
```diff
@@ -333,6 +333,7 @@ def test_malformed_request_line(
     c.close()
 
 
+@pytest.mark.xfail(reason="h11 normalizes the method and ignores non-capitalized methods")
```
webknjaz: Contribution to H11 needed?

cheroot/test/test_conn.py
```diff
@@ -342,6 +345,9 @@ def test_streaming_10(test_client, set_cl):
         'HTTP/1.1',
     ),
 )
+
+
+@pytest.mark.xfail(reason="h11 does not support HTTP/1.0 keepalive")
```
webknjaz: What can we do about it?

webknjaz:
@duper51 any help needed?

webknjaz:
recheck

duper51:
Sorry about that, we're getting about to midterms. I have some changes that I am making, mostly focused on getting tests to work correctly, or disabling tests that no longer apply under the new parser. I was curious if you have any input on how I am going about integrating this or any thoughts on making it more concise/clear?

webknjaz:
It's alright. No new input so far except that maybe we need to contribute some things to h11 instead of trying to shim things on Cheroot side.

P.S. That "recheck" comment was to trigger the new Zuul CI integration I'm playing with on this PR.

webknjaz:
Looks like we need to add https://h11.readthedocs.io/ to intersphinx

duper51:
> The `HTTP/1.0 Connection: keep-alive` pseudo-standard is currently not supported. (Note that this  only affects h11 as a server, because h11 as a client always speaks HTTP/1.1.) Supporting this would be possible, but it's fragile and finicky and I'm suspicious that if we leave it out then no-one will notice or care. HTTP/1.1 is now almost old enough to vote in United States elections. I get that people sometimes write HTTP/1.0 clients because they don't want to deal with annoying stuff like chunked encoding, and I completely sympathize with that, but I'm guessing that you're not going to find too many people these days who care desperately about keep-alive and at the same time are too lazy to implement Transfer-Encoding: chunked. Still, this would be my bet as to the missing feature that people are most likely to eventually complain about...

As an aside, here is an excerpt from https://github.com/python-hyper/h11/blob/master/docs/source/supported-http.rst. It looks like the maintainer is not opposed to having the support as a feature, but isn't super interested in implementing it himself. We could potentially try and implement that on their end, but it would likely be a fairly large overhaul (supporting the HTTP/1.0 server side protocol would be required).

webknjaz:
Okay, let's not care about that then.

njsmith:
The actual implementation of HTTP/1.0 keepalive would probably be pretty straightforward. The tricky part is that there's no spec, and the desired semantics are entirely based on folklore, so it's very hard to know whether you did it right or not.

duper51:
Were we to implement it, implement it similar to how Cheroot currently implements it and port the tests over. Would you be okay with something along those lines?

webknjaz:
@duper51 that sounds good to me. Given that CherryPy's been around for about 18 years and nobody complained, its implementation is as good as any...

duper51:
These are the latest commits that I have. Currently, I am working on some fixes for some conditions that arise in the tests, but overall it seems to be working barring edge cases.

There is still no CONNECT support, but I will be looking into that shortly after completing the test_conn tests.

cheroot/server.py
```diff
@@ -73,6 +73,7 @@
 import traceback as traceback_
 import logging
 import platform
+import h11
```
webknjaz: move this import before `six` as it's third-party

cheroot/server.py
```diff
@@ -341,16 +342,92 @@ def __next__(self):
...
+
+class KnownLengthRFile(RFile):
     """Wraps a file-like object, returning an empty string when exhausted.
 
     :param rfile: ``file`` of a known size
     :param int content_length: length of the file being read
     """
 
-    def __init__(self, rfile, content_length):
+    def __init__(self, h_conn, rfile, wfile, content_length):
```
webknjaz: this signature change seems to be breaking calls in other parts of the code base

duper51: As far as I can tell, the only usages in the project besides the ones that I have modified to work with it is the ones in the old HTTPRequest implementation that I intend to remove when this PR is closer to production-ready.

webknjaz: Check https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-61d0007190a3d7e76102eb80c05d5b57af1e4547. Did you take inheritance into account?

duper51: Oops, left that in as well. The HyperRFile has been removed in favor of modifying the other existing implementations. I have just updated the repository to remove it :)

pytest.ini
```diff
@@ -1,14 +1,14 @@
 [pytest]
 addopts =
     # `pytest-xdist`:
-    --numprocesses=auto
+    # --numprocesses=auto
```
webknjaz: You could override this all in `tox.ini` or use `tox -- -n 0 --testmon-off --durations 0` and so on

duper51: I keep forgetting to remove that from the changelist.

I have to disable those for my debugging workflow. I can't seem to get tox to operate correctly when working with the PyCharm debugger. I was, however, able to get pytest to be stable when modifying the pytest configuration in this manner.

I'm not sure of a better way to go about it, any suggestions?

webknjaz: Oh, in this case, it's probably fine. I haven't used PyCharm in ages — have you tried their remote debugger mode?

duper51: No, but I believe the effect would be the same. Whatever debugger PyCharm uses internally disallows the usage of one of the packages loaded by `testmon`, and `xdist` causes issues where breakpoints may or may not trigger. There also appears to be an issue where if `testmon` is left loaded it sometimes fails to detect any tests at all.

I'll make sure to revert this change and update :)

webknjaz: Right, testmon is a _very_ tricky tool intended to be used mostly during the development.

cheroot/server.py
```diff
@@ -341,16 +342,53 @@ def __next__(self):
...
+        self.rfile = rfile
+        self.wfile = wfile
+        self.conn = conn
+
+    def _send_100_if_needed(self):
+        if self.conn.they_are_waiting_for_100_continue:
+            mini_headers = ()
+            go_ahead = h11.InformationalResponse(status_code=100, headers=mini_headers)
+            bytes_out = self.conn.send(go_ahead)
+            self.wfile.write(bytes_out)
```
webknjaz: let's use a guard check style here
```suggestion
        if not self.conn.they_are_waiting_for_100_continue:
            return
        mini_headers = ()
        go_ahead = h11.InformationalResponse(status_code=100, headers=mini_headers)
        bytes_out = self.conn.send(go_ahead)
        self.wfile.write(bytes_out)
```

cheroot/server.py
```diff
@@ -341,16 +342,53 @@ def __next__(self):
...
+            return b""
+
+    def read(self, size=None):
+        raise NotImplementedError
+
+    def readline(self, size=None):
+        raise NotImplementedError
+
+    def readlines(self, sizehint=None):
+        raise NotImplementedError
```
webknjaz: https://docs.python.org/3/library/abc.html#abc.abstractmethod

cheroot/server.py
```diff
@@ -1204,6 +1247,289 @@ def send_headers(self):
...
+
+        if self.h_conn.their_state is h11.MUST_CLOSE:
+            self.close_connection = True
+
+        # prep for next req cycle if it's available
+        if self.h_conn.our_state == h11.DONE and self.h_conn.their_state == h11.DONE:
+            self.h_conn.start_next_cycle()
+
+    def simple_response(self, status, msg=''):
+        """Write a simple response back to the client."""
```
```suggestion
    def simple_response(self, status, msg=None):
        """Write a simple response back to the client."""
        if msg is None:
            msg = ''
```

cheroot/server.py
```diff
@@ -1204,6 +1247,289 @@ def send_headers(self):
...
+
+        while self.h_conn.their_state is h11.SEND_BODY:
+            # empty their buffer ?
+            self._next_event()
+
+        if self.h_conn.their_state is h11.MUST_CLOSE:
+            self.close_connection = True
+
+        # prep for next req cycle if it's available
+        if self.h_conn.our_state == h11.DONE and self.h_conn.their_state == h11.DONE:
```
webknjaz: For consistency, use the same style of checking for these constants as earlier in code.
```suggestion
        if self.h_conn.our_state is h11.DONE and self.h_conn.their_state is h11.DONE:
```

cheroot/server.py
```diff
@@ -1298,26 +1304,101 @@ def parse_request(self):
...
                 for header in req_line.headers:
                     self.inheaders[header[0]] = header[1]
 
-                if ('transfer-encoding' in self.inheaders and
-                    self.inheaders['transfer-encoding'].lower() == b"chunked"):
+                if (b'transfer-encoding' in self.inheaders and
+                    self.inheaders[b'transfer-encoding'].lower() == b"chunked"):
                     self.chunked_read = True
 
+                uri_is_absolute_form = (scheme or authority)
```
```suggestion
                uri_is_absolute_form = scheme or authority
```

cheroot/server.py
```diff
@@ -1204,6 +1247,289 @@ def send_headers(self):
...
+            # NEED INFO: Should we adjust the tests to match h11's exception text
+            # Or should we continue to shim like this:
+            if e.args[0] == "bad Content-Length":
+                self.simple_response(e.error_status_hint or 400, "Malformed Content-Length Header.")
+            elif e.args[0] == "illegal request line":
+                self.simple_response(e.error_status_hint or 400, 'Malformed Request-Line')
+            elif e.args[0] == "illegal header line":
+                self.simple_response(e.error_status_hint or 400, 'Illegal header line.')
+            else:
+                self.simple_response(e.error_status_hint or 400, str(e))
```
webknjaz: how about
```suggestion
			err_map = {
			    'bad Content-Length': 'Malformed Content-Length Header.',
			    'illegal request line': 'Malformed Request-Line.',
			    'illegal header line': 'Illegal header line.',
			}
			err_msg = err_map.get(e.args[0], str(e))
			self.simple_response(e.error_status_hint or 400, err_msg)
```

cheroot/server.py
duper51: Close is provided for in the parent class `RFile`

cheroot/server.py
duper51: I have renamed it to `_process_next_h11_event`. This seems to be the most accurate; depending on the execution, it may read data into h11's internal buffer, then attempt to get an event from h11. 

There is another function that calls this method that does not read and simply asks h11 if any new events have been queued since the last time an event was produced.

cheroot/test/test_core.py
duper51: This appears to be by design? @njsmith

njsmith: h11 seems to be happy to work with lowercase methods here, so I think the diagnosis is wrong?

```
In [2]: import h11                                                                        

In [3]: c = h11.Connection(our_role=h11.CLIENT)                                           

In [5]: h11.Request(method="get", headers=[("host", "example.com")], target="/")          
Out[5]: Request(method=b'get', target=b'/', headers=[(b'host', b'example.com')], http_version=b'1.1')

In [6]: c.send(_5)                                                                        
Out[6]: b'get / HTTP/1.1\r\nhost: example.com\r\n\r\n'

In [7]: s = h11.Connection(our_role=h11.SERVER)                                           

In [8]: s.receive_data(_6)                                                                

In [9]: s.next_event()                                                                    
Out[9]: Request(method=b'get', target=b'/', headers=[(b'host', b'example.com')], http_version=b'1.1')
```

cheroot/server.py
duper51: Anything other than `h11.Data` here should be undefined behavior. Currently, I believe the only possible way we don't is if we get an EndOfMessage, but I'm pretty sure that were that to happen, h11 would raise an exception. I believe I saw this in the example code that they provide, so I kept it in for fun. 

I can't really think of any other way it would be anything else, so I kept it in as a general catch-all.

webknjaz: Alright. This is a practice people follow when doing contract programming just as a sanity-check. `assert`s are stripped off in optimized mode. But it's useful to keep this as an exit-check.

cheroot/server.py
duper51: It might be worth looking in to: https://mail.python.org/pipermail/python-dev/2014-March/133621.html

It appears that it is stable after Python 3.5?

duper51:
So I noticed while performing a code review on this, the GitHub diff viewer appears to be very confused when displaying the diff between the two files. That's gonna make things somewhat more difficult to view what's changed. Is there any way to resolve that somehow?

Also, on my local machine, this appears to pass all tests, except two.

There are two tests that I'm a little worried about. One is the `tls_client_auth` test, which seems to have erratic behavior, but as far as I can tell, the stack traces produced don't seem to possess any of the cheroot server code.

The other is HTTPS over HTTP. I believe this fails for a similar reason to the bad newline test. I'm not super sure on how to resolve this one.

cheroot/server.py
```diff
@@ -61,6 +61,7 @@
 """
 
 from __future__ import absolute_import, division, print_function
+
```
webknjaz, line 64: This change is unnecessary and shouldn't be done.
```suggestion
```

cheroot/server.py
```diff
@@ -73,35 +74,33 @@
 import traceback as traceback_
 import logging
 import platform
+from abc import ABC, abstractmethod
```
webknjaz, line 77: There's no `ABC` class in Python 2 (that we still haven't dropped), it's new in Python 3.4. Also, the canonical way is to use `ABCMeta`.

We'll be able to revise this decision once Python 2 support is dropped.
```suggestion
from abc import ABCMeta, abstractmethod
```

cheroot/server.py
```diff
@@ -168,189 +161,62 @@
...
-
-
-class DropUnderscoreHeaderReader(HeaderReader):
-    """Custom HeaderReader to exclude any headers with underscores in them."""
-
-    def _allow_header(self, key_name):
-        orig = super(DropUnderscoreHeaderReader, self)._allow_header(key_name)
-        return orig and '_' not in key_name
-
+class RFile(ABC):
```
webknjaz, line 168:
```suggestion
@six.add_metaclass(ABCMeta)
class RFile:
```

cheroot/server.py
```diff
@@ -168,189 +161,62 @@
...
-        :param int size: amount of data to read
-
-        :returns: chunk from ``rfile``, limited by size if specified
-        :rtype: bytes
-        """
-        data = self.rfile.read(size)
-        self.bytes_read += len(data)
-        self._check_length()
-        return data
+        pass
```
webknjaz, line 196: Keep the docstring instead of replacing it with `pass`.

cheroot/server.py
```diff
@@ -690,374 +570,177 @@ def __init__(self, server, conn, proxy_mode=False, strict_mode=True):
...
+            else:
+                # TODO
+                raise NotImplementedError('Only expecting Request object here')
+        except h11.RemoteProtocolError as e:
+            err_map = {
+                'bad Content-Length': 'Malformed Content-Length Header.',
+                'illegal request line': 'Malformed Request-Line.',
+                'illegal header line': 'Illegal header line.',
+            }
+            err_str = [v for k, v in err_map.items() if e.args[0] in k]
```
webknjaz, line 730: Why is it `in` and not `==`?

cheroot/server.py
```diff
@@ -1066,50 +749,72 @@ def respond(self):
...
+            go_ahead = h11.InformationalResponse(
+                status_code=100,
+                headers=mini_headers,
+            )
+            bytes_out = self.h_conn.send(go_ahead)
+            self.conn.wfile.write(bytes_out)
+        try:
+            self.server.gateway(self).respond()
+            self.ready and self.ensure_headers_sent()
+        except errors.MaxSizeExceeded as e:
```
webknjaz, line 768: Normally, you need to have only one statement in the try block. Otherwise, it may be harmful when the exception starts happening in a different place.

webknjaz:
@duper51

> So I noticed while performing a code review on this, the GitHub diff viewer appears to be very confused when displaying the diff between the two files.

Do you mean between two different files? Like when you move code chunks? There's no way of fixing this. Local diff may be a bit helpful but usually, I can live with it being like this.

Also, the diff will be smaller once you revert all those unnecessary style changes.


> appears to pass all tests, except two.

Also, linters are failing (https://github.com/cherrypy/cheroot/pull/262/checks?check_run_id=523525295#step:9:31). And Python 2.7 env is failing because you import `ABC` (https://github.com/cherrypy/cheroot/pull/262/checks?check_run_id=523525384#step:14:59).


So let's fix all of these issues to proceed with a deeper review.

cheroot/test/test_core.py
```diff
@@ -306,14 +307,14 @@ def test_large_request(test_client_with_defaults):
...
+            HTTP_BAD_REQUEST, b'Malformed Request-Line',
         ),
         (
             b'GET / HTTP/1',  # invalid version
-            HTTP_BAD_REQUEST, b'Malformed Request-Line: bad version',
+            HTTP_BAD_REQUEST, b'Malformed Request-Line',
         ),
         (
-            b'GET / HTTP/2.15',  # invalid ver
+            b'GET / HTTP/2.1',  # invalid ver
```
webknjaz, line 317: This change is controversial. Why the RFC's spec is that it has one digit after the dot, the next paragraph gives an example of the two-digit version...

cheroot/server.py
```diff
@@ -1553,9 +1206,9 @@ class HTTPServer:
     Default is 10. Set to None to have unlimited connections."""
 
     def __init__(
-        self, bind_addr, gateway,
-        minthreads=10, maxthreads=-1, server_name=None,
-        peercreds_enabled=False, peercreds_resolve_enabled=False,
+            self, bind_addr, gateway,
+            minthreads=1, maxthreads=1, server_name=None,
```
webknjaz, line 1210: The signature change doesn't look related to this PR? (As well as indentation)

webknjaz:
@duper51 there's conflicts with master in this branch. Mind rebasing?

cheroot/server.py
webknjaz: What about Python 2.7?

webknjaz:
@ian-otto any updates here?

ian-otto:
Wrapping up a move, then I'll rebase and get this ready to go.

cheroot/server.py
ian-otto: It absolutely isn't. I was having problems with the debugger getting stuck due to threading issues and had to tune the settings. I'll pull this out right away

cheroot/server.py
```diff
@@ -75,35 +75,34 @@
 import platform
 import contextlib
 import threading
+import h11
```
webknjaz, line 78: You added a copy of this import below but forgot to also remove this line
```suggestion
```

setup.cfg
```diff
@@ -69,6 +69,7 @@ install_requires =
     six>=1.11.0
     more_itertools>=2.6
     jaraco.functools
+    h11
```
webknjaz, line 72:
```suggestion
    h11 > 0.9.0
```

cheroot/test/test_core.py
webknjaz: @ian-otto ^

cheroot/test/test_core.py
```diff
@@ -195,7 +195,8 @@ def test_parse_uri_invalid_uri(test_client):
     response = _get_http_response(c, method='GET')
     response.begin()
     assert response.status == HTTP_BAD_REQUEST
-    assert response.read(21) == b'Malformed Request-URI'
+    # TODO: h11 doesn't have this specific of messaging. See python-hyper/h11#98
+    # assert response.read(21) == b'Malformed Request-URI'
```
webknjaz, line 199: @ian-otto I think this can be corrected now

ian-otto: Yes and no. My PR over in that repo only handled updating the messaging for a few specific cases that I felt were very important to be distinct. That said, h11 still gives fairly generic errors when it comes to request line problems. Part of me thinks that in that case, it may be okay to let that happen, since request-line issues are generally pretty easy to spot. I'm curious on your opinion though.

ian-otto:
As an aside, one of the tests seems to intermittently hang for seemingly days depending on CI platform. Can we set a lower timeout?

cheroot/test/test_core.py
webknjaz: I think that we could leave it as-is for now and maybe add that in the future. OTOH it'd nice UX/DX-wise to have precise messages in general.

webknjaz:
@ian-otto what CI job are you talking about specifically? There's some known issues that happen flakily and are unrelated to your changes here...

ian-otto:
Back when it was still able to execute tests (running from h11's Git), I had a test timeout after 6 hours on GitHub, and another fail after 12 hours. Here are the related workflow runs:

https://github.com/ian-otto/cheroot/actions/runs/189982387
https://github.com/ian-otto/cheroot/actions/runs/189218067

webknjaz:
@ian-otto there's no 12-hour jobs among your links. Looks like the workflows were stuck because of GitHub's outage. It should be ok now. Also, a few TLS tests are extremely flaky (mostly in slow envs like macOS or Windows) so ignore them.

ian-otto:
I see that you've made a few commits to this PR, but I can't pull them in. I'm not particularly sure how this works, is there a way to fetch them?

webknjaz:
@ian-otto try `git pull --rebase`, this should help if you have extra commits that haven't been pushed to GitHub and now the history diverges. This may produce some conflicts for you to resolve, though. If you have more problems with that, just tell me and I'll try to help.

Also, this branch is a bit out of date and would need to get the updates from master. Normally, I'd suggest rebasing but seeing that your branch already contains some merge commits, I'd say just merge-in master from upstream (because rebasing with merge commits is PITA most of the time).

webknjaz:
@ian-otto hey, do you need me to help you out with Git?

tox.ini
```diff
@@ -9,6 +9,8 @@ deps =
     # Possible workaround for errors in Travis CI under some Pythons
     pip>=19.1
     wheel
+    # FIXME: Drop this temporary unreleased dependency before merging the PR:
+    h11 @ git+https://github.com/python-hyper/h11.git
```
webknjaz, line 13:
```suggestion
```

setup.cfg
```diff
@@ -69,6 +69,7 @@ install_requires =
     six>=1.11.0
     more_itertools>=2.6
     jaraco.functools
+    h11 > 0.9.0
```
webknjaz, line 72:
```suggestion
    h11 >= 0.10.0
```

cheroot/test/test_core.py
```diff
@@ -341,6 +342,7 @@ def test_malformed_request_line(
     c.close()
 
 
+@pytest.mark.xfail(reason='h11 normalizes the method and ignores non-capitalized methods')
```
webknjaz, line 345: @ian-otto looks like v0.11.0 fixes this

jaraco:
Closing as abandoned. Feel free to revive (original poster or others).

PR #277, opened 2020-04-09
Be more graceful when terminating keep-alive connections

### LINKED ISSUE #263: [cheroot==8.1.0 regression] Occasional connection resets with concurrent requests
**I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] question about the decisions made in the repository

**Do you want to request a *feature* or report a *bug*?**

Reporting a bug

**What is the current behavior?**

When running concurrent requests against a cherrypy the client occasionally gets 'Connection reset by peer' errors.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a screenshots and logs of the problem. If you can, show us your code.**

This can be reproduced by running the following server and client code. Tested both on MacOS with python 3.7 and Ubuntu bionic on python 3.6. Both reveal the same connection errors:

```console
ERROR:test:FAILED http://localhost:5000/info: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))
(...)
ERROR:test:FAILED http://localhost:5000/info: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))
ERROR:test:FAILED http://localhost:5000/info: ('Connection aborted.', ConnectionResetError(54, 'Connection reset by peer'))
INFO:test:921 out of 1000 successful
```

When reproducing locally the issue reproduces already with 20 workers occasionally and with 50 workers on every run of the client script. I also played with various values of threadpool and queue size settings but couldn't get it working.

Server code:

```python
import logging
import cherrypy

logging.basicConfig(level=logging.DEBUG)


class TestAPI:

    @cherrypy.expose
    def info(self):
        return 'test\n'


def main():
    route_map = cherrypy.dispatch.RoutesDispatcher()
    api = TestAPI()
    route_map.connect('info', '/info', controller=api, action='info')

    conf = {
        '/': {
            'request.dispatch': route_map
        }
    }

    cherrypy.tree.mount(api, '/', config=conf)

    cherrypy.config.update({
        'global': {
            'environment': 'production',
            'server.socket_host': '0.0.0.0',
            'server.socket_port': 5000,
            # 'server.thread_pool': 20,
            # 'server.thread_pool_max': 20,
            # 'server.socket_queue_size': 128,
            # 'server.accepted_queue_size': 100000,
        },
    })

    cherrypy.engine.start()
    cherrypy.engine.block()


if __name__ == '__main__':
    main()
```

Client code:

```python
from concurrent.futures.thread import ThreadPoolExecutor
from urllib.parse import urljoin

import requests
import logging

baseurl = 'http://localhost:5000/'

log = logging.getLogger('test')


def configure_logging(context):
    logger = logging.getLogger(context)
    logger.setLevel(logging.DEBUG)


def request_info(session):
    url = urljoin(baseurl, f'info')
    try:
        r = session.get(url)
    except requests.exceptions.ConnectionError as e:
        log.error('FAILED %s: %s', url, e)
        return False
    r.raise_for_status()
    return True


def main():
    session = requests.session()
    a = requests.adapters.HTTPAdapter(pool_connections=1, pool_maxsize=1000)
    session.mount('http://', a)

    count = 1000
    success = 0
    with ThreadPoolExecutor(max_workers=50) as tp:
        tasks = []
        for i in range(count):
            tasks.append(tp.submit(request_info, session))

        for task in tasks:
            if task.result():
                success += 1

    log.info('%s out of %s successful', success, count)


if __name__ == "__main__":
    logging.basicConfig()
    configure_logging('test')
    main()
```

We see the same behavior when running a production service between apache as a reverse proxy.

**What is the expected behavior?**

There are no 'Connection reset by peer' errors.

**What is the motivation / use case for changing the behavior?**

**Please tell us about your environment:**

- Cheroot version: 8.2.1
- CherryPy version: 18.5.0
- Python version: 3.6.9 (Ubuntu bionic) and python 3.7 (MacOS)
- OS: MacOS and Ubuntu bionic
- Browser: requests library and apache reverse proxy

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, e.g. stackoverflow, gitter, etc.)

Note that when running with Tornado as HTTP server the problem doesn't reproduce:

```python
import logging
import cherrypy

logging.basicConfig(level=logging.DEBUG)


class TestAPI:

    @cherrypy.expose
    def info(self):
        return 'test\n'


def main():
    import tornado
    import tornado.httpserver
    import tornado.wsgi

    route_map = cherrypy.dispatch.RoutesDispatcher()
    api = TestAPI()
    route_map.connect('info', '/info', controller=api, action='info')

    conf = {
        '/': {
            'request.dispatch': route_map
        }
    }

    wsgiapp = cherrypy.tree.mount(api, '/', config=conf)

    # Disable the autoreload which won't play well
    cherrypy.config.update({'engine.autoreload.on': False})

    # let's not start the CherryPy HTTP server
    cherrypy.server.unsubscribe()

    # use CherryPy's signal handling
    cherrypy.engine.signals.subscribe()

    # Prevent CherryPy logs to be propagated
    # to the Tornado logger
    cherrypy.log.error_log.propagate = False

    # Run the engine but don't block on it
    cherrypy.engine.start()

    # Run thr tornado stack
    container = tornado.wsgi.WSGIContainer(wsgiapp)
    http_server = tornado.httpserver.HTTPServer(container)
    http_server.listen(5000)

    # Publish to the CherryPy engine as if
    # we were using its mainloop
    tornado.ioloop.PeriodicCallback(
        lambda: cherrypy.engine.publish('main'), 100).start()
    tornado.ioloop.IOLoop.instance().start()


if __name__ == '__main__':
    main()

```

the-allanc:
  - [X] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Fixes #263 

Current mechanism for managing keep-alive connections will forcibly close the least recently-used connection to ensure that we keep to the threshold of the maximum number of allowed keep-alive connections.

Examine if we have reached the threshold of keep-alive connections before writing response headers - if we have, then write the appropriate response headers to close the connection, rather than keeping the connection alive (even if the client has requested it).

📋 **Other information**:

📋 **Checklist**:

  - [-] Integration tests for the changes exist (if applicable)
  - [-] Documentation reflects the changes

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/277)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -91,12 +91,6 @@ class ConnectionManager:
         if conn.closeable:
             return
 
-        # Too many connections?
-        ka_limit = self.server.keep_alive_conn_limit
-        if ka_limit is not None and len(self.connections) > ka_limit:
-            conn.closeable = True
-            return
-
         # Connection too old?
         if (conn.last_used + self.server.timeout) < time.time():
             conn.closeable = True
@@ -277,3 +271,9 @@ class ConnectionManager:
         for conn in self.connections[:]:
             conn.close()
         self.connections = []
+
+    @property
+    def can_add_keepalive_connection(self):
+        ka_limit = self.server.keep_alive_conn_limit
+        stats = [ka_limit, len(self.connections)]
+        return ka_limit is None or len(self.connections) < ka_limit
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -274,5 +274,6 @@ class ConnectionManager:
 
     @property
     def can_add_keepalive_connection(self):
+        """Flag whether it is allowed to add a new keep-alive connection."""
         ka_limit = self.server.keep_alive_conn_limit
         return ka_limit is None or len(self.connections) < ka_limit
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -275,5 +275,4 @@ class ConnectionManager:
     @property
     def can_add_keepalive_connection(self):
         ka_limit = self.server.keep_alive_conn_limit
-        stats = [ka_limit, len(self.connections)]
         return ka_limit is None or len(self.connections) < ka_limit
--- a/cheroot/test/test_wsgi.py
+++ b/cheroot/test/test_wsgi.py
@@ -30,7 +30,6 @@ def simple_wsgi_server():
         yield locals()
 
 
-@pytest.mark.xfail(reason='#263')
 def test_connection_keepalive(simple_wsgi_server):
     """Test the connection keepalive works (duh)."""
     session = Session(base_url=simple_wsgi_server['url'])
--- a/cheroot/test/test_wsgi.py
+++ b/cheroot/test/test_wsgi.py
@@ -9,6 +9,10 @@ from requests_toolbelt.sessions import BaseUrlSession as Session
 from jaraco.context import ExceptionTrap
 
 from cheroot import wsgi
+from cheroot._compat import IS_MACOS, IS_WINDOWS
+
+
+IS_SLOW_ENV = IS_MACOS or IS_WINDOWS
 
 
 @pytest.fixture
@@ -24,7 +28,7 @@ def simple_wsgi_server():
 
     host = '::'
     addr = host, port
-    server = wsgi.Server(addr, app, timeout=20)
+    server = wsgi.Server(addr, app, timeout=600 if IS_SLOW_ENV else 20)
     url = 'http://localhost:{port}/'.format(**locals())
     with server._run_in_thread() as thread:
         yield locals()
@@ -44,10 +48,10 @@ def test_connection_keepalive(simple_wsgi_server):
             resp.raise_for_status()
         return bool(trap)
 
-    with ThreadPoolExecutor(max_workers=50) as pool:
+    with ThreadPoolExecutor(max_workers=10 if IS_SLOW_ENV else 50) as pool:
         tasks = [
             pool.submit(do_request)
-            for n in range(1000)
+            for n in range(250 if IS_SLOW_ENV else 1000)
         ]
         failures = sum(task.result() for task in tasks)
 
--- a/cheroot/test/test_wsgi.py
+++ b/cheroot/test/test_wsgi.py
@@ -24,7 +24,7 @@ def simple_wsgi_server():
 
     host = '::'
     addr = host, port
-    server = wsgi.Server(addr, app)
+    server = wsgi.Server(addr, app, timeout=20)
     url = 'http://localhost:{port}/'.format(**locals())
     with server._run_in_thread() as thread:
         yield locals()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+.. scm-version-title:: v8.3.1
+
+- Made terminating keep-alive connections more graceful
+  (:issue:`263` via :pr:`277`).
+
 .. scm-version-title:: v8.3.0
 
 - :cp-issue:`910` via :pr:`243`: Provide TLS-related
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1159,6 +1159,12 @@ class HTTPRequest:
                     # Closing the conn is the only way to determine len.
                     self.close_connection = True
 
+        # Override the decision to not close the connection if the connection
+        # manager doesn't have space for it.
+        if not self.close_connection:
+            can_keep = self.server.connections.can_add_keepalive_connection
+            self.close_connection = not can_keep
+
         if b'connection' not in hkeys:
             if self.response_protocol == 'HTTP/1.1':
                 # Both server and client are HTTP/1.1 or better
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -2024,7 +2024,7 @@ class HTTPServer:
                 # Just drop the conn. TODO: write 503 back?
                 conn.close()
 
-        #self.connections.expire()
+        self.connections.expire()
 
     @property
     def interrupt(self):
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -2024,7 +2024,7 @@ class HTTPServer:
                 # Just drop the conn. TODO: write 503 back?
                 conn.close()
 
-        self.connections.expire()
+        #self.connections.expire()
 
     @property
     def interrupt(self):
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -409,7 +409,7 @@ def test_keepalive_conn_management(test_client):
         http_connection.connect()
         return http_connection
 
-    def request(conn):
+    def request(conn, keepalive=True):
         status_line, actual_headers, actual_resp_body = test_client.get(
             '/page3', headers=[('Connection', 'Keep-Alive')],
             http_conn=conn, protocol='HTTP/1.0',
@@ -418,7 +418,10 @@ def test_keepalive_conn_management(test_client):
         assert actual_status == 200
         assert status_line[4:] == 'OK'
         assert actual_resp_body == pov.encode()
-        assert header_has_value('Connection', 'Keep-Alive', actual_headers)
+        if keepalive:
+            assert header_has_value('Connection', 'Keep-Alive', actual_headers)
+        else:
+            assert not header_exists('Connection', actual_headers)
 
     disconnect_errors = (
         http_client.BadStatusLine,
@@ -437,28 +440,21 @@ def test_keepalive_conn_management(test_client):
     # Reusing the first connection should still work.
     request(c1)
 
-    # Creating a new connection should still work.
+    # Creating a new connection should still work, but we should
+    # have run out of available connections to keep alive, so the
+    # server should tell us to close.
     c3 = connection()
-    request(c3)
+    request(c3, keepalive=False)
 
-    # Allow a tick.
-    time.sleep(0.2)
-
-    # That's three connections, we should expect the one used less recently
-    # to be expired.
+    # Show that the third connection was closed.
     with pytest.raises(disconnect_errors):
-        request(c2)
-
-    # But the oldest created one should still be valid.
-    # (As well as the newest one).
-    request(c1)
-    request(c3)
+        request(c3)
 
     # Wait for some of our timeout.
-    time.sleep(1.0)
+    time.sleep(1.2)
 
-    # Refresh the third connection.
-    request(c3)
+    # Refresh the second connection.
+    request(c2)
 
     # Wait for the remainder of our timeout, plus one tick.
     time.sleep(1.2)
@@ -467,9 +463,10 @@ def test_keepalive_conn_management(test_client):
     with pytest.raises(disconnect_errors):
         request(c1)
 
-    # But the third one should still be valid.
-    request(c3)
+    # But the second one should still be valid.
+    request(c2)
 
+    # Restore original timeout.
     test_client.server_instance.timeout = timeout
 
 


jaraco:
@webknjaz Can you work to clean up the linter error and get the unrelated failures to pass?

cheroot/connections.py
```diff
@@ -277,3 +271,9 @@ def close(self):
         for conn in self.connections[:]:
             conn.close()
         self.connections = []
+
+    @property
+    def can_add_keepalive_connection(self):
+        ka_limit = self.server.keep_alive_conn_limit
+        stats = [ka_limit, len(self.connections)]
```
webknjaz, line 278: Looks like this is left here accidentally.
```suggestion
```

webknjaz:
@jaraco sure, on it.

cheroot/connections.py
```diff
@@ -277,3 +271,8 @@ def close(self):
         for conn in self.connections[:]:
             conn.close()
         self.connections = []
+
+    @property
+    def can_add_keepalive_connection(self):
```
webknjaz, line 276:
```suggestion
    def can_add_keepalive_connection(self):
        """Flag whether it is allowed to add a new keep-alive connection."""
```

webknjaz:
@jaraco that test you wrote seems to fail: https://github.com/cherrypy/cheroot/pull/277/checks?check_run_id=576115030#step:14:352

webknjaz:
*/me also asked @tobiashenkel to check it this fixes his issue*

the-allanc:
> @jaraco that test you wrote seems to fail: https://github.com/cherrypy/cheroot/pull/277/checks?check_run_id=576115030#step:14:352

I wonder whether the test fails under certain circumstances, because some of the connections end up being expired by the server timeout setting. If you increased that (or maybe even disabled it altogether), then you could probably see if that's the case.

cheroot/test/test_wsgi.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def app(environ, start_response):
+        status = '200 OK'
+        response_headers = [('Content-type', 'text/plain')]
+        start_response(status, response_headers)
+        return [b'Hello world!']
+
+    host = '::'
+    addr = host, port
+    server = wsgi.Server(addr, app)
```
webknjaz, line 28:
```suggestion
    server = wsgi.Server(addr, app, timeout=0, accepted_queue_timeout=0)
```

@the-allanc like this?

the-allanc: `ConnectionManager.expire` looks at `server.timeout` to work out whether the socket should be closed - so I think that's the only setting that we should try changing (and limit the effect to just this one test).

I'd probably be more inclined to increase the default timeout to a high number, rather than zero (just because I'm not sure what that will do).

webknjaz: Interesting, changing this revealed some bug: https://github.com/cherrypy/cheroot/pull/277/checks?check_run_id=576581037#step:14:413. It now returns HTTP 500.

webknjaz: And it's now 465 failures, not 10 like before: https://travis-ci.com/github/cherrypy/cheroot/jobs/318105925#L441

cheroot/test/test_wsgi.py
```diff
@@ -0,0 +1,59 @@
+"""Test wsgi."""
+
+import threading
+from concurrent.futures.thread import ThreadPoolExecutor
+
+import pytest
+import portend
+import requests
+from requests_toolbelt.sessions import BaseUrlSession as Session
+from jaraco.context import ExceptionTrap
```
webknjaz, line 10: xfail py2 for now
```suggestion
import sys
import threading

import pytest
import portend
import requests
from requests_toolbelt.sessions import BaseUrlSession as Session
from jaraco.context import ExceptionTrap

try:
    from concurrent.futures.thread import ThreadPoolExecutor
except ImportError:
    pytestmark = pytest.mark.xfail(sys.version_info[0] == 2, reason='no concurrent.futures @ py2')
    if sys.version_info[0] != 2:
        raise
else:
    pytestmark = pytest.mark.xfail(sys.version_info[0] == 2, reason='no concurrent.futures @ py2')
```

cheroot/test/test_wsgi.py
```diff
@@ -0,0 +1,68 @@
...
+
+try:
+    from concurrent.futures.thread import ThreadPoolExecutor
+except ImportError:
+    pytestmark = pytest.mark.xfail(sys.version_info[0] == 2, reason='no concurrent.futures @ py2')
```
webknjaz, line 15:
```suggestion
    pytestmark = pytest.mark.xfail(
        sys.version_info[0] == 2,
        reason='no concurrent.futures @ py2',
    )
```

cheroot/test/test_wsgi.py
```diff
@@ -0,0 +1,68 @@
...
+    if sys.version_info[0] != 2:
+        raise
+else:
+    pytestmark = pytest.mark.xfail(sys.version_info[0] == 2, reason='no concurrent.futures @ py2')
```
webknjaz, line 19:
```suggestion
    pytestmark = pytest.mark.xfail(
        sys.version_info[0] == 2,
        reason='no concurrent.futures @ py2',
    )
```

cheroot/test/test_wsgi.py
webknjaz: > Interesting, changing this revealed some bug: [#277 (checks)](https://github.com/cherrypy/cheroot/pull/277/checks?check_run_id=576581037#step:14:413). It now returns HTTP 500.

Oh, so docs for `io.BufferedReader.read()` don't mention the possibility of returning `None` but https://bugs.python.org/issue35869 suggests that this may happen when in non-blocking mode (hence `timeout=0`)

jaraco: Ideally, it shouldn't be necessary to update this test as it passes on `cheroot < 8.1` (introduction of request-based workers). If the introduction of request-based workers requires the users to supply a non-default parameter where it was not required before, then we'll want to make that explicit in the changelog and give users (and downstream consumers like CherryPy) guidance on how best adapt to this new behavior and we should strongly consider making this parameter the default and adding cautions about failure modes when the current default values are used.

Remember, this test is the confirmed repro reported in the issue. If this test is changed, we'll have to revisit the issue and see what implications this altered test has on that evaluation.

the-allanc: _me waves at Jason_
> Ideally, it shouldn't be necessary to update this test as it passes on cheroot < 8.1

To clarify - the test passes for me locally, but under some setups, it seems to fail (mostly on MacOS machines, according to that test run). Are we sure that the test you added will work on the same machines using cheroot 8.1?

My expectation would be that the tests would be likely to fail in the same way - my theory is that the connections / sockets are being closed anyway, as it then exceeds the socket timeout.

jaraco: I'm not sure the test works reliably. I only tested it a few times, probably Python 3.8 and 2.7 on macOS.

jaraco: I'll try cherry-picking that test onto 8.0.x so we can establish a baseline for the test and validate it against the suite of environments in CI.

webknjaz: @jaraco I'm not sure you ran exactly the same test: this one uses `concurrent.futures` which is not a part of Python 2. Did you have some other version of it?

jaraco: I did not, and I suspect I did not run the test against Python 2. But now we have [evidence that the tests pass on Python 2.7-3.6](https://github.com/cherrypy/cheroot/runs/579811289).

setup.cfg
```diff
@@ -129,6 +129,10 @@ testing =
     # It's a transitive dependency of pytest-watch plugin.
     colorama!=0.4.2; python_version == "3.4"
 
+    portend
```
webknjaz, line 132: @jaraco it's weird that this appears in the PR after the merge

webknjaz:
@the-allanc tests seem to fail in "slow" envs and the timeout of 20 seconds doesn't seem to be helpful.

ssbarnea:
Any updates on this? Addressing it would be really awesome.

webknjaz:
@ssbarnea I think the main problem here is figuring out how to test the change. You may want to contribute to that by verifying if Zuul's case is solved.

ssbarnea:
@webknjaz Sure, I can run zuul tests with code from this change once it does pass its tests, mainly because this would be a manual testing process and I doubt it makes sense to test with broken tests. For the moment I made a small change in zuul to also limit both https://review.opendev.org/#/c/723855/

webknjaz:
@ssbarnea it may be that the fix is ok but the tests are not. And I have no idea how much time it'll take to figure out the tests here. But at least Zuul-side confirmation would be some sort of indication whether the fix is ok or not.

morucci:
Hi I'd like to help to land that change. I'm packaging Zuul for Fedora but the pinning of cheroot in Zuul requirements.txt due to #263 prevent to move forward.

I've run the reproducer proposed by Tobias on the last Fedora Rawhide (cherrypy-18.4.0 and cheroot-8.2.1) and I start to get some ConnectionResetError starting around 11/12 concurrent connexions.  (ThreadPoolExecutor(max_workers=(12)). 10 or less pass w/o errors.

I've tested the reproducer successfully with cheroot 8.0.0. With 8.1.0 and last cheroot master the errors appear.
With this patch on top of master the reproducer does not raise any error.

Let me know how could I help, I'll have a look to the tests.

webknjaz:
@morucci thanks for the feedback. To merge this we have to at least make the CIs pass (mostly GHA+Travis at least, beware that Circle CI is often for unrelated reasons and there's some other flaky jobs). To me, it looks like additional instability is introduced when the workers are slow (which is often the case for macOS).

morucci:
Hi @webknjaz, I did some more testing with this PR, and tried to play with the timeout and the way to close the socket but w/o success on MacOS. Maybe someone with a better knowledge of the code base could help, (I was able to deploy a macos and reproduce the issue quite easily using this howto [how-to-run-macos-on-kvm-qemu](https://computingforgeeks.com/how-to-run-macos-on-kvm-qemu/)).

The PR #287 I used to experiment pass the CI (a copy of this one + some other commits) mostly pass the CI: https://github.com/cherrypy/cheroot/pull/287/checks?check_run_id=738251920

- I added a xfail for test_connection_keepalive only on darwin
- I added a commit to close asap the connection when we reach the limit https://github.com/cherrypy/cheroot/pull/287/commits/2103191ed4b4b6cf86b76c6ab14711f04f9d0b42. I saw sometime the test_keepalive_conn_management test failed and it looks like this commit fixed it.

I've not included it in the PR but it seems adding a max_retries to the HTTPAdapter in test_connection_keepalive make the test pass on MacOS, but not sure that's an acceptable solution ?

Something I noticed in the code is that we set the socket timeout to 1 whatever the value configured with HTTPServer.timeout https://github.com/cherrypy/cheroot/blob/master/cheroot/server.py#L1765.

I'm not going to continue this investigation, I'm out of clues and ideas + I'm not familiar enough with the cheroot codebase.

But It is clear that this PR improve a lot the situation where cheroot fail to pass the test_connection_keepalive on all platform on master. The PR #287 shown it pass the CI GHA, with sometime (I re-triggered it multiple times) 1 connection error for test_connection_keepalive on a windows node.

I'll like to known what's your point of view to move forward and merge that PR even if it does not fix the issue completely and on all platform, but it improves the current master.

webknjaz:
@morucci thanks for the investigation! I'm not opposed to merging things that improve master stability. The only concern is that I want to be careful about it and watch out for any possible regressions.
Increasing retries on macOS is something that I'd like to consider since in my experience those workers are slow by default and often when we spawn one thing and wait for it to be up in some other component, it times out simply because we don't wait long enough.

Issue #276, opened 2020-04-08
[TODO] Fix `test_peercreds_unix_sock[abstract]` and `test_peercreds_unix_sock_with_lookup[abstract]` @ nightly


webknjaz:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**

CI is red for nightly Python.

❓ **What is the motivation / use case for changing the behavior?**

Passing CI

💡 **To Reproduce**

Run the tests under the latest Python dev release under Linux because abstract sockets are only supported there and these tests are skipped on other platforms.

💡 **Expected behavior**

Green CI.

📋 **Details**

```python-traceback
     def _validate_host(self, host):
        """Validate a host so it doesn't contain control characters."""
        # Prevent CVE-2019-18348.
        match = _contains_disallowed_url_pchar_re.search(host)
        if match:
>           raise InvalidURL(f"URL can't contain control characters. {host!r} "
                             f"(found at least {match.group()!r})")
E           http.client.InvalidURL: URL can't contain control characters. '\x00cheroot-test-socketd5260955-8000-467d-856a-0fc47d7d720e' (found at least '\x00')
```

📋 **Environment**

Travis CI python-nightly

📋 **Additional context**

https://travis-ci.com/github/cherrypy/cheroot/jobs/316761085#L635

webknjaz:
Solved by https://github.com/cherrypy/cheroot/commit/0eb77c82ebdb20fedc58682162fd76aa50a212b3. But may be refactored further by #280.

jaraco:
I think this issue should remain open. The symptom was addressed by monkey patching the standard library in tests. That means that users should be expected to monkey patch the stdlib to use this functionality. That seems like the wrong fix to me.

webknjaz:
> That means that users should be expected to monkey patch the stdlib to use this functionality.

Only if they use the low-level `HTTPConnection` I think. I guess, using custom things may be fine. But I see how this may be problematic and we probably have to add tests on the actual way of using this feature.

jaraco:
The issue stems from [bpo-38576](https://bugs.python.org/issue38576].

webknjaz:
[...]

PR #280, opened 2020-04-12
Remove `_patch_URL_validation` fixture assuming it's unneeded

jaraco:
In 45e1bbb07eb75d596cb6648df287ba0a8dc15010, webknjaz added a commit that added this branchy behavior on later Pythons. I did some refactoring on it, but now I'd like to explore why this functionality exists at all.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/280)

### DIFF
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -23,7 +23,6 @@ from ..testing import (
     ANY_INTERFACE_IPV4,
     ANY_INTERFACE_IPV6,
     EPHEMERAL_PORT,
-    get_server_client,
 )
 
 
@@ -171,19 +170,19 @@ class _TestGateway(Gateway):
 
 
 @pytest.fixture
-def peercreds_enabled_server_and_client(http_server, unix_sock_file):
+def peercreds_enabled_server(http_server, unix_sock_file):
     """Construct a test server with ``peercreds_enabled``."""
     httpserver = http_server.send(unix_sock_file)
     httpserver.gateway = _TestGateway
     httpserver.peercreds_enabled = True
-    return httpserver, get_server_client(httpserver)
+    return httpserver
 
 
 @unix_only_sock_test
 @non_macos_sock_test
-def test_peercreds_unix_sock(peercreds_enabled_server_and_client):
+def test_peercreds_unix_sock(peercreds_enabled_server):
     """Check that ``PEERCRED`` lookup works when enabled."""
-    httpserver, testclient = peercreds_enabled_server_and_client
+    httpserver = peercreds_enabled_server
     bind_addr = httpserver.bind_addr
 
     if isinstance(bind_addr, six.binary_type):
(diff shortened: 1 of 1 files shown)


webknjaz:
> but now I'd like to explore why this functionality exists at all.

"this functionality" as in binding to abstract UNIX sockets?

jaraco:
I just mean the monkeypatching that this PR removes.

webknjaz:
Oh, well it was long overdue but the fastest way to fix the test. If it's possible w/o patching, it'd be great.

jaraco:
Aha, so the [failure happens](https://travis-ci.com/github/cherrypy/cheroot/jobs/318631653) in test_peercreds_unix_sock, which is disabled on macOS, which explains why I couldn't replicate the issue locally.

jaraco:
Aha. It looks like you filed a ticket about this in #276. Since neither the code nor the commit linked to the bug, I was unaware of the issue.

webknjaz:
> Aha, so the [failure happens](https://travis-ci.com/github/cherrypy/cheroot/jobs/318631653) in test_peercreds_unix_sock, which is disabled on macOS, which explains why I couldn't replicate the issue locally.

Yep, abstract sockets is a concept that only exists in Linux kernel:

http://man7.org/linux/man-pages/man7/unix.7.html:
>        The abstract socket namespace is a nonportable Linux extension.

https://troydhanson.github.io/network/Unix_domain_sockets.html#_linux_abstract_socket_namespace:
> Linux has a special feature: if the pathname for a UNIX domain socket begins with a null byte `\0`, its name is not mapped into the filesystem. Thus it won’t collide with other names in the filesystem. Also, when a server closes its UNIX domain listening socket in the abstract namespace, its file is deleted; with regular UNIX domain sockets, the file persists after the server closes it.

Refs:
* https://utcc.utoronto.ca/~cks/space/blog/linux/SocketAbstractNamespace
* https://utcc.utoronto.ca/~cks/space/blog/python/AbstractUnixSocketsAndPeercred
* http://blog.eduardofleury.com/archives/2007/09/13

Also, note that [the visual representation of abstract UNIX sockets often uses a `@` prefix instead of the non-printable `\0`](https://unix.stackexchange.com/a/206395/27133) but when it reaches down the socket level, it still has to be converted to `\0` internally.

OTOH our tests use a `requests` adapter `requests-unixsocket` for the representation of URLs involving UNIX domain sockets in general, which only supports [`\0` and not `@` as a marker of an abstract socket](https://github.com/msabramo/requests-unixsocket#abstract-namespace-sockets).

webknjaz:
> Aha. It looks like you filed a ticket about this in #276. Since neither the code nor the commit linked to the bug, I was unaware of the issue.

Yeah, I forgot that I actually filed an issue and didn't reference it. So at some point, I just went ahead and committed the fix https://github.com/cherrypy/cheroot/commit/0eb77c82ebdb20fedc58682162fd76aa50a212b3.

webknjaz:
@jaraco how about we reuse your patch from the past? https://github.com/cherrypy/cherrypy/pull/1807/files#diff-e7f19ec238d48a5d2b77ecbbe03850bcR403

jaraco:
In this case, I think attempting to construct an `HTTPConnection(abstract_socket_name)` is probably invalid usage, and since it doesn't seem that `get_server_client` is used anywhere, I think it's best to just remove it.

PR #282, opened 2020-05-06
Avoid possible race condition on persistent HTTP connections

meaksh:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

❓ **What is the current behavior?**

There is a race condition on persistent HTTP connections when a HTTP client reuses a connection where the `socket.timeout` (defined by HTTP Server "timeout" attribute) exception was already triggered on `cheroot` side, the connection was set to be closed but the actual FIN packet is not yet produced by the `HTTPConnection` ([1] and [2]).

When that happens, the client gets a "connection reset by peer" after writing the request:

```console
$ tcpdump  # cherrypy running at 9080 with default "timeout" = 10
...
5485	2020-05-06 11:07:50,199544	127.0.0.1	127.0.0.1	HTTP	382	POST /run HTTP/1.1  (application/json)
5486	2020-05-06 11:07:50,242602	127.0.0.1	127.0.0.1	TCP	66	9080 → 44766 [ACK] Seq=27867 Ack=10594 Win=96512 Len=0 TSval=693100054 TSecr=693100010
5487	2020-05-06 11:07:50,244742	127.0.0.1	127.0.0.1	TCP	361	9080 → 44766 [PSH, ACK] Seq=27867 Ack=10594 Win=96512 Len=295 TSval=693100056 TSecr=693100010 [TCP segment of a reassembled PDU]
5488	2020-05-06 11:07:50,244750	127.0.0.1	127.0.0.1	TCP	66	44766 → 9080 [ACK] Seq=10594 Ack=28162 Win=465536 Len=0 TSval=693100056 TSecr=693100056
5489	2020-05-06 11:07:50,244813	127.0.0.1	127.0.0.1	HTTP	555	HTTP/1.1 200 OK  (application/json)
5490	2020-05-06 11:07:50,244819	127.0.0.1	127.0.0.1	TCP	66	44766 → 9080 [ACK] Seq=10594 Ack=28651 Win=465536 Len=0 TSval=693100056 TSecr=693100056
5502	2020-05-06 11:08:00,919903	127.0.0.1	127.0.0.1	HTTP	382	POST /run HTTP/1.1  (application/json)
5503	2020-05-06 11:08:00,919911	127.0.0.1	127.0.0.1	TCP	66	9080 → 44766 [ACK] Seq=28651 Ack=10910 Win=98432 Len=0 TSval=693110731 TSecr=693110731
5509	2020-05-06 11:08:03,955156	127.0.0.1	127.0.0.1	TCP	66	9080 → 44766 [RST, ACK] Seq=28651 Ack=10910 Win=98432 Len=0 TSval=693113766 TSecr=693110731
```

Notice the time between last HTTP request finished (response) and next POST > 10 sec, the `socket.timeout` exception is raised and then "connection reset by peer":

```console
5489	2020-05-06 11:07:50,244813	127.0.0.1	127.0.0.1	HTTP	555	HTTP/1.1 200 OK  (application/json)
5490	2020-05-06 11:07:50,244819	127.0.0.1	127.0.0.1	TCP	66	44766 → 9080 [ACK] Seq=10594 Ack=28651 Win=465536 Len=0 TSval=693100056 TSecr=693100056
5502	2020-05-06 11:08:00,919903	127.0.0.1	127.0.0.1	HTTP	382	POST /run HTTP/1.1  (application/json)
```

This commit makes a HTTP client to know about this "Keep-Alive" idle timeout by exposing it on the HTTP "Keep-Alive" response header, so the connection won't be reused by the HTTP client if it was "idle" for that "timeout" after the last request response.

[1]: https://github.com/meaksh/cheroot/blob/107650d59cedf8ca614cee713f56148f975c3140/cheroot/server.py#L1352
[2]: https://github.com/meaksh/cheroot/blob/107650d59cedf8ca614cee713f56148f975c3140/cheroot/server.py#L736

A HTTP client is now aware of "Keep-Alive" timeout and won't reuse that connection if the "idle" timeout was already reached.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/282)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1175,6 +1175,11 @@ class HTTPRequest:
                 if not self.close_connection:
                     self.outheaders.append((b'Connection', b'Keep-Alive'))
 
+        self.outheaders.append((
+            b'Keep-Alive',
+            "timeout={}".format(self.server.timeout).encode('ISO-8859-1'),
+        ))
+
         if (not self.close_connection) and (not self.chunked_read):
             # Read any remaining request body data on the socket.
             # "If an origin server receives a request that does not include an
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1175,7 +1175,7 @@ class HTTPRequest:
                 if not self.close_connection:
                     self.outheaders.append((b'Connection', b'Keep-Alive'))
 
-        if dict(self.outheaders).get(b'Connection', b'') == b'Keep-Alive':
+        if (b'Connection', b'Keep-Alive') in self.outheaders:
             self.outheaders.append((
                 b'Keep-Alive',
                 u'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1177,7 +1177,7 @@ class HTTPRequest:
 
         self.outheaders.append((
             b'Keep-Alive',
-            "timeout={}".format(self.server.timeout).encode('ISO-8859-1'),
+            'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
         ))
 
         if (not self.close_connection) and (not self.chunked_read):
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1178,7 +1178,7 @@ class HTTPRequest:
         if dict(self.outheaders).get(b'Connection', b'') == b'Keep-Alive':
             self.outheaders.append((
                 b'Keep-Alive',
-                'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
+                u'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
             ))
 
         if (not self.close_connection) and (not self.chunked_read):
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1175,10 +1175,11 @@ class HTTPRequest:
                 if not self.close_connection:
                     self.outheaders.append((b'Connection', b'Keep-Alive'))
 
-        self.outheaders.append((
-            b'Keep-Alive',
-            'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
-        ))
+        if dict(self.outheaders).get(b'Connection', b'') == b'Keep-Alive':
+            self.outheaders.append((
+                b'Keep-Alive',
+                'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
+            ))
 
         if (not self.close_connection) and (not self.chunked_read):
             # Read any remaining request body data on the socket.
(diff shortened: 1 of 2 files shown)


webknjaz:
It's a good idea but I'd wait until to wait for the connection manager fix in @the-allanc's PR first.

meaksh:
@webknjaz I've rebased this branch with latest `master` and also pushed a check for the `Keep-Alive` timeout header on `test_conn` tests. I think this is now ready :+1:

cheroot/server.py
```diff
@@ -1175,6 +1175,11 @@ def send_headers(self):
                 if not self.close_connection:
                     self.outheaders.append((b'Connection', b'Keep-Alive'))
 
+        self.outheaders.append((
+            b'Keep-Alive',
+            'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
+        ))
```
webknjaz, line 1181: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive, `Keep-Alive` only has meaning if there's `Connection: Keep-Alive` set.

This is also confirmed by https://tools.ietf.org/html/rfc2068#section-19.7.1.1:
>   If the Keep-Alive header is sent, the corresponding connection token
>   MUST be transmitted. The Keep-Alive header MUST be ignored if
>   received without the connection token.

So this should be done under the condition that `connection` is set to `keep-alive`.

meaksh: @webknjaz I've adjusted code and tests accordingly :+1:

cheroot/server.py
```diff
@@ -1175,6 +1175,12 @@ def send_headers(self):
                 if not self.close_connection:
                     self.outheaders.append((b'Connection', b'Keep-Alive'))
 
+        if dict(self.outheaders).get(b'Connection', b'') == b'Keep-Alive':
```
webknjaz, line 1178: I converting to dict would probably be more expensive than having a lookup helper function that would exit early. @meaksh WDYT?

webknjaz: Or maybe this would be a bit faster
```suggestion
        if (b'Connection', b'Keep-Alive') in set(self.outheaders):
```

cheroot/server.py
```diff
@@ -1175,6 +1175,12 @@ def send_headers(self):
...
+            self.outheaders.append((
+                b'Keep-Alive',
+                'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
```
webknjaz, line 1181: Since we still support py2, let's make sure that we always encode unicode text explicitly
```suggestion
                u'timeout={}'.format(self.server.timeout).encode('ISO-8859-1'),
```

cheroot/server.py
meaksh: Yeah, I think we could simply iterate the `self.outheaders` list and do tuple comparison without any casting - I did some tests and that seems to be the faster way.

meaksh:
@webknjaz I've pushed few commits to address your requests :+1:

webknjaz:
Thanks, @meaksh! Released as v8.4.3: https://pypi.org/project/cheroot/8.4.3/.

meaksh:
Thanks a lot @webknjaz ! :+1:

PR #301, opened 2020-07-17
Use selectors lib to address select() fd num limit

### LINKED ISSUE #249: [cheroot==8.1.0 regression] Spin on Linux when socket filedescriptor is too big for select
Once the filedescriptor for a socket used by cherrypy has a value greater than 1024 a thread goes in to a loop just writing blocks like the following to stderr:
```
[22/Nov/2019:11:55:07] ENGINE Error in HTTPServer.tick
  File "/usr/lib/python3.6/dist-packages/cheroot/server.py", line 1770, in serve
    self.tick()
  File "/usr/lib/python3.6/dist-packages/cheroot/server.py", line 1993, in tick
    conn = self.connections.get_conn(self.socket)
  File "/usr/lib/python3.6/dist-packages/cheroot/connections.py", line 142, in get_conn
    rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
ValueError: filedescriptor out of range in select()
```

This does not happen with cheroot versions prior to 8.1.0

tommilligan:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Fixes #249

`HTTPServer` errors if it tries to open a filedescriptor larger than `1024`.

A failing test `test_high_number_of_file_descriptors` has been added to capture this error.

The call to `select` has been replaced with the `selectors/selectors2` module.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/301)

### DIFF
--- a/cheroot/_compat.py
+++ b/cheroot/_compat.py
@@ -1,3 +1,4 @@
+# pylint: disable=unused-import
 """Compatibility code for using Cheroot with various versions of Python."""
 
 from __future__ import absolute_import, division, print_function
@@ -11,7 +12,10 @@ import six
 try:
     import selectors  # lgtm [py/unused-import]
 except ImportError:
-    import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
+    try:
+        import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
+    except ImportError:
+        import selectors34 as selectors  # noqa: F401 # lgtm [py/unused-import]
 
 try:
     import ssl
--- a/cheroot/_compat.py
+++ b/cheroot/_compat.py
@@ -8,6 +8,12 @@ import re
 
 import six
 
+# Not used in this file, imported elsewhere as 'from ._compat import selectors'
+try:
+    import selectors
+except ImportError:
+    import selectors2 as selectors  # noqa: F401
+
 try:
     import ssl
     IS_ABOVE_OPENSSL10 = ssl.OPENSSL_VERSION_INFO >= (1, 1)
--- a/cheroot/_compat.py
+++ b/cheroot/_compat.py
@@ -8,11 +8,10 @@ import re
 
 import six
 
-# Not used in this file, imported elsewhere as 'from ._compat import selectors'
 try:
-    import selectors
+    import selectors  # lgtm [py/unused-import]
 except ImportError:
-    import selectors2 as selectors  # noqa: F401
+    import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
 try:
     import ssl
--- a/cheroot/_compat.py
+++ b/cheroot/_compat.py
@@ -12,10 +12,7 @@ import six
 try:
     import selectors  # lgtm [py/unused-import]
 except ImportError:
-    try:
-        import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
-    except ImportError:
-        import selectors34 as selectors  # noqa: F401 # lgtm [py/unused-import]
+    import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
 try:
     import ssl
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -5,7 +5,10 @@ __metaclass__ = type
 
 import io
 import os
-import select
+try:
+    import selectors
+except ImportError:
+    import selectors2 as selectors
 import socket
 import time
 
@@ -62,6 +65,7 @@ class ConnectionManager:
         """
         self.server = server
         self.connections = []
+        self._selector = selectors.DefaultSelector()
 
     def put(self, conn):
         """Put idle connection into the ConnectionManager to be managed.
@@ -133,10 +137,9 @@ class ConnectionManager:
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
-            # No available socket.
-            if not rlist:
-                return None
+            for sock in socket_dict:
+                self._selector.register(sock, selectors.EVENT_READ)
+            key_events = self._selector.select(timeout=0.1)
         except OSError:
             # Mark any connection which no longer appears valid.
             for fno, conn in list(socket_dict.items()):
@@ -155,15 +158,29 @@ class ConnectionManager:
 
             # Wait for the next tick to occur.
             return None
+        finally:
+            for sock in socket_dict:
+                self._selector.unregister(sock)
 
-        try:
-            # See if we have a new connection coming in.
-            rlist.remove(ss_fileno)
-        except ValueError:
-            # No new connection, but reuse existing socket.
-            conn = socket_dict[rlist.pop()]
-        else:
-            conn = server_socket
+        # See if we have a new connection coming in.
+        new_connection = False
+        rlist = []
+        for key, event in key_events:
+            if key.fd == ss_fileno:
+                # If we have a new connection, reuse the server socket
+                new_connection = True
+                conn = server_socket
+            else:
+                # Make a list of all the other sockets ready to read
+                rlist.append(key.fd)
+
+        if new_connection is False:
+            if len(rlist) == 0:
+                # If we didn't get any readable sockets, wait for the next tick
+                return None
+            else:
+                # No new connection, but reuse existing socket.
+                conn = socket_dict[rlist.pop()]
 
         # All remaining connections in rlist should be marked as ready.
         for fno in rlist:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -134,9 +134,12 @@ class ConnectionManager:
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            for sock in socket_dict:
-                self._selector.register(sock, selectors.EVENT_READ)
-            key_events = [key.fd for key, _event self._selector.select(timeout=0.1)]
+            for fno in socket_dict:
+                self._selector.register(fno, selectors.EVENT_READ)
+            rlist = [
+                key.fd for key, _event
+                in self._selector.select(timeout=0.1)
+            ]
         except OSError:
             # Mark any connection which no longer appears valid.
             for fno, conn in list(socket_dict.items()):
@@ -156,29 +159,22 @@ class ConnectionManager:
             # Wait for the next tick to occur.
             return None
         finally:
-            for sock in socket_dict:
-                self._selector.unregister(sock)
+            for fno in socket_dict:
+                self._selector.unregister(fno)
 
-        # See if we have a new connection coming in.
         try:
-            key_events.pop(ss_fileno)
-        except KeyError:
-                new_connection = False
-        else:
-                new_connection = True
-                # If we have a new connection, reuse the server socket
-                conn = server_socket
-
-        # Make a list of all the other sockets ready to read
-        rlist = list(key_events.keys())
-
-        if not new_connection:
+            # See if we have a new connection coming in.
+            rlist.remove(ss_fileno)
+        except ValueError:
+            # If we didn't get any readable sockets, wait for the next tick
             if not rlist:
-                # If we didn't get any readable sockets, wait for the next tick
                 return None
 
-            # No new connection, but reuse the existing socket.
+            # No new connection, but reuse an existing socket.
             conn = socket_dict[rlist.pop()]
+        else:
+            # If we have a new connection, reuse the server socket
+            conn = server_socket
 
         # All remaining connections in rlist should be marked as ready.
         for fno in rlist:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -5,14 +5,11 @@ __metaclass__ = type
 
 import io
 import os
-try:
-    import selectors
-except ImportError:
-    import selectors2 as selectors
 import socket
 import time
 
 from . import errors
+from ._compat import selectors
 from .makefile import MakeFile
 
 import six
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -139,7 +139,7 @@ class ConnectionManager:
         try:
             for sock in socket_dict:
                 self._selector.register(sock, selectors.EVENT_READ)
-            key_events = self._selector.select(timeout=0.1)
+            key_events = [key.fd for key, _event self._selector.select(timeout=0.1)]
         except OSError:
             # Mark any connection which no longer appears valid.
             for fno, conn in list(socket_dict.items()):
@@ -163,19 +163,20 @@ class ConnectionManager:
                 self._selector.unregister(sock)
 
         # See if we have a new connection coming in.
-        new_connection = False
-        rlist = []
-        for key, event in key_events:
-            if key.fd == ss_fileno:
-                # If we have a new connection, reuse the server socket
+        try:
+            key_events.pop(ss_fileno)
+        except KeyError:
+                new_connection = False
+        else:
                 new_connection = True
+                # If we have a new connection, reuse the server socket
                 conn = server_socket
-            else:
-                # Make a list of all the other sockets ready to read
-                rlist.append(key.fd)
+
+        # Make a list of all the other sockets ready to read
+        rlist = list(key_events.keys())
 
         if not new_connection:
-            if len(rlist) == 0:
+            if not rlist:
                 # If we didn't get any readable sockets, wait for the next tick
                 return None
 
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -174,7 +174,7 @@ class ConnectionManager:
                 # Make a list of all the other sockets ready to read
                 rlist.append(key.fd)
 
-        if new_connection is False:
+        if not new_connection:
             if len(rlist) == 0:
                 # If we didn't get any readable sockets, wait for the next tick
                 return None
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -178,9 +178,9 @@ class ConnectionManager:
             if len(rlist) == 0:
                 # If we didn't get any readable sockets, wait for the next tick
                 return None
-            else:
-                # No new connection, but reuse existing socket.
-                conn = socket_dict[rlist.pop()]
+
+            # No new connection, but reuse the existing socket.
+            conn = socket_dict[rlist.pop()]
 
         # All remaining connections in rlist should be marked as ready.
         for fno in rlist:
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 import os
+import resource
 import socket
 import tempfile
 import threading
@@ -244,40 +245,42 @@ def test_high_number_of_file_descriptors():
     This test should cause a server crash, as the server will try to
     select() a file-descriptor higher than 1024.
     """
+    # Get current resource limits to restore them later
+    (soft_limit, hard_limit) = resource.getrlimit(resource.RLIMIT_NOFILE)
     increased_nofile_limit = 2048
-    sockets = []
-    try:
-        import resource
-        import socket
 
+    # Create our server
+    httpserver = HTTPServer(
+        bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT), gateway=Gateway,
+    )
+    # Hoard a lot of file descriptors by opening and storing a lot of sockets
+    test_sockets = []
+
+    try:
         # We have to increase the nofile limit above 1024
         # Otherwise we see a 'Too many files open' error, instead of
         # an error due to the file descriptor number being too high
-        (soft_limit, hard_limit) = resource.getrlimit(resource.RLIMIT_NOFILE)
         resource.setrlimit(
             resource.RLIMIT_NOFILE,
             (increased_nofile_limit, increased_nofile_limit),
         )
 
-        # Hold a bunch of open file descriptors, so the next ones the server
+        # Open a lot of file descriptors, so the next ones the server
         # opens are higher than 1024
         for i in range(1024):
-            new_socket = socket.socket()
-            sockets.append(new_socket)
+            test_sockets.append(socket.socket())
 
-        # Create our server
-        httpserver = HTTPServer(
-            bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT), gateway=Gateway,
-        )
         httpserver.prepare()
         # This will trigger a crash if select() is used in the implementation
         httpserver.tick()
 
     finally:
-        # Stop our server, and close our open resources
+        # Close our open resources
+        for test_socket in test_sockets:
+            test_socket.close()
+
+        # Stop our server
         httpserver.stop()
-        for socket in sockets:
-            socket.close()
 
         # Reset the limit back to the soft limit
         # (setting the hard_limit is an error)
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -5,6 +5,7 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
+from contextlib import closing
 import os
 import socket
 import tempfile
@@ -284,11 +285,8 @@ def test_high_number_of_file_descriptors():
         # This will trigger a crash if select() is used in the implementation
         httpserver.tick()
 
-        sock = socket.socket()
-        try:
+        with closing(socket.socket()) as sock:
             assert sock.fileno() >= 1024
-        finally:
-            sock.close()
 
     finally:
         # Close our open resources
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -246,7 +246,8 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
         'is not available on Windows.'
     ),
 )
-def test_high_number_of_file_descriptors():
+@pytest.mark.forked
+def test_high_number_of_file_descriptors(http_server):
     """Test the server does not crash with a high file-descriptor value.
 
     This test should cause a server crash, as the server will try to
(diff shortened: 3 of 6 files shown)


cheroot/test/test_server.py
```diff
@@ -236,3 +236,49 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        httpserver = HTTPServer(
+            bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT), gateway=Gateway,
+        )
+        httpserver.prepare()
+        # This will trigger a crash if select() is used in the implementation
+        httpserver.tick()
+
+    finally:
+        # Stop our server, and close our open resources
+        httpserver.stop()
```
webknjaz, line 278: This may not always exist, in case something explodes before it's initialized. Let's prevent the `NameError` somehow...

cheroot/test/test_server.py
```diff
@@ -236,3 +236,49 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+    """Test the server does not crash with a high file-descriptor value.
+
+    This test should cause a server crash, as the server will try to
+    select() a file-descriptor higher than 1024.
+    """
+    increased_nofile_limit = 2048
+    sockets = []
+    try:
+        import resource
+        import socket
```
webknjaz, line 251: plz move these on the top of the test module.

cheroot/test/test_server.py
```diff
@@ -236,3 +236,49 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+
+    finally:
+        # Stop our server, and close our open resources
+        httpserver.stop()
+        for socket in sockets:
+            socket.close()
+
+        # Reset the limit back to the soft limit
+        # (setting the hard_limit is an error)
+        resource.setrlimit(resource.RLIMIT_NOFILE, (soft_limit, soft_limit))
```
webknjaz, line 284: This causes
```console
>           resource.setrlimit(resource.RLIMIT_NOFILE, (soft_limit, soft_limit))
E           ValueError: not allowed to raise maximum limit
```
_(https://travis-ci.com/github/cherrypy/cheroot/jobs/362118162#L532)_

cheroot/test/test_server.py
tommilligan: Done, and renamed some other variables that were causing issues with the globals.

cheroot/test/test_server.py
tommilligan: Fixed by removing stuff from the `try` block that didn't need to be in there, so now when the `try` starts `httpserver.stop` definitely exists.

cheroot/test/test_server.py
tommilligan: Hmm, this is odd. I've pushed a fix that _I think_ will work, but I can't repro this locally.

cheroot/test/test_server.py
```diff
@@ -236,3 +237,51 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        )
+
+        # Open a lot of file descriptors, so the next ones the server
+        # opens are higher than 1024
+        for i in range(1024):
+            test_sockets.append(socket.socket())
+
+        httpserver.prepare()
+        # This will trigger a crash if select() is used in the implementation
+        httpserver.tick()
```
webknjaz, line 275: Let's add some assertions (it's a test after all)
```suggestion
        assert sockets[-1].fileno() >= 1024
        # This will trigger a crash if select() is used in the implementation
        httpserver.tick()
        with socket.socket() as sock:
            assert sock.fileno() >= 1024
```

tommilligan: Updated, good idea

webknjaz: FYI it's a good tone to accept suggested changes via GH UI because it preserves the patch authorship.

cheroot/test/test_server.py
```diff
@@ -236,3 +237,54 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        with socket.socket() as sock:
+            assert sock.fileno() >= 1024
+
+    finally:
+        # Close our open resources
+        for test_socket in test_sockets:
+            test_socket.close()
```
webknjaz, line 288: I'm not completely sure that it's needed here because opening sockets makes the kernel dedicate file descriptors with sequential numbers. So when you close them, it shouldn't reuse lower identifiers anyway.
How about closing them where you open these descriptors and not having to hold a variable with them?

tommilligan: I tried this initially - it is required in order to increment the file descriptor number. I'm not sure what the internal logic of how/when descriptors are reallocated, but removing the array of open sockets triggers the assertion you suggested:
```python
            try:
>               assert sock.fileno() >= 1024
E               AssertionError: assert 12 >= 1024
```
I also tried just holding the most recently opened socket at any one time (open a socket, open a second socket, discard the first socket, etc.), but this also hit the assertion.

cheroot/test/test_server.py
```diff
@@ -236,3 +237,59 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+
+        httpserver.prepare()
+        # This will trigger a crash if select() is used in the implementation
+        httpserver.tick()
+
+        sock = socket.socket()
+        try:
+            assert sock.fileno() >= 1024
+        finally:
+            sock.close()
```
webknjaz, line 291: why didn't you keep the context manager? any problems with it? Using CM is normally considered the best practice

tommilligan: It's not compatible with Python 2.7, `socket`s aren't context aware in it.

However I see elsewhere you use
```python
with closing(socket.socket(...)) as sock:
```
so I'll update to match that

cheroot/test/test_server.py
webknjaz: Looks like I was wrong. The file descriptors are reused after close.

tommilligan: I'm happy to try some more things if you think there's a problem with the implementation, but this reliably reproduces the bug in a minimal way, and my knowledge of lower-level constructs isn't massive - I'd rather focus on the solution.

A bit of StackOverflowing suggest that file descriptors aren't always allocated in this monotonically increasing way:
- https://stackoverflow.com/questions/9376189/are-file-descriptors-for-linux-sockets-always-in-increasing-order/9376493
- https://stackoverflow.com/questions/24557357/why-only-the-lowest-available-file-descriptor-is-allocated-in-unix

webknjaz: Yeah, I've done some reading that implies that fds are reused after their deallocation. I guess it's fine to keep the current approach. I'll proceed with other aspects of the review now...

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        # opens are higher than 1024
+        for i in range(1024):
+            test_sockets.append(socket.socket())
+        assert test_sockets[-1].fileno() >= 1024
+
+        httpserver.prepare()
+        # This will trigger a crash if select() is used in the implementation
+        httpserver.tick()
+
+        with closing(socket.socket()) as sock:
```
webknjaz, line 288: Is `closing()` necessary? doesn't socket already support a CM protocol? Any corner cases?

webknjaz: Ah, nevermind, that's py3-only

webknjaz: So let's explicitly note the motivation:
```suggestion
        with closing(socket.socket()) as sock:  # closing is here for py2-compat
```

cheroot/connections.py
```diff
@@ -5,7 +5,10 @@
 
 import io
 import os
-import select
+try:
+    import selectors
+except ImportError:
+    import selectors2 as selectors
```
webknjaz, line 11: let's move this under `_compat` and just do
```python
from ._compat import selectors
```
in the local import section below

cheroot/connections.py
```diff
@@ -155,15 +158,29 @@ def get_conn(self, server_socket):
...
+        for key, event in key_events:
+            if key.fd == ss_fileno:
+                # If we have a new connection, reuse the server socket
+                new_connection = True
+                conn = server_socket
+            else:
+                # Make a list of all the other sockets ready to read
+                rlist.append(key.fd)
+
+        if new_connection is False:
```
webknjaz, line 177: I believe that the Pythonic way is to do
```suggestion
        if not new_connection:
```

cheroot/connections.py
```diff
@@ -155,15 +158,29 @@ def get_conn(self, server_socket):
...
+            if len(rlist) == 0:
+                # If we didn't get any readable sockets, wait for the next tick
+                return None
+            else:
+                # No new connection, but reuse existing socket.
+                conn = socket_dict[rlist.pop()]
```
webknjaz, line 183: Let's reduce the nesting as we already have a guard-expression above
```suggestion

            # No new connection, but reuse the existing socket.
            conn = socket_dict[rlist.pop()]
```

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+    ),
+)
+def test_high_number_of_file_descriptors():
+    """Test the server does not crash with a high file-descriptor value.
+
+    This test should cause a server crash, as the server will try to
+    select() a file-descriptor higher than 1024.
+    """
+    # Imported here, as it is not available on Windows
+    import resource
```
webknjaz, line 256: Import statements must be on top. The conditionally available modules can be ignored with try/except idiom.

But there's another way. pytest has https://docs.pytest.org/en/stable/reference.html#pytest-importorskip helper and seems to be perfect for this use-case:
```suggestion
    resource = pytest.importorskip(
        'resource',
        reason='The "resource" module is Unix-specific',
    )
```

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
         peercreds_text_resp = requests.get(unix_base_uri + PEERCRED_TEXTS_URI)
         peercreds_text_resp.raise_for_status()
         assert peercreds_text_resp.text == expected_textcreds
+
+
+@pytest.mark.skipif(
+    IS_WINDOWS,
+    reason=(
+        'This regression test is for a Linux bug, and the resource module '
+        'is not available on Windows.'
```
webknjaz, line 246: Most of the reason strings don't have a trailing period. So let's keep the same style here too.
```suggestion
        'is not available on Windows'
```

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+    ),
+)
+def test_high_number_of_file_descriptors():
```
webknjaz, line 249: I think, we need to isolate this test with `pytest-forked`
```suggestion
@pytest.mark.forked
def test_high_number_of_file_descriptors():
```

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+
+    This test should cause a server crash, as the server will try to
+    select() a file-descriptor higher than 1024.
+    """
+    # Imported here, as it is not available on Windows
+    import resource
+
+    # Get current resource limits to restore them later
+    (soft_limit, hard_limit) = resource.getrlimit(resource.RLIMIT_NOFILE)
+    increased_nofile_limit = 2048
```
webknjaz, line 260: It may be a good idea to isolate the resource manipulations into a fixture.

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+
+    # Create our server
+    httpserver = HTTPServer(
+        bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT), gateway=Gateway,
+    )
```
webknjaz, line 265: There's `http_server` fixture for this, let's just reuse it in order to keep the same style
```suggestion
    httpserver = http_server.send((ANY_INTERFACE_IPV4, EPHEMERAL_PORT))
```

cheroot/test/test_server.py
```diff
@@ -236,3 +237,65 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        # an error due to the file descriptor number being too high
+        resource.setrlimit(
+            resource.RLIMIT_NOFILE,
+            (increased_nofile_limit, hard_limit),
+        )
+
+        # Open a lot of file descriptors, so the next ones the server
+        # opens are higher than 1024
+        for i in range(1024):
+            test_sockets.append(socket.socket())
```
webknjaz, line 281: There may be a small optimization here: instead of creating this many sockets, we could create new ones until we get an fd>1024.

cheroot/connections.py
```diff
@@ -155,15 +158,29 @@ def get_conn(self, server_socket):
...
-            # See if we have a new connection coming in.
-            rlist.remove(ss_fileno)
-        except ValueError:
-            # No new connection, but reuse existing socket.
-            conn = socket_dict[rlist.pop()]
-        else:
-            conn = server_socket
+        finally:
+            for sock in socket_dict:
+                self._selector.unregister(sock)
```
webknjaz, line 163: I'm not yet familiar with selectors, could you please tell me what happens with sockets between `unregister()` and the following `register()` at some point in time? Since the previous major contribution caused serious regressions, I'd like to be extra careful now...

cheroot/connections.py
```diff
@@ -155,15 +158,29 @@ def get_conn(self, server_socket):
...
+            if key.fd == ss_fileno:
+                # If we have a new connection, reuse the server socket
+                new_connection = True
+                conn = server_socket
+            else:
+                # Make a list of all the other sockets ready to read
+                rlist.append(key.fd)
+
+        if new_connection is False:
+            if len(rlist) == 0:
```
webknjaz, line 178: the idiomatic way is to do
```suggestion
            if not rlist:
```

cheroot/connections.py
webknjaz, line 183: Looking closer at this, I see that it's probably more readable to follow [EAFP](https://stackoverflow.com/a/11360880/595220)
```suggestion
            try:
                # No new connection, but reuse the existing socket.
                conn = socket_dict[rlist.pop()]
            except KeyError:
                raise  # fd not in the dict?
            except IndexError:
                # If we didn't get any readable sockets, wait for the next tick
                return None
```

cheroot/connections.py
```diff
@@ -133,10 +137,9 @@ def get_conn(self, server_socket):
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
-            # No available socket.
-            if not rlist:
-                return None
+            for sock in socket_dict:
+                self._selector.register(sock, selectors.EVENT_READ)
+            key_events = self._selector.select(timeout=0.1)
```
webknjaz, line 142: would it be easier if we converted this into a dict right here?
something like
```suggestion
            key_events = {key.fd: event for key, event in self._selector.select(timeout=0.1)}
```

and this would allow to try/exept poping `ss_fileno` below and turning the rest of it into `rlist = list(key_events.keys())` (making the control flow nice and flat)

cheroot/connections.py
```diff
@@ -155,15 +158,29 @@ def get_conn(self, server_socket):
...
+        new_connection = False
+        rlist = []
+        for key, event in key_events:
+            if key.fd == ss_fileno:
+                # If we have a new connection, reuse the server socket
+                new_connection = True
+                conn = server_socket
+            else:
+                # Make a list of all the other sockets ready to read
+                rlist.append(key.fd)
```
webknjaz, line 175: paired with https://github.com/cherrypy/cheroot/pull/301/files#r456684464 this could be
```suggestion
        try:
            key_events.pop(ss_fileno)
        except KeyError:
                new_connection = False
        else:
                new_connection = True
                # If we have a new connection, reuse the server socket
                conn = server_socket

        # Make a list of all the other sockets ready to read
        rlist = list(key_events.keys())
```

cheroot/connections.py
webknjaz: oh, wait. does the order matter? then maybe use an ordered dict?

webknjaz: Also, I noticed just now that `event` is not used anywhere. So it may make sense to just use a list and drop it. Like
```suggestion
            key_events = [key.fd for key, _event self._selector.select(timeout=0.1)]
```
WDYT?

webknjaz:
@tommilligan thanks for the PR! Now that I see that it also tries to provide a bugfix, the title and the description should be adjusted accordingly.
I'm also marking it as a draft, feel free to undraft it once you think that the effort is close to being mergeable.

cheroot/connections.py
```diff
@@ -133,10 +137,9 @@ def get_conn(self, server_socket):
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
-            # No available socket.
-            if not rlist:
-                return None
+            for sock in socket_dict:
+                self._selector.register(sock, selectors.EVENT_READ)
+            key_events = [key.fd for key, _event self._selector.select(timeout=0.1)]
```
webknjaz, line 142: :laughing: "coding" in browser is hard:
```suggestion
            key_events = [key.fd for key, _event in self._selector.select(timeout=0.1)]
```
(LGTM has found this mistake @ https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-3818976e81a56bcb93f77451590a1d1acc801594 / https://github.com/cherrypy/cheroot/pull/301#issuecomment-660675296)

webknjaz:
@tommilligan FYI I've added an extra commit on top to correct a typo in a list comprehension.

webknjaz:
You should probably add `pytest-forked` to `setup.cfg` for it to work.

cheroot/connections.py
```diff
@@ -133,10 +137,9 @@ def get_conn(self, server_socket):
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
-            # No available socket.
-            if not rlist:
-                return None
+            for sock in socket_dict:
+                self._selector.register(sock, selectors.EVENT_READ)
+            key_events = [key.fd for key, _event in self._selector.select(timeout=0.1)]
```
webknjaz, line 142:
```suggestion
            key_events = [
                key.fd
                for key, _event in self._selector.select(timeout=0.1)
            ]
```

cheroot/test/test_server.py
tommilligan: Implemented

cheroot/connections.py
tommilligan: I'm not quite sure what you're asking here - are you saying we should be tracking events from sockets throughout their whole lifetime? i.e. as a connection is added to `self.connections`, we begin listening for events on it, and then as it's removed, we `unregister` it?

I haven't used sockets a lot before, so guidance/another solution from anyone else is very welcome

cheroot/test/test_server.py
tommilligan: I would say this is only required if we want to reuse the manipulation. As is I feel this would probably not be worth the effort, especially combined with the conditional import of the `resource` module

tommilligan:
@webknjaz updated to pass locally for me, and responded to your comments

cheroot/_compat.py
```diff
@@ -8,6 +8,12 @@
 
 import six
 
+# Not used in this file, imported elsewhere as 'from ._compat import selectors'
+try:
+    import selectors
+except ImportError:
+    import selectors2 as selectors  # noqa: F401
```
webknjaz, line 15: The top comment is redundant: the whole point of this module is to have shims like this.
```suggestion
try:
    import selectors  # lgtm [py/unused-import]
except ImportError:
    import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
```

cheroot/connections.py
webknjaz: I was just wondering how things work because I haven't used this API before. But that's fine if you don't have a detailed answer.

cheroot/test/test_server.py
webknjaz: It's not only about reusing things. Dividing the code into small pieces makes it more easily readable and maintainable. It also helps to have cleaner abstractions w/o leaks...

cheroot/test/test_server.py
```diff
@@ -236,3 +237,77 @@ def test_peercreds_unix_sock_with_lookup(peercreds_enabled_server):
...
+        'This regression test is for a Linux bug, and the resource module '
+        'is not available on Windows'
+    ),
+)
+@pytest.mark.forked
+def test_high_number_of_file_descriptors(http_server):
+    """Test the server does not crash with a high file-descriptor value.
+
+    This test should cause a server crash, as the server will try to
+    select() a file-descriptor higher than 1024.
```
webknjaz, line 254: This is inaccurate. The test should check that it doesn't crash under the buggy conditions.
```suggestion
    This test shouldn't cause a server crash when trying to access
    file-descriptor higher than 1024.

    The earlier implementation used to rely on `select()` syscall that
    doesn't support file descriptors with numbers higher than 1024.
```

cheroot/connections.py
```diff
@@ -155,14 +158,22 @@ def get_conn(self, server_socket):
 
             # Wait for the next tick to occur.
             return None
+        finally:
+            for fno in socket_dict:
```
webknjaz, line 162: This causes `KeyError: '29 is not registered'` on the next line.
Maybe we need to do
```suggestion
            for fno in rlist:
```
WDYT?

webknjaz: https://travis-ci.com/github/cherrypy/cheroot/jobs/362558590#L1622

webknjaz:
@tommilligan I've reshuffled the test code a bit to make it simpler. The major concern now is this https://github.com/cherrypy/cheroot/pull/301/files#r456973220 (https://travis-ci.com/github/cherrypy/cheroot/jobs/362558590#L1622). Do you have any ideas?

webknjaz:
Ah, nevermind, that CI run uses the fixture as opposed to constructing an HTTP server manually. I'm removing the draft label now and I'm going to point some people to this PR to test it too.

cheroot/connections.py
webknjaz: Resolving per https://github.com/cherrypy/cheroot/pull/301#issuecomment-660733161

tommilligan:
@webknjaz thanks, all your changes look good to me.

I don't know if you have any performance regression tests/benchmarks you can run against this change? I'd be interested to see the results myself.

Don't think I have anything else to contribute at this point, happy for you to implement any changes/requests from reviews.

webknjaz:
Right now I just want some folks to run this and confirm that it doesn't explode.

webknjaz:
Hey @kmduk @timurgen @the-allanc @jakeczyz @liamstask @hardikmodha @manthey @kavehtp @andebor, I invite y'all to check out this PR branch and test it in your envs in order to confirm that it doesn't introduce any obvious regressions.

webknjaz:
recheck

cheroot/connections.py
```diff
@@ -133,10 +134,12 @@ def get_conn(self, server_socket):
         ss_fileno = server_socket.fileno()
         socket_dict[ss_fileno] = server_socket
         try:
-            rlist, _, _ = select.select(list(socket_dict), [], [], 0.1)
-            # No available socket.
-            if not rlist:
-                return None
+            for fno in socket_dict:
+                self._selector.register(fno, selectors.EVENT_READ)
```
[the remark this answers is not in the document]
webknjaz: Thanks, @liamstask! Makes sense. @tommilligan WDYT? Mind addressing this?

webknjaz:
Turns out that we hit https://github.com/sethmlarson/selectors2/pull/10 that is fixed in selectors2 master but is unreleased (and the repo is archived now so no hope for the support): https://ci.appveyor.com/project/CherryPy/cheroot/builds/34226356/job/rw8uglw87w3x93il#L762.

OTOH it looks like selectors34 don't suffer from this issue https://github.com/berkerpeksag/selectors34/blob/1.2.0/selectors34.py#L51. So I'm going to add it as another fallback now. Just for Python 2 under Windows.

webknjaz:
So I'm going to merge this today/tomorrow if nobody will object by then. This suggestion https://github.com/cherrypy/cheroot/pull/301#discussion_r458248844 seems desirable but is not a blocker so if @tommilligan will choose not to implement it or do it in a separate PR, it's fine.

webknjaz:
Thanks, @tommilligan! I released this as v8.4.0: https://pypi.org/project/cheroot/8.4.0/.

P.S. Feel free to contribute a follow-up addressing https://github.com/cherrypy/cheroot/pull/301/files#r458248844.

tommilligan:
@webknjaz thanks for all your help on this, and getting a new release out so promptly!

I like the suggestion of refactoring how we manage connections a bit so that the use of `selectors` is more ergonomic - I have started looking at it. I need to think about it a bit more, as we also care about the **order** of the connections, and `selectors.get_map()` will just return an unordered mapping I think. Will keep thinking about it over the weekend.

webknjaz:
Great, thanks! I've filed an issue for the refactoring: https://github.com/cherrypy/cheroot/issues/304.

hardikmodha:
Thanks @tommilligan for working on this. I'll test it out on my setup.

webknjaz:
FTR the socket allocation fixture seems to be flaky under PyPy:
```console
>               assert sock.fileno() >= resource_limit
E               assert 24 >= 2048
E                +  where 24 = <bound method _socketobject.fileno of <socket._socketobject object at 0x000000000a9037c0>>()
E                +    where <bound method _socketobject.fileno of <socket._socketobject object at 0x000000000a9037c0>> = <socket._socketobject object at 0x000000000a9037c0>.fileno
```
_(https://travis-ci.com/github/cherrypy/cheroot/jobs/365058505#L543)_

It'd be interesting to explore why it is different there...

hardikmodha:
Tested the fix with my application. Fix seems to be working. Thanks @@tommilligan @webknjaz!

Issue #305, opened 2020-07-23
[cheroot==8.1.0 regression] Significant performance drop


mar10:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**

This more a note than a bug report:
I run some benchmark against my project ([WsgiDAV](https://github.com/mar10/wsgidav)] and found that there is a drastical drop in performance between cheroot versions 8.0.0 and 8.1.0.

The test uses a homegrown tool ([stressor](https://github.com/mar10/stressor)) that runs a test script ([see here](https://github.com/mar10/wsgidav/blob/master/tests/stressor/test_rw.yaml)):

- 10 users in parallel
  - GET a common file 
  - PUT another file (one per user)
  - GET that file
- Running for 30 seconds

Benchmarks show 11k activities per 30 seconds until v8.0.
Starting with cheroot 8.1 this drops to 2.9k activities per 30 sec.

I also tried with gevent as backend, and it delivered 12.4k activities.

Please note that these results may be due to poor implementation on my tools, but all I changed was WSGI server behind WsgiDAV, so I thought I let you know.
 

💡 **To Reproduce**

See https://github.com/mar10/wsgidav/blob/master/tests/stressor/results.md

If you are interested, I can provide more details.

Cheers
Martin

webknjaz:
Hey @mar10, thanks for the heads up! v8.1.0 introduced a number of functional regressions that are got some fixes in v8.3.0 and v8.4.0 so please also test with these. It was a result of refactoring the connection management to decouple keep-alive TCP connections from HTTP requests. So I'd say that you should also check whether your tool reuses the connections or creates new ones of each HTTP request.

mar10:
Hi Sviatoslav,

I already [tested with 8.3.1 and 8.4.0](https://github.com/mar10/wsgidav/blob/master/tests/stressor/results.md) with same results.

[stressor](https://github.com/mar10/stressor/) uses the requests library with sessions, which *should* [support keep-alive](https://requests.readthedocs.io/en/master/user/advanced/#keep-alive), but I can double check.

However, even if it did *not* reuse connections, would the refactoring explain that cheroot now slows down to 25% (for my specific benchmark)?

webknjaz:
[...]

PR #308, opened 2020-07-26
Fix a significant performance regression introduced in v8.1.0

mar10:
  - [ ] 🐞 bug fix
  - [X] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Hotfixes #305 

Starting with v8.1, Cheroot's performance dropped in certain scenarios. See #305 for details.

The suggested change restores (and tops) the performance again (see benchmarks in #305)

📋 **Other information**:

N/A

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/308)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -136,9 +136,12 @@ class ConnectionManager:
         try:
             for fno in socket_dict:
                 self._selector.register(fno, selectors.EVENT_READ)
+            # The timeout value impacts performance and should be carefully
+            # chosen. Ref:
+            # github.com/cherrypy/cheroot/issues/305#issuecomment-663985165
             rlist = [
                 key.fd for key, _event
-                in self._selector.select(timeout=0.1)
+                in self._selector.select(timeout=0.01)
             ]
         except OSError:
             # Mark any connection which no longer appears valid.


cheroot/connections.py
```diff
@@ -136,9 +136,11 @@ def get_conn(self, server_socket):
         try:
             for fno in socket_dict:
                 self._selector.register(fno, selectors.EVENT_READ)
+            # The timeout value impacts performance and should be carefully
+            # chosen (issue #305)
```
webknjaz, line 140: I prefer having a complete URL in the comment (this is friendlier to envs outside of GH)
```suggestion
            # chosen. Ref:
            # github.com/cherrypy/cheroot/issues/305#issuecomment-663985165
```

webknjaz:
@mar10, thanks for taking a stab at it! Mind cleaning up the branch a bit?

webknjaz:
cc @tommilligan @liamstask @kmduk @timurgen @jakeczyz @liamstask @hardikmodha @manthey @kavehtp @andebor @the-allanc

liamstask:
it looks to me that this could be related to the fact that, currently in `server.tick()`, only one request can be added per tick even if several are available to be processed in `ConnectionManager.connections`.

in this case, since `ConnectionManager` does not register connections previously marked as read-ready with the selector, it's possible that _none_ of the remaining connections in the selector are active, meaning the server is forced to wait for the full timeout duration each time `tick()` is called.

fundamentally, the value of this timeout should not really be particularly sensitive, since the selector should wake up once connections are ready to be processed, regardless of the timeout.

webknjaz:
Thanks for the insight, @liamstask! What you say sounds reasonable, I haven't considered it...
Looks like improving this would require more fundamental changes and the hotfix of changing the timeout would be acceptable short-term. WDYT?

liamstask:
yep - that seems like an OK short term hotfix to me 👍

webknjaz:
Hotfix released as v8.4.2: https://pypi.org/project/cheroot/8.4.2/ (https://travis-ci.com/github/cherrypy/cheroot/jobs/365552834#L519)

PR #309, opened 2020-07-29
connections: refactor, clarify selectors use

### LINKED ISSUE #304: [TODO] Refactor `cheroot.connections` to use `selectors.get_map()`
❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [X] 🐣 refactoring request

_@liamstask suggested in https://github.com/cherrypy/cheroot/pull/301/files#r458248844:_
> not strictly necessary, but it would be preferable to avoid re-registering and unregistering all sockets each time this `select()`s - each `register()` and `unregister()` will typically incur a syscall per socket, which reduces performance unnecessarily, particularly with large numbers of sockets.
>
> the selector already maintains a map of which sockets are registered - see [`get_map()`](https://docs.python.org/3/library/selectors.html#selectors.BaseSelector.get_map) for instance, and the [SelectorKey](https://docs.python.org/3/library/selectors.html#selectors.SelectorKey) allows storing application-specific state in the `data` field if desired.
>
> this may require a bit more reorganization than the current patch, so could be deferred to a separate PR (if the performance impact is acceptable), but should be considered in a follow up.

_@tommilligan said in https://github.com/cherrypy/cheroot/pull/301#issuecomment-662916411:_
> I like the suggestion of refactoring how we manage connections a bit so that the use of `selectors` is more ergonomic - I have started looking at it. I need to think about it a bit more, as we also care about the **order** of the connections, and `selectors.get_map()` will just return an unordered mapping I think. Will keep thinking about it over the weekend.

liamstask:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [x] 📋 refactoring
  - [ ] 💥 other

Resolves #304

The usage of the selector in the connections module is not as clear as it could be, and results in unnecessary register/unregister calls.

Should be no behavioral change.

A few unnecessary register/unregister calls to the selector are removed, and the code is (hopefully) a bit clearer.

📋 **Other information**:

This passes tests locally for me on macOS, python 3.8 but i couldn't easily figure out how to run the linting steps.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/309)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -11,6 +11,7 @@ import time
 
 from . import errors
 from ._compat import selectors
+from ._compat import suppress
 from .makefile import MakeFile
 
 import six
@@ -98,15 +99,15 @@ class ConnectionManager:
 
         # find any connections still registered with the selector
         # that have not been active recently enough.
-        now = time.time()
-        for _, key in self._selector.get_map().items():
-            if key.data == self.server:
-                continue
-
-            conn = key.data
-            if (conn.last_used + self.server.timeout) < now:
-                self._selector.unregister(key.fd)
-                conn.close()
+        threshold = time.time() - self.server.timeout
+        timed_out_connections = (
+            (sock_fd, conn)
+            for _, (_, sock_fd, _, conn) in self._selector.get_map().items()
+            if conn != self.server and conn.last_used < threshold
+        )
+        for sock_fd, conn in timed_out_connections:
+            self._selector.unregister(sock_fd)
+            conn.close()
 
     def get_conn(self):
         """Return a HTTPConnection object which is ready to be handled.
@@ -122,13 +123,9 @@ class ConnectionManager:
             cheroot.server.HTTPConnection instance, or None.
 
         """
-        # Grab file descriptors from sockets, but stop if we find a
-        # connection which is already marked as ready.
-
-        try:
+        # return a readable connection if any exist
+        with suppress(IndexError):
             return self._readable_conns.popleft()
-        except IndexError:
-            pass
 
         # Will require a select call.
         try:
@@ -276,8 +273,9 @@ class ConnectionManager:
 
         self._selector.close()
 
+    @property
     def _num_connections(self):
-        """Return the current number of connections.
+        """The current number of connections.
 
         Includes any in the readable list or registered with the selector,
         minus one for the server socket, which is always registered
@@ -289,4 +287,4 @@ class ConnectionManager:
     def can_add_keepalive_connection(self):
         """Flag whether it is allowed to add a new keep-alive connection."""
         ka_limit = self.server.keep_alive_conn_limit
-        return ka_limit is None or self._num_connections() < ka_limit
+        return ka_limit is None or self._num_connections < ka_limit
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -3,6 +3,7 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
+import collections
 import io
 import os
 import socket
@@ -61,8 +62,10 @@ class ConnectionManager:
                 that uses this ConnectionManager instance.
         """
         self.server = server
-        self.connections = []
+        self._readable_conns = collections.deque()
         self._selector = selectors.DefaultSelector()
+
+        self._selector.register(server.socket.fileno(), selectors.EVENT_READ, data=server)
 
     def put(self, conn):
         """Put idle connection into the ConnectionManager to be managed.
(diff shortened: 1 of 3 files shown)


cheroot/connections.py
```diff
@@ -61,9 +62,11 @@ def __init__(self, server):
                 that uses this ConnectionManager instance.
         """
         self.server = server
-        self.connections = []
+        self._readable_conns = collections.deque()
```
liamstask, line 65: general restructuring: only maintain connections that the selector has previously indicated are ready to read in `_readable_conns`.

all other connections remain registered with the selector, to avoid registering/unregistering them unnecessarily.

cheroot/workers/threadpool.py
```diff
@@ -111,11 +111,6 @@ def run(self):
                 if conn is _SHUTDOWNREQUEST:
                     return
 
-                # Just close the connection and move on.
-                if conn.closeable:
```
liamstask, line 115: this is not strictly needed, but there's no reason to send the conn through the queue to be closed, when we can close it immediately within `connections`

cheroot/server.py
```diff
@@ -2017,10 +2013,7 @@ def resolve_real_bind_addr(socket_):
 
     def tick(self):
         """Accept a new connection and put it on the Queue."""
-        if not self.ready:
-            return
```
liamstask, line 2021: slight reorg here - a single check for `self.ready` at the beginning of this routine is not sufficient, since `stop()` can clear it & close the socket and connections even after this initial check.

this results in `self.connections.get_conn()` or `self.connections.expire()` failing since the selector has been closed.

instead, make `stop()` ensure that `serve()` has returned before closing down the socket and connections.

cheroot/connections.py
```diff
@@ -107,88 +118,61 @@ def get_conn(self, server_socket):
...
-            conn = None
 
-        # We have a connection ready for use.
-        if conn:
-            self.connections.remove(conn)
-            return conn
+        try:
+            return self._readable_conns.popleft()
+        except IndexError:
+            pass
```
webknjaz, line 131: I think it's better semantically to use suppress:
```suggestion
        with suppress(IndexError):
            return self._readable_conns.popleft()
```
_(add `from ._compat import suppress` at the top)_

cheroot/connections.py
```diff
@@ -107,88 +118,61 @@ def get_conn(self, server_socket):
...
 
-        Args:
-            server_socket (socket.socket): Socket to listen to for new
-            connections.
         Returns:
             cheroot.server.HTTPConnection instance, or None.
 
         """
         # Grab file descriptors from sockets, but stop if we find a
         # connection which is already marked as ready.
```
webknjaz, line 126: Is this comment still relevant in this place?

cheroot/connections.py
```diff
@@ -282,12 +266,27 @@ def _from_server_socket(self, server_socket):
...
-        self.connections = []
+        self._readable_conns.clear()
+
+        for _, key in self._selector.get_map().items():
+            if key.data != self.server:  # server closes its own socket
+                key.data.socket.close()
+
+        self._selector.close()
+
+    def _num_connections(self):
```
webknjaz, line 279: This should be a property semantically
```suggestion
    @property
    def _num_connections(self):
```

cheroot/connections.py
```diff
@@ -282,12 +266,27 @@ def _from_server_socket(self, server_socket):
...
+        with the selector.
+        """
+        return len(self._readable_conns) + len(self._selector.get_map()) - 1
 
     @property
     def can_add_keepalive_connection(self):
         """Flag whether it is allowed to add a new keep-alive connection."""
         ka_limit = self.server.keep_alive_conn_limit
-        return ka_limit is None or len(self.connections) < ka_limit
+        return ka_limit is None or self._num_connections() < ka_limit
```
webknjaz, line 292:
```suggestion
        return ka_limit is None or self._num_connections < ka_limit
```

liamstask:
those suggestions lgtm - applied in https://github.com/cherrypy/cheroot/pull/309/commits/3135db560ffa102fd75620bbda9ddef4b22a3c72

webknjaz:
Perfect, thanks!

webknjaz:
> i couldn't easily figure out how to run the linting steps.

FYI it's just `tox -e pre-commit`

webknjaz:
Released as v8.4.4: https://pypi.org/project/cheroot/8.4.4/.

cheroot/connections.py
```diff
@@ -83,21 +96,20 @@ def expire(self):
...
-        # Connection too old?
-        if (conn.last_used + self.server.timeout) < time.time():
-            conn.closeable = True
-            return
+        # find any connections still registered with the selector
+        # that have not been active recently enough.
+        threshold = time.time() - self.server.timeout
+        timed_out_connections = (
+            (sock_fd, conn)
+            for _, (_, sock_fd, _, conn) in self._selector.get_map().items()
```
webknjaz, line 105: The map should be copied before the iteration.

PR #310, opened 2020-08-04
Uncaught socket exception during timeout handling

### LINKED ISSUE #210: Uncaught socket exception during timeout handling.
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request

If a client goes away while a request timeout is being processed, the `socket.error` exception produced in `/cheroot/server.py` `_conditional_error()` bubbles up uncaught all the way to the thread worker. This generates a stack trace regardless of logging config.

This is producing STDERR errors in our container log, preventing us from handling errors properly. I also suspect given enough occurrence it may lead to thread pool starvation although I have not experimented with it.

💡 **To Reproduce**
This is hard to reproduce, I suspect timing is heavily involved. We are seeing it occur in our environment when requests timeout. These requests are coming from a load balancer.

💡 **Expected behavior**
The exception should be caught and treated as a simple failure and simply trigger completion and cleanup of the client's request (the client simply has gone away).

📋 **Details**
Here is the traceback we're seeing:
```
  File "/usr/local/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/python2.7/site-packages/cheroot/workers/threadpool.py", line 114, in run
    conn.communicate()
  File "/usr/local/lib/python2.7/site-packages/cheroot/server.py", line 1286, in communicate
    self._conditional_error(req, '408 Request Timeout')
  File "/usr/local/lib/python2.7/site-packages/cheroot/server.py", line 1334, in _conditional_error
    req.simple_response(response)
  File "/usr/local/lib/python2.7/site-packages/cheroot/server.py", line 1115, in simple_response
  File "/usr/local/lib/python2.7/site-packages/cheroot/makefile.py", line 78, in write
    data_mv[bytes_sent:bytes_sent + SOCK_WRITE_BLOCKSIZE],
  File "/usr/local/lib/python2.7/site-packages/cheroot/ssl/pyopenssl.py", line 144, in send
    *args, **kwargs
  File "/usr/local/lib/python2.7/site-packages/cheroot/ssl/pyopenssl.py", line 91, in _safe_call
    raise socket.error(errnum)
error: -1
```

📋 **Environment**
- Cheroot version: 6.5.4
- CherryPy version: 18.1.2
- Python version: 3.7.3
- OS: Linux Alpine

📋 **Additional context**
I believe a simple try/except/pass for the `socket.error` in the context of https://github.com/cherrypy/cheroot/blob/v6.5.5/cheroot/server.py#L1276 should fix the issue without any negative side-effects since the server is already dealing with a client that has gone away.

ajyoung:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Fixes #210 

When connections with no content that last for longer than the connection timeout are made to a cheroot server backed by pyOpenSSL, the connection timeout on the server side also causes an unhandled socket error that causes the thread to be dropped. Eventually, the server will extinguish its thread pool and become unresponsive.

See https://github.com/cherrypy/cheroot/issues/210#issuecomment-668867894 

Threads are not dropped when connections with no content that last longer than the connection timeout are made to the server.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/310)

### DIFF
--- a/cheroot/ssl/pyopenssl.py
+++ b/cheroot/ssl/pyopenssl.py
@@ -105,8 +105,14 @@ class SSLFileobjectMixin:
             except SSL.WantWriteError:
                 time.sleep(self.ssl_retry)
             except SSL.SysCallError as e:
-                if is_reader and e.args == (-1, 'Unexpected EOF'):
-                    return b''
+                if e.args == (-1, 'Unexpected EOF'):
+                    if is_reader:
+                        return b''
+                    else:
+                        # See #210. Prevents DOS attack caused by
+                        # silent connections lasting beyond connection
+                        # timeout length.
+                        raise errors.FatalSSLAlert(*e.args)
 
                 errnum = e.args[0]
                 if is_reader and errnum in errors.socket_errors_to_ignore:
--- a/cheroot/ssl/pyopenssl.py
+++ b/cheroot/ssl/pyopenssl.py
@@ -108,11 +108,10 @@ class SSLFileobjectMixin:
                 if e.args == (-1, 'Unexpected EOF'):
                     if is_reader:
                         return b''
-                    else:
-                        # See #210. Prevents DOS attack caused by
-                        # silent connections lasting beyond connection
-                        # timeout length.
-                        raise errors.FatalSSLAlert(*e.args)
+                    # See #210. Prevents DOS attack caused by
+                    # silent connections lasting beyond connection
+                    # timeout length.
+                    raise errors.FatalSSLAlert(*e.args)
 
                 errnum = e.args[0]
                 if is_reader and errnum in errors.socket_errors_to_ignore:


ajyoung:
The windows test suite failure has been happening on master for the last ten days as well. But I'm not sure yet about the ci/circleci: macos-build and continuous-integration/appveyor/pr failures.

cheroot/server.py
```diff
@@ -1337,6 +1337,8 @@ def _conditional_error(self, req, response):
 
         try:
             req.simple_response(response)
+        except socket.error:
```
webknjaz, line 1340: I think this would need a long comment explaining why this is necessary.

webknjaz:
@ajyoung there's a few TLS tests that are flaky on slow machines like macOS and Windows. If they happen on master, ignore those.

webknjaz (changes requested):
@ajyoung thanks for the PR! The fix looks correct but this patch also needs a regression test so that it's checked in the CI and is not only confirmed working in your specific env.

webknjaz:
@ajyoung any updates?

ajyoung:
@webknjaz -- Sorry for the delay. I haven't forgotten about it but I need to the get the environment setup again so I can run and write the required tests. I'll try to get back to this before Thanksgiving in the US (November 27)

ajyoung:
@webknjaz -- It looks like the test_wsgi tests will no longer run under python 2.7 due to warnings like "CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release"

Do you know a way to suppress this warning when running the tests under python 2.7?

webknjaz:
It's already suppressed on master and doesn't prevent anything from being executed in this PR. You may be looking at some old CI runs. Your test fails because you attempt using socket as a CM while it doesn't have such interface.

ajyoung:
@webknjaz -- I fixed the python 2.7 syntax. The tests pass here on linux but one concern is that I tried the test on the master branch without the fix and the test passes there as well. It seems that either the attempt to set the thread count to 1 or the attempt to set the timeout to 1s is incorrect in the test (so the server isn't actually dropping all threads with just two client calls). 

I can verify that when I run the master branch manually and the 210 branch manually, that I can reproduce the issue with the same python client code. I run the code below eleven times against the server running on the master branch and it drops all threads. But that doesn't seem to be happening in the automated test.

`[root@ajym-1 ~]# cat sslperftest-simple-client-py2.py
#!/usr/bin/python2

import socket
import time

socket_type = socket.AF_INET

s = socket.socket(socket_type, socket.SOCK_STREAM)
s.connect(('ajym-1.ajym.root.hwx.site', 11827))
time.sleep(12)
s.close()`

webknjaz:
@ajyoung that looks like a half of a reproducer, it'd be useful to see the exact server part of the repro too.

ajyoung:
@webknjaz  -- The server reproduction is [here](https://github.com/cherrypy/cheroot/issues/210#issuecomment-668867894).

```
[root@host-4 ~]vi sslperftest-simple-pyopenssl.py py

#!/usr/bin/python2
certfile='/tmp/ssl-cert.pem'
keyfile='/tmp/ssl-cert-pk.pem'
cafile=None
ciphers=None

host='host-4.example.com'
port=11827

def raw_wsgi_app(environ, start_response):
    status = '200 OK'
    response_headers = [('Content-type','text/plain')]
    start_response(status, response_headers)
    return ['Hello world!']

from cheroot.ssl.pyopenssl import pyOpenSSLAdapter
from cheroot import wsgi

bind_addr = (host, port)
server = wsgi.Server(bind_addr, raw_wsgi_app, request_queue_size=32)
server.ssl_adapter = pyOpenSSLAdapter(certfile, keyfile, certificate_chain=cafile, ciphers=ciphers)

try:
    server.start()
except KeyboardInterrupt:
    server.stop()
```

cheroot/server.py
```diff
@@ -1343,6 +1343,11 @@ def _conditional_error(self, req, response):
...
```
webknjaz, line 1346: After giving it some thought, I think that this a wrong fix. It relies on the leaking implementation detail.

Instead, we need to change the pyopenssl adapter code to raise a standardized exception defined in https://github.com/cherrypy/cheroot/blob/master/cheroot/errors.py, like in other except-blocks. This would abstract us from the underlying causes. The exception should probably inherit https://docs.python.org/3/library/exceptions.html#ConnectionError.

webknjaz:
@ajyoung looks like this failing test is related to the changes in this PR: https://github.com/cherrypy/cheroot/pull/310/checks?check_run_id=1397448854#step:14:682

cheroot/test/test_ssl.py
```diff
@@ -729,3 +749,69 @@ def test_http_over_https_error(
...
+        tls_certificate_chain_pem_path, tls_certificate_private_key_pem_path,
+    )
+    if adapter_type == 'pyopenssl':
+        tls_adapter.context = tls_adapter.get_context()
+
+    tls_certificate.configure_cert(tls_adapter.context)
+
+    interface, _host, port = _get_conn_data(ip_addr)
+    tlshttpserver = \
+        tls_single_thread_http_server((interface, port), tls_adapter)
```
webknjaz, line 796: Escaping EOL is harmful, wrap with parentheses instead
```suggestion
    tlshttpserver = (
        tls_single_thread_http_server((interface, port), tls_adapter)
    )
```

cheroot/test/test_ssl.py
```diff
@@ -729,3 +749,69 @@ def test_http_over_https_error(
...
+    fqdn = interface
+    socket_type = socket.AF_INET
+    if ip_addr is ANY_INTERFACE_IPV6:
+        fqdn = '[{fqdn}]'.format(**locals())
+        socket_type = socket.AF_INET6
+
+    s = socket.socket(socket_type, socket.SOCK_STREAM)
+    s.connect((ip_addr, port))
+    time.sleep(3)
+    s.close()
```
webknjaz, line 811: Maybe close this socket at the end of the test function?

cheroot/test/test_ssl.py
```diff
@@ -729,3 +749,69 @@ def test_http_over_https_error(
...
+    ),
+)
+@pytest.mark.parametrize(
+    'ip_addr',
+    (
+        ANY_INTERFACE_IPV4,
+        pytest.param(ANY_INTERFACE_IPV6, marks=missing_ipv6),
+    ),
+)
+def test_server_timeout_before_content_error(
```
webknjaz, line 768: This test doesn't have any asserts: I'd expect it to use something like `capfd` fixture and check what server outputs.

cheroot/server.py
ajyoung: @webknjaz  -- I suppose in that case, we'd want to just add the errnum to errors.socket_errors_to_ignore and not throw an exception at all. But in this case, the errnum is -1. It cannot be translated to a specific error. Do we want to handle *all* -1 errors this way?

webknjaz: TL;DR We need to have our own exception declared for communicating this.

`-1` is not at error code, it's just how OpenSSL signals that there's *something* wrong AFAIU. IIUC OpenSSL is implemented as a state machine and to figure out what error state it is in, one should call [`SSL_get_error()`](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html). For example, here's what PyOpenSSL does: https://github.com/pyca/pyopenssl/blob/52341e8/src/OpenSSL/SSL.py#L1530.

Not throwing an exception is a no-go. It'll probably break some of the mechanisms. Also, exceptions is how we communicate problems while returning some values mean that there's no problems at all which is not at all true.

ajyoung:
Just an update on this -- I will get back to it after Christmas, just before the new year.

cheroot/server.py
ajyoung: @webknjaz -- I've looked at this more closely now. At the point highlighted above where the socket.error occurs, it is already clear that the underlying error is an [SSL.SysCallError](https://github.com/cherrypy/cheroot/blob/0ebe8eebe83685656ff6c14bd1ab0fd2911e9b2b/cheroot/ssl/pyopenssl.py#L106-L113). At the point of execution highlighted in server.py above, the ssl context isn't available but the stack trace makes this clear:

`Traceback (most recent call last):
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/workers/threadpool.py", line 125, in run
    keep_conn_open = conn.communicate()
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/server.py", line 1290, in communicate
    self._conditional_error(req, '408 Request Timeout')
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/server.py", line 1339, in _conditional_error
    req.simple_response(response)
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/server.py", line 1111, in simple_response
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/makefile.py", line 78, in write
    data_mv[bytes_sent:bytes_sent + SOCK_WRITE_BLOCKSIZE],
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/ssl/pyopenssl.py", line 163, in send
    *args, **kwargs
  File "/opt/cloudera/parcels/KEYTRUSTEE_SERVER/lib/python2.7/site-packages/cheroot/ssl/pyopenssl.py", line 110, in _safe_call
    raise socket.error(errnum)
error: -1`

webknjaz: Sounds like we need to raise a better error there. How about `FatalSSLAlert`?

ajyoung: I've pushed a new commit that makes the fix by raising FatalSSLAlert in cheroot/ssl/pyopenssl.py instead of ignoring the socket.error in cheroot/server.py.

PR #311, opened 2020-08-13
server: process connections as they become active, rather than on tick

### LINKED ISSUE #305: [cheroot==8.1.0 regression] Significant performance drop

liamstask:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [x] 📋 refactoring
  - [ ] 💥 other

Fixes #305

Currently, the server `tick()`s in order to retrieve connections to process. However, in the current implementation, even if connections are previously marked as ready for processing, they are not returned from `connections.get_conn()` until after polling the selector. This means the processing of each connection may incur the latency of the polling timeout interval.

This addresses #305 in a more robust way, since this means the timeout specified to `select()` does not meaningfully impact the latency with which connections are processed.

Instead of caching read-ready connections in `connections`, instead push new connections directly into the thread pool (via the server) as they become available. Any back pressure from the thread pool queue will propagate to `connections`, meaning subsequent polling of the selector will be delayed in the event that no workers are available, which is appropriate.

📋 **Other information**:

Tests and lint pass locally for me on macOS, python 3.8

It is probably worth repeating some of the performance testing done in https://github.com/cherrypy/cheroot/issues/305 to verify this - I have not yet reproduced per the notes in that issue, if anybody can assist that would be 👍

Performance gains may not be substantial, but I think the structure is easier to understand in any event.

Please let me know how best to handle what may be a breaking api change from `tick()` -> `serve()` or please update the PR directly if you prefer.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/311)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -3,7 +3,6 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-import collections
 import io
 import os
 import socket
@@ -126,8 +125,10 @@ class ConnectionManager:
             server (cheroot.server.HTTPServer): web server object
                 that uses this ConnectionManager instance.
         """
+        self._serving = False
+        self._stop_requested = False
+
         self.server = server
-        self._readable_conns = collections.deque()
         self._selector = _ThreadsafeSelector()
 
         self._selector.register(
@@ -145,13 +146,13 @@ class ConnectionManager:
         # if this conn doesn't have any more data waiting to be read,
         # register it with the selector.
         if conn.rfile.has_data():
-            self._readable_conns.append(conn)
+            self.server.process_conn(conn)
         else:
             self._selector.register(
                 conn.socket.fileno(), selectors.EVENT_READ, data=conn,
             )
 
-    def expire(self):
+    def _expire(self):
         """Expire least recently used connections.
 
         This happens if there are either too many open connections, or if the
@@ -171,48 +172,57 @@ class ConnectionManager:
             self._selector.unregister(sock_fd)
             conn.close()
 
-    def get_conn(self):
-        """Return a HTTPConnection object which is ready to be handled.
+    def stop(self):
+        """Stop the selector loop in run() synchronously.
 
-        A connection returned by this method should be ready for a worker
-        to handle it. If there are no connections ready, None will be
-        returned.
-
-        Any connection returned by this method will need to be `put`
-        back if it should be examined again for another request.
-
-        :returns: HTTPConnection instance, or None
-        :rtype: cheroot.server.HTTPConnection
+        May take up to half a second.
         """
-        # return a readable connection if any exist
-        with suppress(IndexError):
-            return self._readable_conns.popleft()
+        self._stop_requested = True
+        while self._serving:
+            time.sleep(0.01)
 
-        # Will require a select call.
-        try:
-            # The timeout value impacts performance and should be carefully
-            # chosen. Ref:
-            # github.com/cherrypy/cheroot/issues/305#issuecomment-663985165
-            active_list = self._selector.select(timeout=0.01)
-        except OSError:
-            self._remove_invalid_sockets()
-            # Wait for the next tick to occur.
-            return None
+    def run(self, expiration_interval):
+        """Run the connections selector indefinitely.
 
-        for (sock_fd, conn) in active_list:
-            if conn is self.server:
-                # New connection
-                return self._from_server_socket(self.server.socket)
+        Args:
+            expiration_interval (float): Interval, in seconds, at which
+                connections will be checked for expiration.
 
-            # unregister connection from the selector until the server
-            # has read from it and returned it via put()
-            self._selector.unregister(sock_fd)
-            self._readable_conns.append(conn)
+        Connections that are ready to process are submitted via
+        self.server.process_conn()
 
-        try:
-            return self._readable_conns.popleft()
-        except IndexError:
-            return None
+        Connections submitted for processing must be `put()`
+        back if they should be examined again for another request.
+
+        Can be shut down by calling `stop()`.
+        """
+        self._serving = True
+        last_expiration_check = time.time()
+
+        while not self._stop_requested:
+            try:
+                active_list = self._selector.select(timeout=0.01)
+            except OSError:
+                self._remove_invalid_sockets()
+                continue
+
+            for (sock_fd, conn) in active_list:
+                if conn is self.server:
+                    # New connection
+                    new_conn = self._from_server_socket(self.server.socket)
+                    self.server.process_conn(new_conn)
+                else:
+                    # unregister connection from the selector until the server
+                    # has read from it and returned it via put()
+                    self._selector.unregister(sock_fd)
+                    self.server.process_conn(conn)
+
+            now = time.time()
+            if (now - last_expiration_check) > expiration_interval:
+                self._expire()
+                last_expiration_check = now
+
+        self._serving = False
 
     def _remove_invalid_sockets(self):
         """Clean up the resources of any broken connections.
@@ -332,10 +342,6 @@ class ConnectionManager:
 
     def close(self):
         """Close all monitored connections."""
-        for conn in self._readable_conns:
-            conn.close()
-        self._readable_conns.clear()
-
         for (_, conn) in self._selector.connections:
             if conn is not self.server:  # server closes its own socket
                 conn.close()
@@ -345,11 +351,11 @@ class ConnectionManager:
     def _num_connections(self):
         """Return the current number of connections.
 
-        Includes any in the readable list or registered with the selector,
+        Includes all conns registered with the selector,
         minus one for the server socket, which is always registered
         with the selector.
         """
-        return len(self._readable_conns) + len(self._selector) - 1
+        return len(self._selector) - 1
 
     @property
     def can_add_keepalive_connection(self):
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -7,12 +7,13 @@ sticking incoming connections onto a Queue::
 
     server = HTTPServer(...)
     server.start()
-    ->  while True:
-            tick()
-            # This blocks until a request comes in:
-            child = socket.accept()
-            conn = HTTPConnection(child, ...)
-            server.requests.put(conn)
+    ->  serve()
+        while ready:
+            _connections.run()
+                while not stop_requested:
+                    child = socket.accept()  # blocks until a request comes in
+                    conn = HTTPConnection(child, ...)
+                    server.process_conn(conn)  # adds conn to threadpool
 
 Worker threads are kept in a pool and poll the Queue, popping off and then
 handling each connection in turn. Each connection can consist of an arbitrary
@@ -1523,6 +1524,11 @@ class HTTPServer:
     timeout = 10
     """The timeout in seconds for accepted connections (default 10)."""
 
+    expiration_interval = 0.5
+    """The interval, in seconds, at which the server checks for
+    expired connections (default 0.5).
+    """
+
     version = 'Cheroot/{version!s}'.format(version=__version__)
     """A version string for the HTTPServer."""
 
@@ -1592,7 +1598,6 @@ class HTTPServer:
         self.requests = threadpool.ThreadPool(
             self, min=minthreads or 1, max=maxthreads,
         )
-        self.serving = False
 
         if not server_name:
             server_name = self.version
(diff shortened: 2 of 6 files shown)


cheroot/connections.py
```diff
@@ -99,76 +99,87 @@ def expire(self):
         # find any connections still registered with the selector
         # that have not been active recently enough.
         threshold = time.time() - self.server.timeout
-        timed_out_connections = (
+        timed_out_connections = [
             (sock_fd, conn)
             for _, (_, sock_fd, _, conn) in self._selector.get_map().items()
             if conn != self.server and conn.last_used < threshold
-        )
+        ]
```
liamstask, line 106: i found that without capturing these to a separate list, i was seeing errors modifying the selector map during iteration in the `for sock_fd, conn in timed_out_connections` block below

cheroot/connections.py
```diff
@@ -99,76 +99,87 @@ def expire(self):
...
 
-                try:
-                    os.fstat(key.fd)
-                except OSError:
-                    # Socket is invalid, close the connection
-                    self._selector.unregister(key.fd)
+        Can be shut down by calling `stop()`.
+        """
+        self._serving = True
+        expiration_interval = 0.5
```
liamstask, line 132: this value is a bit arbitrary, but seemed reasonable enough. let me know if you'd like to see this configurable in some other way.

cheroot/connections.py
```diff
@@ -99,76 +99,87 @@ def expire(self):
...
 
-    def get_conn(self):  # noqa: C901  # FIXME
-        """Return a HTTPConnection object which is ready to be handled.
+    def stop(self):
+        """Stop the selector loop in run() synchronously.
 
-        A connection returned by this method should be ready for a worker
-        to handle it. If there are no connections ready, None will be
-        returned.
+        May take up to half a second.
```
liamstask, line 114: i am not aware of a cross-platform way to wake up the selector in response to an event - on Linux I'd typically use something like `eventfd` or create a pipe, but the selector on Windows can only monitor sockets.

Assuming that shutdown does not need to be optimized in practical usage, I figured it was easy enough to just check based on the selector poll timeout instead.

cheroot/connections.py
webknjaz: Yeah: https://github.com/cherrypy/cheroot/pull/309/files#r469927485 / #312.

But the proper fix would be to `.copy()` the dict early. This problem arises from the change in Python 3 (compared to Python 2): `.items()` returns a lazy iterator under py3 while in py2 it was a list.

Just turning this into a list may still hit this race condition because you're iterating over a moving target.

Using `.copy()` early should be atomic enough to prevent this and is recommended to mitigate such cases, in general.

webknjaz, line 106: The outer generator expression may be kept given that we'll make the iterable target immutable by using `.copy()` early.

webknjaz: Also, I think all of the `get_map()` calls should be followed by `.copy()` as long as it's used in a threaded env: https://github.com/cherrypy/cheroot/issues/312#issuecomment-673477303.

cheroot/connections.py
```diff
@@ -99,76 +99,87 @@ def expire(self):
...
-                return self._from_server_socket(self.server.socket)
+        self._serving = False
 
-            conn = key.data
-            # unregister connection from the selector until the server
-            # has read from it and returned it via put()
+    def _remove_invalid_sockets(self):
+        # Mark any connection which no longer appears valid
+        invalid_keys = []
+        for _, key in self._selector.get_map().items():
```
webknjaz, line 167:
```suggestion
        for _, key in self._selector.get_map().copy().items():
```

webknjaz: (some articles suggest using `list(dict().items())` but that's wrong because casting to list iterates over a moving target vs `dict().copy().items()` that allows to iterate over some version of dict that is guaranteed to be immutable)

cheroot/server.py
```diff
@@ -1791,19 +1790,15 @@ def prepare(self):  # noqa: C901  # FIXME
...
-        self.serving = False
+        try:
+            self.connections.run()
+        except (KeyboardInterrupt, SystemExit):
+            raise
+        except Exception:
+            self.error_log(
+                'Error in HTTPServer.serve', level=logging.ERROR,
+                traceback=True,
+            )
```
webknjaz, line 1801: Is this a breaking change? It probably requires updating the module docstring that contains some lifecycle pointers.

cheroot/connections.py
liamstask: it's not clear to me that this is a threading issue - it's an error even in a single-threaded context to modify a dict during iteration, which is what's happening here.

in any event, `get_map()` returns a `collections.abc.Mapping` which does not support `copy()` - what would you prefer in this case? `copy.copy(self._selector.get_map()).items()`?

also - this code appears to be exercised by tests, i don't understand why it's not caught as an error there.

liamstask: https://github.com/cherrypy/cheroot/pull/313 if `copy.copy()` works for you

webknjaz: Interesting... So probably in production envs selector map may change on the low-level more often, I guess.

webknjaz:
Hm. This introduces a number of windows failures in GHA that haven't been happening in the past: https://github.com/cherrypy/cheroot/pull/311/checks?check_run_id=979099442#step:14:351.

(only py2 win jobs were flaky)

cheroot/connections.py
liamstask: oh - no, i think it's just the simple fact that `self._selector.unregister()` modifies the selector map, and if that's called while iterating, an error is raised

webknjaz: Yeah, looks like you're correct.

webknjaz:
@liamstask I believe this needs rebasing

cheroot/connections.py
```diff
@@ -55,40 +55,70 @@ class ConnectionManager:
...
-        self._readable_conns = collections.deque()
         self._selector = selectors.DefaultSelector()
 
         self._selector.register(
             server.socket.fileno(),
             selectors.EVENT_READ, data=server,
         )
 
+        # create & register control socket
+        self._ctrl_rx, self._ctrl_tx = socket.socketpair()
```
liamstask, line 81: NOTE: python 3.5+ includes a pure-python implementation of this on windows, which is provided by https://pypi.org/project/backports.socketpair - let me know if that's an acceptable dependency to introduce

liamstask:
@webknjaz @cyraxjoe i also introduced a commit to address the issues observed in #317 - f3e09256257e335c2fd2a6005355ad2b52273322 (relies on an as yet unimplemented backport on some platforms)

i think it's preferable to only add contention in the case that `put()` is being called (infrequent) rather than all times the selector is accessed. let me know what you think.

liamstask:
also split this out as a separate PR in https://github.com/cherrypy/cheroot/pull/319

cheroot/connections.py
webknjaz: Yes, using backports is reasonable. In fact, this migration to selectors has introduced such a dep recently. Just make sure to use env markers and put the try/except shim under `_compat`:
* https://github.com/cherrypy/cheroot/blame/master/setup.cfg#L68
* [ee7261b](https://github.com/cherrypy/cheroot/pull/301/commits/ee7261b6f8d1a4cf1fbbde10f311b061626d6da9) / [84f8616](https://github.com/cherrypy/cheroot/pull/301/commits/84f86162f76fd18f116822e56fd55fee38cd7bb6)

cheroot/connections.py
```diff
@@ -55,40 +55,70 @@ class ConnectionManager:
...
+            with self._ctrl_lock:
+                self._ctrl_tx.send(self._CTRL_MSG_PUT)
+
+    def _selector_conns(self):
+        # return conns registered with the selector,
+        # omitting the server and control sockets
+        return (
+            (sock_fd, conn)
+            for _, (_, sock_fd, _, conn)
+            in self._selector.get_map().items()
```
webknjaz, line 117: Would it be reasonable to copy the mapping here?

cheroot/connections.py
```diff
@@ -55,40 +55,70 @@ class ConnectionManager:
...
         else:
-            self._selector.register(
-                conn.socket.fileno(), selectors.EVENT_READ, data=conn,
-            )
+            # store the conn in our queue and notify the selector
+            self._put_q.append(conn)
+            with self._ctrl_lock:
+                self._ctrl_tx.send(self._CTRL_MSG_PUT)
+
+    def _selector_conns(self):
```
webknjaz, line 111: This should probably be a property to use a verb in the name.

liamstask:
@webknjaz i probably shouldn't have added that commit to this PR, since it's addressing a separate issue. if you agree, i've created a separate PR with just the fix in #319 - i can always come and rebase this afterwards.

webknjaz:
@liamstask alright, let's work on #319 first.

liamstask:
marking as draft, will revisit after #337

jaraco:
Now that #317 is closed, I believe this one is tee'd up for a refresh. Please resolve the conflicts and take a look.

cheroot/server.py
```diff
@@ -1523,6 +1523,11 @@ class HTTPServer:
     timeout = 10
     """The timeout in seconds for accepted connections (default 10)."""
 
+    expiration_interval = 0.5
+    """The interval, in seconds, at which the server checks for
+    expired connections (default 0.5).
```
liamstask, line 1528: a bit arbitrary, but this matches the [socketserver default](https://docs.python.org/3/library/socketserver.html#socketserver.BaseServer.serve_forever)

liamstask:
ok, i believe this is ready for another look. this PR is not strictly needed if you're in the process of preparing a more major release, but this should address the issues from #305 in a more direct/robust way, per the notes in the PR description.

cheroot/connections.py
```diff
@@ -170,48 +171,57 @@ def expire(self):
             self._selector.unregister(sock_fd)
             conn.close()
 
-    def get_conn(self):
```
webknjaz, line 173: Removal of this generates a failure @ https://github.com/cherrypy/cheroot/pull/311/checks?check_run_id=1451730559 because it's linked from Sphinx. To avoid it, you could add it to the ignore list @ `docs/conf.py`. See https://github.com/cherrypy/cheroot/blob/c99085a/docs/conf.py#L104-L111.

cheroot/server.py
```diff
@@ -8,7 +8,7 @@
     server = HTTPServer(...)
     server.start()
     ->  while True:
-            tick()
+            serve()
```
webknjaz, line 11: This doesn't seem right. `while True`?

cheroot/server.py
```diff
@@ -2039,17 +2039,13 @@ def resolve_real_bind_addr(socket_):
...
-                # Just drop the conn. TODO: write 503 back?
-                conn.close()
-
-        self._connections.expire()
+    def process_conn(self, conn):
+        """Process an incoming HTTPConnection."""
+        try:
+            self.requests.put(conn)
+        except queue.Full:
+            # Just drop the conn. TODO: write 503 back?
```
webknjaz, line 2047: 503 sounds good

webknjaz:
@liamstask did you verify how this works with CherryPy? Maybe run its test suite?

cheroot/server.py
liamstask: this is just moving existing code - would just as soon avoid changing it further in this PR

cheroot/server.py
liamstask: updated - please take a look

liamstask:
updated, and tested successfully locally against cherrypy/cherrypy master by:
* editing the [cheroot-master testenv](https://github.com/cherrypy/cherrypy/blob/2a8aaccd649eb1011382c39f5cd93f76f980c0b1/tox.ini#L30) to point to this PR:  `deps = git+git://github.com/liamstask/cheroot.git@0c3fb0d597c492d891361cc6dcc003045ec37da6#egg=cheroot`
* running `python3 -m tox -e cheroot-master`

jaraco (approved):
Overall, this approach seems sane and in some ways simplifies the logic. Since it appears to be having the desired effect and tests appear to be passing, I'm okay proceeding with it.

cheroot/server.py
jaraco: This TODO is unchanged for this PR, so changing to 503 is probably out of scope for this change.

Issue #312, opened 2020-08-13
timed_out_connections - dictionary changed size during iteration


Safihre:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**
Traceback in the logs:
```python-traceback
Traceback (most recent call last):
  File "/opt/lib/python3.8/site-packages/cheroot/server.py", line 1795, in serve
    self.tick()
  File "/opt/lib/python3.8/site-packages/cheroot/server.py", line 2030, in tick
    self.connections.expire()
  File "/opt/lib/python3.8/site-packages/cheroot/connections.py", line 107, in expire
    for sock_fd, conn in timed_out_connections:
  File "/opt/lib/python3.8/site-packages/cheroot/connections.py", line 102, in <genexpr>
    timed_out_connections = (
  File "/opt/lib/python3.8/_collections_abc.py", line 743, in __iter__
RuntimeError: dictionary changed size during iteration
```

❓ **What is the motivation / use case for changing the behavior?**

💡 **To Reproduce**

💡 **Expected behavior**

📋 **Details**

📋 **Environment**

- Cheroot version: 8.4.4
- CherryPy version: 18.6.0 (if applicable)
- Python version: 3.8.X
- OS: NAS
- Browser: all

📋 **Additional context**

Safihre:
Reported by a user with fresh install. Gathering more info about installed CherryPy/cheroot versions.

Safihre:
@webknjaz I can now confirm this was introduced between 8.4.4 and 8.4.2, because I just updated `cheroot` on my own system and am also getting these problems now!

Safihre:
Must be #309.

webknjaz:
Yep, I think I know what the problem is.

webknjaz:
https://github.com/cherrypy/cheroot/pull/309/files#r469927485

webknjaz:
@liamstask ^

webknjaz:
If somebody gets to sending a PR, the fix should be changing `get_map()` to `get_map().copy()` in all places where it's used. I'll probably be too busy today at least but will happily merge somebody else's PR if it's ready.

webknjaz:
The fix has been released as v8.4.5: https://pypi.org/project/cheroot/8.4.5/.

Thanks to @liamstask @cyraxjoe for help!

[...]

PR #316, opened 2020-08-23
Regression tests for issue 312

cyraxjoe:
  - [ ] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [X] 📋 tests/coverage improvement
  - [ ] 📋 refactoring
  - [ ] 💥 other

#312

Errors that causes issues #312 weren't detected by the testsuite. 

Detect the errors that caused issues #312 

📋 **Other information**:

This PR implements the tests that are meant to be passed by PR #313, it expands the test coverage of the `coonections` module.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/316)

### DIFF
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -5,6 +5,9 @@ __metaclass__ = type
 
 import socket
 import time
+import logging
+import traceback as traceback_
+from collections import namedtuple
 
 from six.moves import range, http_client, urllib
 
@@ -108,7 +111,7 @@ class Controller(helper.Controller):
 
 
 @pytest.fixture
-def testing_server(wsgi_server_client):
+def testing_server(wsgi_server_client, monkeypatch):
     """Attach a WSGI app to the given server and preconfigure it."""
     app = Controller()
 
@@ -121,6 +124,27 @@ def testing_server(wsgi_server_client):
     wsgi_server.timeout = timeout
     wsgi_server.server_client = wsgi_server_client
     wsgi_server.keep_alive_conn_limit = 2
+
+    # patch the error_log calls of the server instance
+    class ErrorLog:
+        """Mock class to access the server error_log calls made by the server.
+        """
+        FuncCall = namedtuple('ErrorLogCall', ['msg', 'level', 'traceback'])
+
+        def __init__(self):
+            self.calls = []
+
+        def __call__(self, msg='', level=20, traceback=False):
+            if traceback:
+                tblines = traceback_.format_exc()
+            else:
+                tblines = ''
+            self.calls.append(
+                ErrorLog.FuncCall(msg, level, tblines)
+            )
+
+    monkeypatch.setattr(wsgi_server, 'error_log', ErrorLog())
+
     return wsgi_server
 
 
@@ -480,6 +504,17 @@ def test_keepalive_conn_management(test_client):
 
     # But the second one should still be valid.
     request(c2)
+
+    # Verify the server error log, just case something happened
+    # when expiring connections.
+    # In particular, verify that we didn't have any call with level
+    # logging.ERROR.
+    for c in test_client.server_instance.error_log.calls:
+        assert c.level != logging.ERROR, (
+            "Found error in the error log: \n"
+            "message = '{}'\n"
+            "{}".format(c.msg, c.traceback)
+        )
 
     # Restore original timeout.
     test_client.server_instance.timeout = timeout
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -133,8 +133,10 @@ def testing_server(wsgi_server_client, monkeypatch):
 
         def __init__(self):
             self.calls = []
+            # to be used the the teardown validation
+            self.ignored_msgs = []
 
-        def __call__(self, msg='', level=20, traceback=False):
+        def __call__(self, msg='', level=logging.INFO, traceback=False):
             if traceback:
                 tblines = traceback_.format_exc()
             else:
@@ -144,8 +146,17 @@ def testing_server(wsgi_server_client, monkeypatch):
             )
 
     monkeypatch.setattr(wsgi_server, 'error_log', ErrorLog())
-
-    return wsgi_server
+    yield wsgi_server
+    # Teardown verification, in case that the server logged an
+    # error that wasn't notified to the client or we just made a mistake.
+    for c in wsgi_server.error_log.calls:
+        if c.level > logging.WARNING:
+            if c.msg not in wsgi_server.error_log.ignored_msgs:
+                raise AssertionError(
+                    "Found error in the error log: "
+                    "message = '{}', level = '{}'\n"
+                    "{}".format(c.msg, c.level, c.traceback)
+                )
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


cheroot/test/test_conn.py
```diff
@@ -121,7 +125,39 @@ def _timeout(req, resp):
...
+            self.calls = []
+            # to be used the the teardown validation
+            self.ignored_msgs = []
+
+        def __call__(self, msg='', level=logging.INFO, traceback=False):
+            if traceback:
+                tblines = traceback_.format_exc()
+            else:
+                tblines = ''
+            self.calls.append(ErrorLog.FuncCall(msg, level, tblines))
```
webknjaz, line 147: Looks like this should go to the module level, no need to keep it inside the fixture function + there would be no need for noqa too.

cheroot/test/test_conn.py
```diff
@@ -121,7 +125,39 @@ def _timeout(req, resp):
...
+    # Teardown verification, in case that the server logged an
+    # error that wasn't notified to the client or we just made a mistake.
+    for c in wsgi_server.error_log.calls:
+        if c.level > logging.WARNING:
+            if c.msg not in wsgi_server.error_log.ignored_msgs:
+                raise AssertionError(
+                    'Found error in the error log: '
+                    "message = '{}', level = '{}'\n"
+                    '{}'.format(c.msg, c.level, c.traceback),
+                )
```
webknjaz, line 160: 1) this level of nesting is too deep & unnecessary
2) it's better to use actual `assert` statement instead of raising the error manually:
```suggestion
        if c.level <= logging.WARNING:
            continue

        assert c.msg in wsgi_server.error_log.ignored_msgs, (
            'Found error in the error log: '
            "message = '{c.msg!s}', level = '{c.level!s}'\n"
            '{c.traceback!s}'.format(**locals()),
        )
```

cheroot/test/test_conn.py
```diff
@@ -948,8 +984,13 @@ def test_Content_Length_out(
 
     assert actual_status == expected_resp_status
     assert actual_resp_body == expected_resp_body
-
     conn.close()
```
webknjaz, line 987: I'd prefer to keep the separator because close isn't strictly related to the assertions
```suggestion

    conn.close()

```

cheroot/test/test_conn.py
cyraxjoe: I personally prefer to keep these close to where they are relevant for test. Otherwise they are just _noise_ for the rest of the tests (ideally having _only_ tests at the top level of the module), but I must admit that this _philosophy_ comes from other projects (I believe I got this from pyramid) where there is a clear distinction in regular code standards and test code in terms of nesting functionality and imports inside tests for example .  This indeed becomes a problem conforming for pydoc style but having the same standard in the doctrings for tests that are never going to be parsed by any tool other than the eyes of the developer doesn't make a lot of sense to me, other than been dogmatic to conform to the linter. 

Having said that... if you _really_ want that I can make it  happen.

cheroot/test/test_conn.py
```diff
@@ -1000,3 +1041,73 @@ def test_No_CRLF(test_client, invalid_terminator):
     expected_resp_body = b'HTTP requires CRLF terminators'
     assert actual_resp_body == expected_resp_body
     conn.close()
+
+
+def test_invalid_selected_connection(test_client, monkeypatch):  # noqa: D200,D205,D400,E501
+    """Test the error handling segment of
+    ``cheroot.connections.ConnectionManager.get_conn``.
```
webknjaz, line 1048: I'm trying to maintain as low number of noqa as possible across the project:
```suggestion
def test_invalid_selected_connection(test_client, monkeypatch):
    """Test the error handling segment of HTTP connection selection.

    See ``cheroot.connections.ConnectionManager.get_conn``.
```

cheroot/test/test_conn.py
```diff
@@ -1000,3 +1041,73 @@ def test_No_CRLF(test_client, invalid_terminator):
...
+            self.socket_closed = False
+
+        def __call__(self):
+            if self.sabotage_conn:
+                for _, (*_, conn) in self.original_get_map().items():
+                    if isinstance(conn, cheroot.server.HTTPConnection):
+                        # close the socket to cause OSError
+                        conn.close()
+                        self.socket_closed = True
+            return self.original_get_map()
```
webknjaz, line 1076: I'd rather keep these on the module level since they don't seem to depend on test-local vars.

cheroot/test/test_conn.py
```diff
@@ -1000,3 +1041,73 @@ def test_No_CRLF(test_client, invalid_terminator):
...
+            self.original_select = original_select
+            self.request_served = False
+            self.os_error_triggered = False
+
+        def __call__(self, timeout):
+            if self.request_served:
+                self.os_error_triggered = True
+                raise OSError('Error while selecting the client socket.')
+            else:
+                return self.original_select(timeout)
```
webknjaz, line 1061: Let's simplify this
```suggestion

            return self.original_select(timeout)
```

cheroot/test/test_conn.py
```diff
@@ -1000,3 +1041,73 @@ def test_No_CRLF(test_client, invalid_terminator):
...
+            self.sabotage_conn = False
+            self.socket_closed = False
+
+        def __call__(self):
+            if self.sabotage_conn:
+                for _, (*_, conn) in self.original_get_map().items():
+                    if isinstance(conn, cheroot.server.HTTPConnection):
+                        # close the socket to cause OSError
+                        conn.close()
+                        self.socket_closed = True
```
webknjaz, line 1075: Let's deal with nesting a bit
```suggestion
            sabotage_targets = (
                conn for _, (*_, conn) in self.original_get_map().items()
                if isinstance(conn, cheroot.server.HTTPConnection)
            ) if self.sabotage_conn else ()

            for conn in sabotage_targets:
                # close the socket to cause OSError
                conn.close()
                self.socket_closed = True
```

cheroot/test/test_conn.py
webknjaz: This request doesn't only come from the linter violations. The true origin is that such things make noise inside the test itself making it harder to consume its contents. I usually prefer to have such things in fixtures so that the test only operates things on one level of abstraction.

webknjaz: Or in this case, it'd be another fixture dependency maybe...

cheroot/test/test_conn.py
cyraxjoe: I really don't like the alternative that you are suggesting. To me is just complicating things for the impression of linearity for a generator that is guaranteed that will yield one element. Is a functional pattern in a non-functional code-base.

But! If you _really_ want that, then ok.

cyraxjoe:
@webknjaz I think I implemented all of the comments. Just a note on using `assert` instead of rising the exception.. it seems that the errors messages in the test run are less readable, the traceback is not fully shown. Not sure if that is configurable by the pytest invocation.

Compare this run https://github.com/cherrypy/cheroot/runs/1018974455  vs https://github.com/cherrypy/cheroot/runs/1019226287

cheroot/test/test_conn.py
```diff
@@ -1000,3 +1062,88 @@ def test_No_CRLF(test_client, invalid_terminator):
...
+            conn for _, (*_, conn) in self.original_get_map().items()
+            if isinstance(conn, cheroot.server.HTTPConnection)
+        ) if self.sabotage_conn else ()
+
+        for conn in sabotage_targets:
+            # close the socket to cause OSError
+            conn.close()
+            self.socket_closed = True
+
+        return self.original_get_map()
```
webknjaz, line 1106: I'm a bit worried about `self.original_get_map()` being called twice. Can it return different values for each invocation? Should it be called once and the result reused?

webknjaz:
> the traceback is not fully shown. Not sure if that is configurable by the pytest invocation.

I think you can make it more verbose with something like `-vvvvv --full-trace`

webknjaz:
I noticed some Python 2 issues after the merge, trying to address them in master.

Issue #317, opened 2020-08-24
dictionary changed size during iteration


AndreLouisCaron:
Hi there!

I've just upgraded from `cheroot==8.4.4` to `cheroot==8.4.5` and I noticed what looks like a regression.

❓ **I'm submitting a ...**
- [X] 🐞 bug report

🐞 **Describe the bug. What is the current behavior?**

I see this in my server logs:

```
Error in HTTPServer.tick
Traceback (most recent call last):
  File ".../lib/site-packages/cheroot/server.py", line 1798, in serve
    self.tick()
  File ".../lib/site-packages/cheroot/server.py", line 2046, in tick
    self._connections.expire()
  File ".../lib/site-packages/cheroot/connections.py", line 105, in expire
    in self._selector.get_map().items()
  File ".../lib/_abcoll.py", line 414, in items
    return [(key, self[key]) for key in self]
RuntimeError: dictionary changed size during iteration
```

❓ **What is the motivation / use case for changing the behavior?**

There is occasional log spam (for a low traffic environment), but I assume this has some other undesired side-effects...

💡 **To Reproduce**

I'm not sure exactly.  We have a test environment with synthetic checks that does a couple requests per second and the behaviour occurs "naturally"...

📋 **Details**

We very recently upgraded from version 8.4.4 to 8.4.5, but it's possible the error was already present in 8.4.4.  We use `pip-compile` and `cheroot` is not pinned, so we usually get updates shortly after they are released.

📋 **Environment**
- Cheroot version: 8.4.5
- CherryPy version: 17.4.2 (I don't think we really use it)
- Python version: 2.7
- OS: Windows
- Browser: requests

cyraxjoe:
:disappointed: I think that this issue is bound to happen unless we use a lock or we _really_ have a perfect management of the underlying selectors. But given how I saw that the `selectors` module implemented the mapping we have to manage that contention ourselves.

This situation could have happen if concurrently a [worker thread is putting a connection in a thread](https://github.com/cherrypy/cheroot/blob/master/cheroot/workers/threadpool.py#L123) and the [server is expiring connections.](https://github.com/cherrypy/cheroot/blob/master/cheroot/server.py#L2046)

1. WorkerThread -> returning a connection to be managed because is using `keep-alive`.
2. ServerThread -> expiring connections, exactly at the time when is iterating the mapping.

[...]

PR #318, opened 2020-08-27
Implement contention via lock for the selectors in ConnectionManager.

cyraxjoe:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [x] 📋 refactoring
  - [ ] 💥 other

#317 

There is a race condition between the internal server `tick()` method to expire connections and the
worker threads that can modify the underlying sockets to be selected, while is been iterating over
by the expire method of `ConnectionManager`.

Limits the access methods that could  modify the underlying sockets/fd that are been selected by the
`selectors.DefaultSelector` by having a write lock. An indirection layer that will guarantee consistency
between thread at the cost of a little overhead.

📋 **Other information**:

This locking approach could be temporary in case that we refactor the whole use of `selectors`, `ConnectionManager`
and `Server`.

Additionally: 
 - this PR expands the use of the `ErrorLogManager` helper test class, now is also used by `test_wsgi`, this test module
seems to be extensive enough to trigger some of the errors that were reported in issue #60 
- remove the `xfail` mark in one test that seems that it wasn't failing after all, it just wasn't returning the right value the test wsgi app.
- remove the py2 exception to tolerate the serve error_log messages, it was covering an issue in the different ways that `Server` closes the connections, for py2 it closed directly the OS socket, given the comment that was in that segment (from around 12 years)  seems that the goal was to send the FIN packet, which can be accomplished by socket.shutdown, instead of relying in the OS by closing the file descriptor. (verified with linux/wireshark/pdb at localhost).

📋 **Checklist**:

  - [X] Integration tests for the changes exist (if applicable)

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/318)

### DIFF
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -1013,15 +1013,10 @@ def test_No_CRLF(test_client, invalid_terminator):
 
 
 class FaultySelectorManager:
-    """Mock class to insert errors in th execution.
-
-    Wraps: :py:class:`!cheroot.connections._SelectorManager`.
-    """
+    """Mock class to insert errors in th execution."""
 
     def __init__(self, orig_select_mgr):
         """Prepare the wrapper to intervene the execution.
-
-        Wraps: :py:class:`!cheroot.connections._SelectorManager`.
 
         Initialize the flags to trigger the errors inside the
         _SelectorManager to put to test the error handling segments
@@ -1051,7 +1046,7 @@ class FaultySelectorManager:
         return len(self.orig_select_mgr)
 
     def __iter__(self):
-        """Intercept the calls to _SelectorManager iterator."""
+        """Intercept the calls to the iterator."""
         result = tuple(self.orig_select_mgr)
         sabotage_targets = (
             conn for _, _, _, conn in result
@@ -1067,7 +1062,7 @@ class FaultySelectorManager:
         return iter(result)
 
     def select(self, timeout):
-        """Intercept the calls to `:py:meth:!_SelectorManager.select`."""
+        """Intercept the select call."""
         if self.request_served:
             self.os_error_triggered = True
             raise OSError('Error while selecting the client socket.')
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -1013,17 +1013,22 @@ def test_No_CRLF(test_client, invalid_terminator):
 
 
 class FaultySelectorManager:
-    """Mock class to insert errors in a `_SelectorManager`."""
+    """Mock class to insert errors in th execution.
+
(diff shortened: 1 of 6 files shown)


cheroot/test/test_conn.py
```diff
@@ -1013,16 +1013,11 @@ def test_No_CRLF(test_client, invalid_terminator):
 
 
 class FaultySelectorManager:
-    """Mock class to insert errors in th execution.
-
-    Wraps: :py:class:`!cheroot.connections._SelectorManager`.
```
webknjaz: @cyraxjoe FYI look how I've ignored similar references to undocumented objects in the past: https://github.com/cherrypy/cheroot/blob/d2f28f2/docs/conf.py#L95-L96

cheroot/test/test_conn.py
```diff
@@ -72,7 +68,7 @@ def err_before_read(req, resp):
 
     def one_megabyte_of_a(req, resp):
         """Render 1MB response."""
-        return ['a' * 1024] * 1024
+        return 'a' * 1024 * 1024
```
webknjaz, line 71: Why did you change a series of chunks with one blob?

cheroot/test/test_conn.py
```diff
@@ -1017,12 +963,11 @@ def test_Content_Length_out(
     ))
 
 
-@pytest.mark.xfail(
```
webknjaz, line 1020: @cyraxjoe this was failing but flakily. I was unable to reproduce this on every run.

webknjaz:
@cyraxjoe I want to explore #319 for solving this first. Do you think you could extract tests into a separate PR so that they could be merged earlier? Plus, it'd make this one easier to review.

cheroot/test/test_conn.py
cyraxjoe: I think that I did this because of the way that the  helper `Controller` is implemented https://github.com/cherrypy/cheroot/blob/master/cheroot/test/helper.py#L159-L169, it doesn't calculate the right size for the response if the handler didn't specified the `Content-Length` header. I remember that I was looking over the traffic with wireshark and saw that the response was there but the application only read the first 1024 bytes because of the `Content-Length` header, the rest was discarded. I'll double check then, I think my "fix" was only fixing the integration with testing code.

cyraxjoe:
@webknjaz  I think that only the `ErrorLogManitor` refactor could be helpful in the other PR (to include the verification in the test  in `test_wsgi.py`). The other test is _heavily_ dependent on the final implementation, in this case is tightly coupled with the `_SelectorManager`, @liamstask  would need to verify the triggers of the errors paths to trigger those and verify the output in the coverage report.

PR #319, opened 2020-08-31
connections: put conns safely

### LINKED ISSUE #317: dictionary changed size during iteration

liamstask:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

Fixes #317

seeing errors while iterating over the map of connections within the selector in the event that a `put()` call overlaps an `expire()` call

instead of registering sockets with the selector from other threads, ensure they're registered from within the thread running the selector.

📋 **Other information**:

This introduces https://pypi.org/project/backports.socketpair in order to provide a pure-python implementation of `socket.socketpair()` for platforms that don't have have it, prior to python 3.5+.

We could possibly use some of the enhanced tests from https://github.com/cherrypy/cheroot/pull/318 to validate

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/319)

### DIFF
--- a/cheroot/_compat.py
+++ b/cheroot/_compat.py
@@ -15,6 +15,14 @@ except ImportError:
     import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
 try:
+    from backports.socketpair import socket
+except ImportError:
+    import socket
+finally:
+    socketpair = socket.socketpair
+    del socket
+
+try:
     import ssl
     IS_ABOVE_OPENSSL10 = ssl.OPENSSL_VERSION_INFO >= (1, 1)
     del ssl
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -74,6 +74,17 @@ class ConnectionManager:
         self._put_q = collections.deque()
         self._selector = selectors.DefaultSelector()
 
+        # _num_conns tracks the number of HTTPConnection objects
+        # being managed, which may be in one of:
+        #
+        # * _readable_conns
+        # * the selector
+        # * _put_q
+        #
+        # it is tracked independently of those containers to ensure a stable
+        # count, even as conns are transferred between them
+        self._num_conns = 0
+
         self._selector.register(
             server.socket.fileno(),
             selectors.EVENT_READ, data=server,
@@ -87,6 +98,16 @@ class ConnectionManager:
         # protects writes to self._ctrl_tx
         self._ctrl_lock = threading.Lock()
 
+    def _pop_readable_conn(self):
+        conn = self._readable_conns.popleft()
+        self._num_conns -= 1
+        return conn
+
+    def _remove_conn(self, sock_fd, conn):
+        self._selector.unregister(sock_fd)
+        conn.close()
+        self._num_conns -= 1
+
     def put(self, conn):
         """Put idle connection into the ConnectionManager to be managed.
 
@@ -94,9 +115,11 @@ class ConnectionManager:
             conn (cheroot.server.HTTPConnection): HTTP connection
                 to be managed.
         """
+        self._num_conns += 1
         conn.last_used = time.time()
-        # if this conn doesn't have any more data waiting to be read,
-        # register it with the selector.
+
+        # if this conn has more data waiting to be read,
+        # store it in the readable queue.
         if conn.rfile.has_data():
             self._readable_conns.append(conn)
             return
@@ -129,8 +152,7 @@ class ConnectionManager:
             if conn.last_used < threshold
         ]
         for sock_fd, conn in timed_out_connections:
-            self._selector.unregister(sock_fd)
-            conn.close()
+            self._remove_conn(sock_fd, conn)
 
     def get_conn(self):
         """Return a HTTPConnection object which is ready to be handled.
@@ -148,7 +170,7 @@ class ConnectionManager:
         """
         # return a readable connection if any exist
         with suppress(IndexError):
-            return self._readable_conns.popleft()
+            return self._pop_readable_conn()
 
         # Will require a select call.
         try:
@@ -180,7 +202,7 @@ class ConnectionManager:
             self._readable_conns.append(conn)
 
         try:
-            return self._readable_conns.popleft()
+            return self._pop_readable_conn()
         except IndexError:
             return None
 
@@ -197,16 +219,15 @@ class ConnectionManager:
         # Mark any connection which no longer appears valid
         # If the server or ctrl sockets are invalid,
         # we'll just shutdown.
-        invalid_keys = []
+        invalid_conns = []
         for sock_fd, conn in self._get_selector_conns():
             try:
                 os.fstat(sock_fd)
             except OSError:
-                invalid_keys.append((sock_fd, conn))
+                invalid_conns.append((sock_fd, conn))
 
-        for sock_fd, conn in invalid_keys:
-            self._selector.unregister(sock_fd)
-            conn.close()
+        for sock_fd, conn in invalid_conns:
+            self._remove_conn(sock_fd, conn)
 
     def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
         try:
@@ -313,23 +334,7 @@ class ConnectionManager:
         self._selector.close()
 
     @property
-    def _num_connections(self):
-        """Return the current number of connections.
-
-        Includes any in the readable list, the put queue,
-        or registered with the selector,
-        minus two for the server socket and control socket,
-        which are always registered with the selector.
-        """
-        return (
-            len(self._readable_conns) +
-            len(self._put_q) +
-            len(self._selector.get_map())
-            - 2
-        )
-
-    @property
     def can_add_keepalive_connection(self):
         """Flag whether it is allowed to add a new keep-alive connection."""
         ka_limit = self.server.keep_alive_conn_limit
-        return ka_limit is None or self._num_connections < ka_limit
+        return ka_limit is None or self._num_conns < ka_limit
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -7,10 +7,12 @@ import collections
 import io
 import os
 import socket
+import threading
 import time
 
 from . import errors
 from ._compat import selectors
+from ._compat import socketpair
 from ._compat import suppress
 from .makefile import MakeFile
 
@@ -55,6 +57,11 @@ class ConnectionManager:
     This is for connections which are being kept-alive for follow-up requests.
     """
 
+    # NOTE: The selector must only be accessed within the thread
+    # running server.serve().
+
+    _CTRL_MSG_PUT = b'P'
+
     def __init__(self, server):
         """Initialize ConnectionManager object.
 
@@ -64,12 +71,21 @@ class ConnectionManager:
         """
         self.server = server
         self._readable_conns = collections.deque()
+        self._put_q = collections.deque()
         self._selector = selectors.DefaultSelector()
 
         self._selector.register(
             server.socket.fileno(),
             selectors.EVENT_READ, data=server,
         )
+
+        self._ctrl_rx, self._ctrl_tx = socketpair()
+        self._selector.register(
+            self._ctrl_rx.fileno(),
+            selectors.EVENT_READ, data=self._ctrl_rx,
+        )
+        # protects writes to self._ctrl_tx
+        self._ctrl_lock = threading.Lock()
 
     def put(self, conn):
         """Put idle connection into the ConnectionManager to be managed.
@@ -83,10 +99,18 @@ class ConnectionManager:
         # register it with the selector.
         if conn.rfile.has_data():
             self._readable_conns.append(conn)
-        else:
-            self._selector.register(
-                conn.socket.fileno(), selectors.EVENT_READ, data=conn,
-            )
+            return
+
+        # otherwise, register it with the selector.
+        self._put_q.append(conn)
+        with self._ctrl_lock:
+            self._ctrl_tx.send(self._CTRL_MSG_PUT)
+
+    def _get_selector_conns(self):
+        """Retrieve client connections registered with the selector."""
+        for _, (_, sock_fd, _, conn) in self._selector.get_map().items():
+            if conn not in {self.server, self._ctrl_rx}:
+                yield (sock_fd, conn)
 
     def expire(self):
         """Expire least recently used connections.
@@ -101,15 +125,14 @@ class ConnectionManager:
         threshold = time.time() - self.server.timeout
         timed_out_connections = [
             (sock_fd, conn)
-            for _, (_, sock_fd, _, conn)
-            in self._selector.get_map().items()
-            if conn != self.server and conn.last_used < threshold
+            for (sock_fd, conn) in self._get_selector_conns()
+            if conn.last_used < threshold
         ]
         for sock_fd, conn in timed_out_connections:
             self._selector.unregister(sock_fd)
             conn.close()
 
-    def get_conn(self):  # noqa: C901  # FIXME
+    def get_conn(self):
         """Return a HTTPConnection object which is ready to be handled.
 
         A connection returned by this method should be ready for a worker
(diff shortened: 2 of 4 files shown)


cheroot/connections.py
```diff
@@ -137,27 +155,15 @@ def get_conn(self):  # noqa: C901  # FIXME
                 in self._selector.select(timeout=0.01)
             ]
         except OSError:
-            # Mark any connection which no longer appears valid
-            invalid_entries = []
```
liamstask, line 141: factoring this out with no changes in order to address the `noqa` FIXME above

cheroot/server.py
```diff
@@ -1804,6 +1804,7 @@ def serve(self):
                     traceback=True,
                 )
 
+        self._connections.close()
```
liamstask, line 1807: moving this call inside the thread that is running the selector to ensure the selector is only ever accessed from a single thread

cheroot/connections.py
```diff
@@ -16,6 +17,11 @@
 
 import six
 
+try:
+    import backports.socketpair  # noqa: F401  # lgtm [py/unused-import]
```
liamstask, line 21: this is not as simple to import through `_compat` since it adds the function directly to the `socket` module - [link](https://github.com/mhils/backports.socketpair/blob/f834f594cc89481995f7e532138789996c55b47b/backports/socketpair/__init__.py#L56) - so an `import as` statement doesn't quite work.

let me know how you'd like to manage this.

cheroot/_compat.py
```diff
@@ -14,6 +14,14 @@
 except ImportError:
     import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
+try:
+    import socket  # noqa: F401  # lgtm [py/unused-import]
```
webknjaz, line 18: It's better to have just one instruction in try/except block + this import triggers two LGTM violations, not just one:
```suggestion
import socket  # noqa: F401  # lgtm [py/unused-import] lgtm [py/import-and-import-from]
try:
```

cheroot/connections.py
```diff
@@ -84,9 +100,15 @@ def put(self, conn):
         if conn.rfile.has_data():
             self._readable_conns.append(conn)
         else:
-            self._selector.register(
-                conn.socket.fileno(), selectors.EVENT_READ, data=conn,
-            )
+            self._put_q.append(conn)
+            with self._ctrl_lock:
+                self._ctrl_tx.send(self._CTRL_MSG_PUT)
```
webknjaz, line 105: Let's dedent this a bit
```suggestion
            return

        # if this conn doesn't have any more data waiting to be read,
        # register it with the selector.
        self._put_q.append(conn)
        with self._ctrl_lock:
            self._ctrl_tx.send(self._CTRL_MSG_PUT)
```

cheroot/connections.py
```diff
@@ -84,9 +100,15 @@ def put(self, conn):
...
+
+    def _get_selector_conns(self):
+        # helper to retrieve conns registered with the selector
```
webknjaz, line 108: I think this goes better as a docstring
```suggestion
        """Retrieve connections registered with the selector."""
```

cheroot/connections.py
```diff
@@ -84,9 +100,15 @@ def put(self, conn):
...
+        for _, (_, sock_fd, _, conn) in self._selector.get_map().items():
+            if conn != self.server and conn != self._ctrl_rx:
```
webknjaz, line 110: how about
```suggestion
            if conn not in {self.server, self._ctrl_rx}:
```

webknjaz:
@cyraxjoe I'd like to hear your take on this PR

@liamstask oh, and looks like this needs rebasing, I've added a few ignores for those cryptography warnings in master

cheroot/_compat.py
liamstask: i am not totally sure on this - i *think* it may be needed because `backports.socketpair` modifies the `socket` module, and those modifications wouldn't be visible outside that module unless we imported it here first, but i am not familiar enough w the python import mechanism, and don't have an affected platform to test on directly. suggestions welcome :)

webknjaz: Have you tested whether it's different with this line removed?

cheroot/connections.py
```diff
@@ -84,9 +100,15 @@ def put(self, conn):
...
-            self._selector.register(
-                conn.socket.fileno(), selectors.EVENT_READ, data=conn,
-            )
+            self._put_q.append(conn)
+            with self._ctrl_lock:
+                self._ctrl_tx.send(self._CTRL_MSG_PUT)
+
+    def _get_selector_conns(self):
+        # helper to retrieve conns registered with the selector
+        for _, (_, sock_fd, _, conn) in self._selector.get_map().items():
```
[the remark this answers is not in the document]
liamstask: the strategy in this PR is to ensure that only a single thread is ever touching the selector, so from a thread safety perspective it should not be necessary. do you have another safety case in mind?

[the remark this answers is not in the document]
webknjaz: Because this is in a separate method, the callers would need to account for the implementation details. Hence this is unsafe and will probably introduce future bugs if somebody will attempt to refactor the calling code (turn two loops into one) w/o having the context.

cheroot/_compat.py
liamstask: not yet - i don't have an affected platform to test on, and the platforms in ci that would exercise that code were unresponsive prior to rebasing to get some of those ssl fixes. i can push a couple experiments.

liamstask:
rebased

cheroot/_compat.py
webknjaz: Ah, so only did it because of the example in their readme? I'm pretty sure that it's unnecessary.

cheroot/connections.py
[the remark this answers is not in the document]
liamstask: given that this is marked as a 'private' helper, i'd just as soon indicate in the docstring that this is returning references and callers should not store them, etc. if you feel strongly, it should be ok to copy them, but this is mainly intended just to provide a filter on the iteration of the map items.

cheroot/_compat.py
```diff
@@ -14,6 +14,14 @@
 except ImportError:
     import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
+import socket  # noqa: F401, E501  # lgtm [py/unused-import] lgtm [py/import-and-import-from]
```
webknjaz, line 17: They just monkey-patch the socket module: https://github.com/mhils/backports.socketpair/blob/master/backports/socketpair/__init__.py#L56

It happens to the globally cached module object, the import doesn't change anything

```suggestion
```

cheroot/connections.py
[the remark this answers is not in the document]
webknjaz: I understand this, I'm mostly thinking about preventing future regressions caused by the human factor.

cheroot/server.py
webknjaz: This probably needs a code comment

cheroot/_compat.py
```diff
@@ -14,6 +14,13 @@
 except ImportError:
     import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-import]
 
+try:
+    import backports.socketpair  # noqa: F401  # lgtm [py/unused-import]
+except ImportError:
+    pass
+
+from socket import socketpair  # noqa: F401  # lgtm [py/unused-import]
```
webknjaz, line 22: I think this version should work even better:
```suggestion
try:
    from backports.socketpair import socket
except ImportError:
    import socket
finally:
    socketpair = socket.socketpair
    del socket
```

liamstask:
looks like some sporadic failures on windows at the moment - will need to look more closely to see what's going on.

webknjaz:
Looks like `test_keepalive_conn_management` became flakier.

cheroot/connections.py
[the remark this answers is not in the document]
cyraxjoe: @webknjaz about the idea of copying the`get_map` return value, the internal map is not a dictionary, but rather is generated by making `__getitem__` calls, there is no `copy()` method (and the copy module will basically iterate over the content). The class of the object that is returned as a map is https://github.com/python/cpython/blob/master/Lib/selectors.py#L61-L78 which if you see the implementation of the `Mapping` abstract class, you'll see that the `get_map().items()` will end up in https://github.com/python/cpython/blob/master/Lib/_collections_abc.py#L764, and it works by iterating the keys and making the `__getitem__` method call. 

The alternative is to use the private dictionaries of that class, but I think that it would be a very bad idea to break that _interface contract_, that's why I preferred to manage an external set of key instead of having the overhead of iterating all the keys to make the copy (register/unregister returns the key) and the additional method calling, plus a lock to make sure we have a copy that can be iterated over in a threaded context.

liamstask:
@cyraxjoe thanks for taking a look.

> Adding a new dependency for not a good enough reason (internal communication, within the same instance).

given that this is a backport, it seems not too controversial, but i'll defer to the project maintainers.

> Mixing the file descriptors to be selected just for the purpose of internal communication (why not just a queue.Queue?)

a queue is not useful in this case because there's no execution context to wait on it - we're already waiting on the selector, so sending a message on a control socket is a standard way to wake it up.

> Delaying the addition of keep alive connections from the worker until get_conn is called, in effect we lost one "tick" that could be used to handle an event in the socket.
> By delaying the addition of the connections we are introducing a side effect in get_conn, it doesn't really just gets you a connections, its more like update the internal state and then gives you a connection.

i agree, but also think the fact that the server is driven by a `tick()` is undesirable in the first place - i've removed this in https://github.com/cherrypy/cheroot/pull/311 in favor of processing sockets as they become active in the selector rather than needlessly polling via `tick()`. `get_conn()` is removed altogether in this strategy.

> To manage the new additional communication the class now has to keep a connection counter (and make sure we keep this in mind for the future), a new pair of sockets, one of them added to the internal fd to be selected, a lock and a new deque, all of these loosely coupled in the implementation, which indeed is prone to errors when the next person comes along. I would advice to factor those out in another class, in case that you really are convinced with this approach.

definitely open to suggestions here - we could conceivably break out the concept of the control socket to a separate class, which would include both control sockets and the lock.

> We open the gate on the idea of using something in the very core of cheroot for message passing, later someone could come along an add an additional special message to be included and then carefully excluded, because we are no longer sure if this is a real connection or internal communication.

i'm not sure i understand this point - we always know whether it's internal or external comms based on the socket it arrived on, same as we distinguish the server socket from other sockets today.

cyraxjoe:
@liamstask  addressing some of your comments:

> > Mixing the file descriptors to be selected just for the purpose of internal communication (why not just a queue.Queue?)
> 
> a queue is not useful in this case because there's no execution context to wait on it - we're already waiting on the selector, so sending a message on a control socket is a standard way to wake it up.
> 
From my perspective, the instance of the `ConnectionManager` class is the shared execution context. It controls all the calls that it receives from any thread and shares the state in `self`.

For example:
 1. Adding into a `self.expire_queue.put_nowait(conn)` from the `ConnectionManager.put` method
 2. Have a new method to be called by `thick()` like `self.connections.sync_keep_alive_conns()`, before the `get_conn()` method call.
 3. The `sync_keep_alive_conn` consumes and registers the connections.
 4. The `get_conn` works without any additional verification.

> > Delaying the addition of keep alive connections from the worker until get_conn is called, in effect we lost one "tick" that could be used to handle an event in the socket.
> > By delaying the addition of the connections we are introducing a side effect in get_conn, it doesn't really just gets you a connections, its more like update the internal state and then gives you a connection.
> 
> i agree, but also think the fact that the server is driven by a `tick()` is undesirable in the first place - i've removed this in #311 in favor of processing sockets as they become active in the selector rather than needlessly polling via `tick()`. `get_conn()` is removed altogether in this strategy.

This is an architectural decision, not a matter of handling connections. It does changes the fundamentals on how the server works (also make sure you update those docstrings that refers to the workings of the server, the one in `server.py` is already outdated given the recent changes). Before considering the PR  #311 we should have as a priority to stabilize cheroot (currently we have broken cheroot that emits errors messages at the first connection it receives), have better test coverage of `connections.py` and `server.py` (specially of the error handling paths), some sort of basic performance benchmark to compare and then consider the architecture of `Server` in conjunction with the `ConnectionManager`. I'm not in favor of the current architecture, I'm in favor of _stability_, specially in the very core of cheroot.

> > We open the gate on the idea of using something in the very core of cheroot for message passing, later someone could come along an add an additional special message to be included and then carefully excluded, because we are no longer sure if this is a real connection or internal communication.
> 
> i'm not sure i understand this point - we always know whether it's internal or external comms based on the socket it arrived on, same as we distinguish the server socket from other sockets today.

I mean to open the _conceptual_ pattern that is laid out, I'm not saying that is technically backwards or that you are not using the APIs as they should, I'm saying that from my perspective, they are not required in this case, it seems overkill and mixes the conceptual structure of what are we selecting on. The `_selector` property became a message bus, not a client/server connection manager.

We are no longer only selecting in sockets dedicated for the purpose of managing the server and client connections, but to pass messages between the server and worker threads via the `ConnectionManager` methods, is not even used outside the same instance, if it was used from a loosely integrated object running in another thread (like a plugin) it may make more sense.

liamstask:
> From my perspective, the instance of the ConnectionManager class is the shared execution context. It controls all the calls that it receives from any thread and shares the state in self.

I'm referring more specifically to a thread of execution - there is one, and it is responsible for waiting on the selector & processing events as they occur.

> Adding into a self.expire_queue.put_nowait(conn) from the ConnectionManager.put method
> Have a new method to be called by tick() like self.connections.sync_keep_alive_conns(), before the get_conn() method call.
> The sync_keep_alive_conn consumes and registers the connections.
> The get_conn works without any additional verification.

This is undesirable because the next `tick()` call is already delayed by `_selector.select()` which is blocking. The control msg is required in order to wake up the select call, to reduce this delay (and ensure that the timeout specified to `select()` does not limit performance, as was the case in https://github.com/cherrypy/cheroot/issues/305).

> This is an architectural decision, not a matter of handling connections. It does changes the fundamentals on how the server works (also make sure you update those docstrings that refers to the workings of the server, the one in server.py is already outdated given the recent changes). Before considering the PR #311 we should have as a priority to stabilize cheroot (currently we have broken cheroot that emits errors messages at the first connection it receives), have better test coverage of connections.py and server.py (specially of the error handling paths), some sort of basic performance benchmark to compare and then consider the architecture of Server in conjunction with the ConnectionManager.

I would propose that the control flow for connection handling is an architectural decision - it informs whether the system is polling or event driven.

> I'm not in favor of the current architecture, I'm in favor of stability, specially in the very core of cheroot.

These are separate concerns - the proposed change is not inherently unstable. I agree stability should be improved.

> I mean to open the conceptual pattern that is laid out, I'm not saying that is technically backwards or that you are not using the APIs as they should, I'm saying that from my perspective, they are not required in this case, it seems overkill and mixes the conceptual structure of what are we selecting on. The _selector property became a message bus, not a client/server connection manager.

That is a fair perspective, i just think it is not aligned with the fact that this is a well understood pattern, as evidenced by the presence of APIs like [eventfd](https://linux.die.net/man/2/eventfd) and [signalfd](https://man7.org/linux/man-pages/man2/signalfd.2.html) which address this specific issue. Those happen to be Linux-specific and as such are not cross-platform, but this takes a similar approach.

Again, if the maintainers prefer another approach, that's totally fine - I think this one is preferable since it reduces contention (ie within `put()`,  the only part of the api that is called from other threads) and means data is not copied unnecessarily.

cyraxjoe:
I see... thank you for your feedback @liamstask, ok then! I suppose is up to @webknjaz or @jaraco, is been a while since I've been actively involved with the project, that I believe I no longer have enough _street cred_ to take those decision, it has to be made by any of those two.

My involvement in all of these was just due to the recent unstable releases. :sweat_smile:

cheroot/connections.py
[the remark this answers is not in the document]
liamstask: i agree that if the additional locking & copying are acceptable, the approach in https://github.com/cherrypy/cheroot/pull/318 from @cyraxjoe makes sense to me. this PR is predicated on the assumption that those are undesirable.

given the request to copy these to account for developer error in the future, it sounds like that might be preferable? either way is ok by me.

liamstask:
closing this out in favor of a simpler approach

PR #325, opened 2020-09-14
connections: add threadsafe selector wrapper

liamstask:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#317

The selector can be modified from thread pool threads, so can 

Adds a thread safe wrapper for the selector.

This is an alternative to #319, and is inspired by #318 but does not go as far as copying all connections, and instead only locks around access to the selector. This avoids adding a new dependency, and the implementation is simpler.

I still think #318 is also a reasonable approach, but it's not necessary to copy the connections, since the fundamental error is accessing the selector in different threads, as opposed to the connections themselves.

📋 **Other information**:

This also modifies the server to use the stdlib logger, such that we can use the pytest caplog fixture to capture and assert on the validity of log messages, which replaces the existing `ErrorLogMonitor` monkeypatch.

This also increases the scope in which these teardown verifications are done - previously it was only within test_conn, but this now effectively covers all tests.

I am open to keeping the existing `ErrorLogMonitor` implementation if that's preferable, but my assumption was that using the standard logger and pytest caplog infra, rather than a custom implementation, were worth it.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/325)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -220,7 +220,12 @@ class ConnectionManager:
 
         for sock_fd, conn in invalid_conns:
             self._selector.unregister(sock_fd)
-            conn.close()
+            # One of the reason on why a socket could cause an error
+            # is that the socket is already closed, ignore the
+            # socket error if we try to close it at this point.
+            # This is equivalent to OSError in Py3
+            with suppress(socket.error):
+                conn.close()
 
     def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
         try:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -7,6 +7,7 @@ import collections
 import io
 import os
 import socket
+import threading
 import time
 
 from . import errors
@@ -49,6 +50,58 @@ else:
         fcntl.fcntl(fd, fcntl.F_SETFD, old_flags | fcntl.FD_CLOEXEC)
 
 
+class _ThreadsafeSelector:
+    """Threadsafe wrapper around a DefaultSelector.
+
+    The expected read/write patterns are:
+    * iter: selector thread
+    * register: selector thread and threadpool, via put()
+    * unregister: selector thread only
+
+    Notably, this means ConnectionManager never needs to worry
+    that connections will be removed behind its back.
+
+    The lock is held when iterating or modifying the selector,
+    but is not required when select()ing on it.
+    """
+
+    def __init__(self):
+        self._selector = selectors.DefaultSelector()
+        self._lock = threading.Lock()
+
+    def __len__(self):
+        return len(self._selector.get_map())
+
+    @property
+    def connections(self):
+        """Retrieve connections registered with the selector."""
+        with self._lock:
+            for _, (_, sock_fd, _, conn) in self._selector.get_map().items():
+                yield (sock_fd, conn)
+
+    def register(self, fileobj, events, data=None):
+        with self._lock:
+            return self._selector.register(fileobj, events, data)
+
+    def unregister(self, fileobj):
+        with self._lock:
+            return self._selector.unregister(fileobj)
+
+    def select(self, timeout=None):
+        """Wrap the selectors.select call.
+
+        Returns entries ready to read in the form:
+            (socket_file_descriptor, connection)
+        """
+        return (
+            (key.fd, key.data)
+            for key, _ in self._selector.select(timeout=timeout)
+        )
+
+    def close(self):
+        self._selector.close()
+
+
 class ConnectionManager:
     """Class which manages HTTPConnection objects.
 
@@ -64,7 +117,7 @@ class ConnectionManager:
         """
         self.server = server
         self._readable_conns = collections.deque()
-        self._selector = selectors.DefaultSelector()
+        self._selector = _ThreadsafeSelector()
 
         self._selector.register(
             server.socket.fileno(),
@@ -101,15 +154,14 @@ class ConnectionManager:
         threshold = time.time() - self.server.timeout
         timed_out_connections = [
             (sock_fd, conn)
-            for _, (_, sock_fd, _, conn)
-            in self._selector.get_map().items()
+            for (sock_fd, conn) in self._selector.connections
             if conn != self.server and conn.last_used < threshold
         ]
         for sock_fd, conn in timed_out_connections:
             self._selector.unregister(sock_fd)
             conn.close()
 
-    def get_conn(self):  # noqa: C901  # FIXME
+    def get_conn(self):
         """Return a HTTPConnection object which is ready to be handled.
 
         A connection returned by this method should be ready for a worker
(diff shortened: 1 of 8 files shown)


cheroot/test/conftest.py
```diff
@@ -12,21 +12,40 @@
...
+    check_log_messages,
+    cheroot_server,
+    get_server_client,
 )
-from ..testing import get_server_client
 
 
 @pytest.fixture
-def wsgi_server_client(wsgi_server):  # noqa: F811
+def wsgi_server(caplog):
```
liamstask, line 26: these were both moved from `testing.py` in order to resolve the `F811` lint error. they were previously `scope=module` - i removed that in order to ensure that they're brought up and torn down cleanly for each test.

cheroot/test/test_core.py
```diff
@@ -283,7 +279,6 @@ def test_content_length_required(test_client):
 
 
 @pytest.mark.xfail(
-    IS_CIRCLE_CI_ENV,
```
liamstask, line 286: with the test changes in this PR, i found this test was failing reliability locally for me as well, rather than just on circleci - i noticed this in response to the relocation of the `wsgi_server` and `native_server` fixtures above.

cheroot/testing.py
```diff
@@ -51,6 +52,7 @@ def cheroot_server(server_factory):
             break
 
     httpserver.shutdown_timeout = 0  # Speed-up tests teardown
+    httpserver.ignored_log_msgs = []  # test-only, log errors to ignore
```
liamstask, line 55: please let me know if there's a better/preferred way to make the list of messages to ignore available to the `check_log_messages()` verification step

cheroot/connections.py
```diff
@@ -49,6 +50,58 @@ def prevent_socket_inheritance(sock):
...
+        Returns entries ready to read in the form:
+            (socket_file_descriptor, connection)
+        """
+        return (
+            (key.fd, key.data)
+            for key, _ in self._selector.select(timeout=timeout)
+        )
+
+    def close(self):
+        self._selector.close()
```
webknjaz, line 102: I believe the methods here need some docstrings.

cheroot/connections.py
```diff
@@ -132,47 +184,49 @@ def get_conn(self):  # noqa: C901  # FIXME
             # The timeout value impacts performance and should be carefully
             # chosen. Ref:
             # github.com/cherrypy/cheroot/issues/305#issuecomment-663985165
-            rlist = [
-                key for key, _
-                in self._selector.select(timeout=0.01)
-            ]
+            active_list = self._selector.select(timeout=0.01)
```
webknjaz, line 187: Is turning a list into a generator safe here?

cheroot/connections.py
```diff
@@ -132,47 +184,49 @@ def get_conn(self):  # noqa: C901  # FIXME
...
             self._readable_conns.append(conn)
 
         try:
             return self._readable_conns.popleft()
         except IndexError:
             return None
 
+    def _remove_invalid_sockets(self):
+        # Mark any connection which no longer appears valid
+        # If the server is invalid, we'll just shutdown.
```
webknjaz, line 210: Not sure if this comment belongs at the top of this method... I'm not even sure what it means in this context. `s/server/client/`?

cheroot/connections.py
```diff
@@ -132,47 +184,49 @@ def get_conn(self):  # noqa: C901  # FIXME
...
+            self._selector.unregister(sock_fd)
             self._readable_conns.append(conn)
 
         try:
             return self._readable_conns.popleft()
         except IndexError:
             return None
 
+    def _remove_invalid_sockets(self):
+        # Mark any connection which no longer appears valid
```
webknjaz, line 209: It'd be great if this could be a docstring.

cheroot/test/conftest.py
webknjaz: My intention was to have things that would later become official testing helpers in `cheroot.testing`. That's why I kept them there. Of course, the fixtures will need refactoring before they are truly ready for public consumption (the generator based interfaces are not extendable, nor composable).
So I'm not sure if moving them around is a good thing.

cheroot/testing.py
```diff
@@ -61,18 +63,22 @@ def cheroot_server(server_factory):
...
-    for srv in cheroot_server(cheroot.wsgi.Server):
-        yield srv
-
-
-@pytest.fixture(scope='module')
-def native_server():
-    """Set up and tear down a Cheroot HTTP server instance."""
-    for srv in cheroot_server(cheroot.server.HTTPServer):
-        yield srv
+def check_log_messages(ignored_msgs, caplog):
```
webknjaz, line 66: As I wrote earlier, it'd like this module to only contain things that have the potential to be exposed as pytest-fixtures for end-users to test their Cheroot-based apps. I'm not sure if this helper should be here. What do you think?

cheroot/test/test_cli.py
```diff
@@ -34,7 +34,7 @@
 def test_parse_wsgi_bind_addr(raw_bind_addr, expected_bind_addr):
     """Check the parsing of the --bind option.
 
-    Verify some of the supported addresses and the excpected return value.
+    Verify some of the supported addresses and the expected return value.
```
webknjaz, line 37: This is unrelated and is fixed on master.

cheroot/test/test_core.py
```diff
@@ -283,7 +279,6 @@ def test_content_length_required(test_client):
...
     reason='https://github.com/cherrypy/cheroot/issues/106',
     strict=False,  # sometimes it passes
```
webknjaz, line 283: If it's failing reliably we should probably make this strict or/and change the conditional.

cheroot/testing.py
webknjaz: I'd probably say that it's reasonable to have the checks in test functions rather than outside... Except for the generic case. But it's probably okay for starters.

cheroot/server.py
```diff
@@ -1464,20 +1480,6 @@ def peer_group(self):
...
-    def _close_kernel_socket(self):
-        """Close kernel socket in outdated Python versions.
-
-        On old Python versions,
-        Python's socket module does NOT call close on the kernel
-        socket when you call socket.close(). We do so manually here
-        because we want this server to send a FIN TCP segment
-        immediately. Note this must be called *before* calling
-        socket.close(), because the latter drops its reference to
-        the kernel socket.
```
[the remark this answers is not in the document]
cyraxjoe: Yes, this is the original commit where this change was introduced https://github.com/cyraxjoe/cheroot/commit/b8d0865cabad55eb6bb52cd84d695d1731acdfdf, I remember stepping on the debugger while having a wireshark session open to make sure the connection was closed at the _OS_ level with shutdown. As for the special case for pyopenssl... I based that decision solely on pyopenssl docs, the plain `shutdown` exposed doesn't receive any arguments, later I found that the approach of using that method instead of the one for the underlying socket was also used in https://github.com/cherrypy/cheroot/blob/master/cheroot/workers/threadpool.py#L300-L314. Alternatively we could also use that here, and verify that it works with tcpdump or wireshark.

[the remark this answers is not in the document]
webknjaz: The main motivation of this question is to ensure there's no regression since the change is not annotated with the explanation comments confirming this.

cheroot/server.py
```diff
@@ -1307,9 +1307,7 @@ def communicate(self):  # noqa: C901  # FIXME
         except errors.NoSSLError:
             self._handle_no_ssl(req)
         except Exception as ex:
-            self.server.error_log(
-                repr(ex), level=logging.ERROR, traceback=True,
-            )
+            self.server.log.exception(repr(ex))
```
webknjaz, line 1310: `logging.exception()` can retrieve the exception from the current context so passing it explicitly is not necessary. But as I said earlier, it's something to do in a separate PR.

cheroot/server.py
```diff
@@ -1353,7 +1351,25 @@ def close(self):
         self.rfile.close()
 
         if not self.linger:
-            self._close_kernel_socket()
+            try:
```
webknjaz, line 1354: I believe that try/except should only wrap the shutdown call, not the logic of retrieving the proper method.

webknjaz (commented):
Hey @liamstask, I've finally found some time to post this review. Do you think you could split the changes so we could continue addressing this?

liamstask:
Sounds good - I think I'll open another PR with just the selector changes, and we can discuss separately whether/how to make the testing and logging changes, if that makes sense to you.

webknjaz:
@liamstask perfect, thanks!

liamstask:
marking as draft for now, in favor of https://github.com/cherrypy/cheroot/pull/337

jaraco:
Sounds like #337 supersedes the effort here (though more selectively), so I'm going to close this one for now. We can re-open or re-submit as needed.

PR #332, opened 2020-11-02
fixes #245 - handle OpenSSL.SSL.WantReadError and WantWriteError

### LINKED ISSUE #245: OpenSSL.SSL.WantReadError
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request

We are upgrading from python 2.7 to 3.7 and hence also upgraded the cheroot 8.2.0.
Cherrypy : 18+
This stack trace:
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cprequest.py", line 628, in respond
    self._do_respond(path_info)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cprequest.py", line 680, in _do_respond
    self.body.process()
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cpreqbody.py", line 982, in process
    super(RequestBody, self).process()
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cpreqbody.py", line 559, in process
    proc(self)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cpreqbody.py", line 145, in process_urlencoded
    qs = entity.fp.read()
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cherrypy/_cpreqbody.py", line 817, in read
    data = self.fp.read(chunksize)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cheroot/server.py", line 383, in read
    data = self.rfile.read(size)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/cheroot/makefile.py", line 420, in read
    val = super().read(*args, **kwargs)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/_pyio.py", line 1014, in read
    return self._read_unlocked(size)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/_pyio.py", line 1054, in _read_unlocked
    chunk = self.raw.read(wanted)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1335, in recv_into
    self._raise_ssl_error(self._ssl, result)
  File "/usr/local/pyenv/versions/3.7.0/lib/python3.7/site-packages/OpenSSL/SSL.py", line 1149, in _raise_ssl_error
    raise WantReadError()
OpenSSL.SSL.WantReadError

💡 **To Reproduce**

1. Run '...'
2. Make request '....'
3. See error

💡 **Expected behavior**

📋 **Details**

📋 **Environment**

- Cheroot version: X.X.X
- CherryPy version: X.X.X (if applicable)
- Python version: 3.6.X
- OS: XXX
- Browser: [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

📋 **Additional context**

vashek:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#245 

📋 **Other information**:

📋 **Checklist**:

### DIFF
--- a/cheroot/makefile.py
+++ b/cheroot/makefile.py
@@ -36,13 +36,19 @@ class BufferedWriter(io.BufferedWriter):
             self._flush_unlocked()
             return len(b)
 
+    def _safe_call(self, is_reader, call, *args, **kwargs):  # noqa: C901
+        """Method to be overridden in subclasses/mix-ins. Intended to call the
+        supplied callable with retries, as needed."""
+        return call(*args, **kwargs)
+
     def _flush_unlocked(self):
         self._checkClosed('flush of closed file')
         while self._write_buf:
             try:
                 # ssl sockets only except 'bytes', not bytearrays
                 # so perhaps we should conditionally wrap this for perf?
-                n = self.raw.write(bytes(self._write_buf))
+                n = self._safe_call(False, self.raw.write,
+                                    bytes(self._write_buf))
             except io.BlockingIOError as e:
                 n = e.characters_written
             del self._write_buf[:n]
--- a/cheroot/ssl/pyopenssl.py
+++ b/cheroot/ssl/pyopenssl.py
@@ -150,13 +150,19 @@ class SSLFileobjectMixin:
             size,
         )
 
-    def sendall(self, *args, **kwargs):
-        """Send whole message to the socket."""
+    def read(self, *args, **kwargs):
+        """Read from the wrapped socket, with retry."""
         return self._safe_call(
-            False,
-            super(SSLFileobjectMixin, self).sendall,
+            True,
+            super(SSLFileobjectMixin, self).read,
             *args, **kwargs
         )
+
+    def sendall(self, *args, **kwargs):
+        """Send whole message to the socket - not supported due to
+        https://github.com/pyca/pyopenssl/issues/176."""
+        raise NotImplementedError("sendall() not supported on pyOpenSSL due "
+                                  "to issue #176")
 
     def send(self, *args, **kwargs):
         """Send some part of message to the socket."""


cheroot/makefile.py
```diff
@@ -36,13 +36,22 @@ def write(self, b):
             self._flush_unlocked()
             return len(b)
 
+    def _safe_call(self, is_reader, call, *args, **kwargs):  # noqa: C901
```
webknjaz, line 39: Are you sure it's C901? It doesn't look like it's complex.

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,23 @@ def readline(self, size=-1):
...
-            super(SSLFileobjectMixin, self).sendall,
+            True,
+            super(SSLFileobjectMixin, self).read,
             *args, **kwargs
         )
 
+    def sendall(self, *args, **kwargs):
+        """Send whole message to the socket.
+
+        Not supported due to https://github.com/pyca/pyopenssl/issues/176.
```
webknjaz, line 164: Could you please expand the explanation in a code comment?

cheroot/makefile.py
vashek: Oh, right, sorry, I just copied the method signature from `pyopenssl.SSLFileobjectMixin` including the comment :)

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,23 @@ def readline(self, size=-1):
...
+        """
+        raise NotImplementedError(
+            'sendall() not supported on pyOpenSSL due to issue #176'
```
webknjaz, line 167: I think the issue number is out of context if it happens in runtime. So it's probably better not to mention it at all. The caller wouldn't know that the number belongs to some other repo.
```suggestion
            'sendall() is unsupported by pyOpenSSL',
```

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,23 @@ def readline(self, size=-1):
...
+        )
```
webknjaz, line 168: Also, wouldn't it be cleaner if this was reimplemented using a series of `send()` calls in a loop?

vashek: I suppose it would, but that would mean working around a third-party bug in a method that we don't actually use at all. I could also just remove the method, but I like that this was makes the non-support explicit.

webknjaz:
@vashek have you considered fixing pyopenssl itself? https://github.com/pyca/pyopenssl/issues/176#issuecomment-720471766

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,26 @@ def readline(self, size=-1):
...
+        """Send whole message to the socket.
+
+        Not supported due to https://github.com/pyca/pyopenssl/issues/176.
+        Until that bug is fixed, sendall() may throw SSL.WantWriteError, but
+        there is no correct way to retry the call because we don't know how
+        many bytes were already transmitted. We could work around this by
+        reimplementing sendall() using send(), but we don't actually use
+        sendall() anywhere.
+        """
+        raise NotImplementedError('sendall() is unsupported by pyOpenSSL')
```
webknjaz, line 171: How about taking https://github.com/pyca/pyopenssl/blob/124a013/src/OpenSSL/SSL.py#L1649-L1677 and adding retries and then resubmitting the fix to the upstream too?

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,26 @@ def readline(self, size=-1):
...
+    def sendall(self, *args, **kwargs):
+        """Send whole message to the socket.
+
+        Not supported due to https://github.com/pyca/pyopenssl/issues/176.
+        Until that bug is fixed, sendall() may throw SSL.WantWriteError, but
+        there is no correct way to retry the call because we don't know how
+        many bytes were already transmitted. We could work around this by
+        reimplementing sendall() using send(), but we don't actually use
+        sendall() anywhere.
+        """
```
webknjaz, line 170: Could you please make this a code comment, I don't think it belongs in a docstring.
```suggestion
        """
        # Until that bug is fixed, sendall() may throw SSL.WantWriteError, but
        # there is no correct way to retry the call because we don't know how
        # many bytes were already transmitted. We could work around this by
        # reimplementing sendall() using send(), but we don't actually use
        # sendall() anywhere.
```

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,26 @@ def readline(self, size=-1):
...
 
+    def sendall(self, *args, **kwargs):
+        """Send whole message to the socket.
+
+        Not supported due to https://github.com/pyca/pyopenssl/issues/176.
+        Until that bug is fixed, sendall() may throw SSL.WantWriteError, but
+        there is no correct way to retry the call because we don't know how
+        many bytes were already transmitted. We could work around this by
+        reimplementing sendall() using send(), but we don't actually use
+        sendall() anywhere.
```
webknjaz, line 169: When you say "we don't use" do you mean not even some internal calls hit it? How do we get the error, then?

cheroot/ssl/pyopenssl.py
webknjaz: P.S. FYI it's a good tone to accept suggestions via GH UI because that preserves authorship.

cheroot/ssl/pyopenssl.py
vashek: We don't, as far as I know, it's just something I noticed while googling the WantWriteError issue.
There is no occurrence of `sendall` in the whole cheroot project, other than in `pyopenssl.py`. (Also, my server seems to work with this change.)

webknjaz: Well, if it's unrelated to the PR, it should be in a separate PR.

cheroot/ssl/pyopenssl.py
webknjaz: Looks like it's not really possible to do this now, after all.

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,24 @@ def readline(self, size=-1):
...
 
+    def sendall(self, *args, **kwargs):
+        """Send whole message to the socket. Unsupported, do not use."""
+        # Not supported due to https://github.com/pyca/pyopenssl/issues/176.
+        # Until that bug is fixed, sendall() may throw SSL.WantWriteError, but
+        # there is no correct way to retry the call because we don't know how
+        # many bytes were already transmitted. We could work around this by
+        # reimplementing sendall() using send(), but we don't actually use
+        # sendall() anywhere.
+        raise NotImplementedError('sendall() is unsupported by pyOpenSSL')
```
webknjaz, line 169: I've given this some thought. I think we shouldn't proactively block this but rather call, catch the errors and turn those errors into `RuntimeError` referring to the explanation that it will work sometimes and sometimes it won't.

webknjaz:
Hm... it looks like `sendall` was used as `write()` + `flush()` at some point but then got replaced with `write`: https://github.com/cherrypy/cheroot/commit/60cd709551d564551d91a5b170c311bb23ad2ac6 / https://github.com/cherrypy/cheroot/commit/a0b1d018c912aa21473bac3f99d46018faed5563.

webknjaz:
Some jobs are failing because of a deprecation warning in cryptography. I've fixed that on `master`. To get the changes, rebase this PR branch.

cheroot/ssl/pyopenssl.py
vashek: In general I would much prefer a randomly failing function to just never work than randomly work, but the only thing that really matters to me at the moment is fixing #245, so whatever will cause the PR to be accepted.

webknjaz: I think this has to be taken out into a separate PR. This is one of the problems of non-atomic PRs: many things can block them which they could've been debated separately.

cheroot/ssl/pyopenssl.py
```diff
@@ -150,14 +150,24 @@ def readline(self, size=-1):
             size,
         )
 
-    def sendall(self, *args, **kwargs):
-        """Send whole message to the socket."""
+    def read(self, *args, **kwargs):
+        """Read from the wrapped socket, with retry."""
```
webknjaz, line 154: I think that retry is an unnecessary implementation detail here.

webknjaz:
Do you think this can be tested somehow?

cheroot/ssl/pyopenssl.py
vashek: ok

cheroot/ssl/pyopenssl.py
vashek: fine, undone

webknjaz:
@vashek so what do you think about tests?

webknjaz:
The test will probably need to make the sending and/or receiving socket full like I tried here: https://github.com/pyca/pyopenssl/pull/955/files#diff-5fbedfbbf8f0780aeee680927973f302330dc10fe915c7e7deecf4b0556c492cR3157

vashek:
Sorry for letting this rot. Quite honestly, I'd appreciate if someone could help with the tests - or accept it without them.

vashek:
Any chance of accepting this?

cheroot/ssl/pyopenssl.py
webknjaz: Another interesting code snippet: https://github.com/Lawouach/WebSocket-for-Python/issues/174#issuecomment-150142032.

webknjaz:
@vashek I suppose it's hard to write tests for this which is why I'll probably accept it without tests. But https://github.com/cherrypy/cheroot/issues/245#issuecomment-1465341867 seems to imply that it may be causing problems for retries. Do you have any thoughts on that?

PR #337, opened 2020-11-23
connections: add threadsafe selector wrapper

liamstask:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#317

The 'connections' selector can be modified from thread pool threads, so the underlying selector may be modified during iteration.

Add a threadsafe wrapper for the selector.

📋 **Other information**:

This is the primary change from https://github.com/cherrypy/cheroot/pull/325 but does not include the other changes to logging & testing included there, in order to reduce scope.

This does not include a direct reproducer, but does include a change from https://github.com/cherrypy/cheroot/pull/325/commits/0688b50826c53bffb923d613dcfc53788e5057d9 that removes the suppression of a related error case - this suppression was previously allowing this error to go unnoticed.

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/337)

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -51,22 +51,22 @@ else:
 
 
 class _ThreadsafeSelector:
-    """Threadsafe wrapper around a DefaultSelector.
+    """Thread-safe wrapper around a DefaultSelector.
 
     There are 2 thread contexts in which it may be accessed:
     * the selector thread
     * one of the worker threads in workers/threadpool.py
 
     The expected read/write patterns are:
-    * iter: selector thread
-    * register: selector thread and threadpool, via put()
-    * unregister: selector thread only
+    * :py:meth:`iter`: selector thread
+    * :py:meth:`register`: selector thread and threadpool, via put()
+    * :py:meth:`unregister`: selector thread only
 
-    Notably, this means _ThreadsafeSelector never needs to worry
+    Notably, this means :py:class:`_ThreadsafeSelector` never needs to worry
     that connections will be removed behind its back.
 
-    The lock is held when iterating or modifying the selector,
-    but is not required when select()ing on it.
+    The lock is held when iterating or modifying the selector but is not
+    required when :py:meth:`~selectors.BaseSelector.select`-ing on it.
     """
 
     def __init__(self):
@@ -74,11 +74,8 @@ class _ThreadsafeSelector:
         self._lock = threading.Lock()
 
     def __len__(self):
-        # NOTE: the _SelectorMapping returned by get_map()
-        # implements its own __len__ routine which does not
-        # iterate the map, so it is not required
-        # to hold the lock here.
-        return len(self._selector.get_map())
+        with self._lock:
+            return len(self._selector.get_map() or {})
 
     @property
     def connections(self):
@@ -89,17 +86,17 @@ class _ThreadsafeSelector:
                 yield (sock_fd, conn)
 
     def register(self, fileobj, events, data=None):
-        """Register fileobj with the selector."""
+        """Register ``fileobj`` with the selector."""
         with self._lock:
             return self._selector.register(fileobj, events, data)
 
     def unregister(self, fileobj):
-        """Unregister fileobj from the selector."""
+        """Unregister ``fileobj`` from the selector."""
         with self._lock:
             return self._selector.unregister(fileobj)
 
     def select(self, timeout=None):
-        """Wrap the selectors.select call.
+        """Return socket fd and data pairs from selectors.select call.
 
         Returns entries ready to read in the form:
             (socket_file_descriptor, connection)
@@ -217,8 +214,12 @@ class ConnectionManager:
             return None
 
     def _remove_invalid_sockets(self):
-        # Mark any connection which no longer appears valid
-        # If the server is invalid, we'll just shutdown.
+        """Clean up the resources of any broken connections.
+
+        This method attempts to detect any connections in an invalid state,
+        unregisters them from the selector and closes the file descriptors of
+        the corresponding network sockets where possible.
+        """
         invalid_conns = []
         for sock_fd, conn in self._selector.connections:
             if conn is self.server:
@@ -335,7 +336,7 @@ class ConnectionManager:
         self._readable_conns.clear()
 
         for (_, conn) in self._selector.connections:
-            if conn != self.server:  # server closes its own socket
+            if conn is not self.server:  # server closes its own socket
                 conn.close()
         self._selector.close()
 
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -231,7 +231,12 @@ class ConnectionManager:
 
         for sock_fd, conn in invalid_conns:
             self._selector.unregister(sock_fd)
-            conn.close()
+            # One of the reason on why a socket could cause an error
+            # is that the socket is already closed, ignore the
+            # socket error if we try to close it at this point.
+            # This is equivalent to OSError in Py3
+            with suppress(socket.error):
+                conn.close()
 
     def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
         try:
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -7,6 +7,7 @@ import collections
 import io
 import os
 import socket
+import threading
 import time
 
 from . import errors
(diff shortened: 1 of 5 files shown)


cheroot/test/test_server.py
```diff
@@ -135,7 +135,7 @@ def raise_keyboard_interrupt(*args):
     serve_thread.start()
 
     # The thread should exit right away due to the interrupt.
-    serve_thread.join(0.5)
+    serve_thread.join(1.5)
```
liamstask, line 138: i am not sure this is needed - but saw this timeout at least once without it.

cheroot/connections.py
```diff
@@ -130,47 +195,49 @@ def get_conn(self):  # noqa: C901  # FIXME
...
-            self._selector.unregister(key.fd)
+            self._selector.unregister(sock_fd)
             self._readable_conns.append(conn)
 
         try:
             return self._readable_conns.popleft()
         except IndexError:
             return None
 
+    def _remove_invalid_sockets(self):
```
webknjaz, line 219:
```suggestion
    def _remove_invalid_sockets(self):
        """Clean up the resources of any broken connections.

        This method attempts to detect any connections in the invalid state,
        unregisters them from the selector and closes the file descriptors of
        the corresponding network sockets where possible.
        """
```

cheroot/connections.py
```diff
@@ -267,14 +334,9 @@ def close(self):
...
-        connections = (
-            conn.data.socket
-            for _, conn in (self._selector.get_map() or {}).items()
-            if conn.data != self.server  # server closes its own socket
-        )
-        for connection in connections:
-            connection.close()
-
+        for (_, conn) in self._selector.connections:
+            if conn != self.server:  # server closes its own socket
```
webknjaz, line 338: Using the identity check is probably better here:
```suggestion
            if conn is not self.server:  # server closes its own socket
```

cheroot/server.py
```diff
@@ -1353,7 +1354,27 @@ def close(self):
...
+            sock_shutdown = getattr(
+                # The special method sock_shutdown is used for
+                # the PyOpenSSL connections that can be used
+                # as socket.
+                self.socket, 'sock_shutdown',
+                self.socket.shutdown,
+            )
+
+            try:
+                sock_shutdown(socket.SHUT_RDWR)
```
webknjaz, line 1367:
```suggestion
                sock_shutdown(socket.SHUT_RDWR)  # actually send a TCP FIN
```

cheroot/test/test_server.py
webknjaz: do you have any logs of what the traceback was when this happens?

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+        with self._lock:
+            return self._selector.register(fileobj, events, data)
+
+    def unregister(self, fileobj):
+        """Unregister fileobj from the selector."""
+        with self._lock:
+            return self._selector.unregister(fileobj)
+
+    def select(self, timeout=None):
+        """Wrap the selectors.select call.
```
webknjaz, line 102:
```suggestion
        """Return socket fd and data pairs from selectors.select call.
```

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+    """Threadsafe wrapper around a DefaultSelector.
+
+    There are 2 thread contexts in which it may be accessed:
+    * the selector thread
+    * one of the worker threads in workers/threadpool.py
+
+    The expected read/write patterns are:
+    * iter: selector thread
+    * register: selector thread and threadpool, via put()
+    * unregister: selector thread only
```
webknjaz, line 63:
```suggestion
    * :py:meth:`iter`: selector thread
    * :py:meth:`register`: selector thread and threadpool, via put()
    * :py:meth:`unregister`: selector thread only
```

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
         fcntl.fcntl(fd, fcntl.F_SETFD, old_flags | fcntl.FD_CLOEXEC)
 
 
+class _ThreadsafeSelector:
+    """Threadsafe wrapper around a DefaultSelector.
```
webknjaz, line 54: Looks like the spell-checker doesn't like a few things: https://github.com/cherrypy/cheroot/pull/337/checks?check_run_id=1440728038#step:9:277

```suggestion
    """Thread-safe wrapper around a DefaultSelector.
```

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+    There are 2 thread contexts in which it may be accessed:
+    * the selector thread
+    * one of the worker threads in workers/threadpool.py
+
+    The expected read/write patterns are:
+    * iter: selector thread
+    * register: selector thread and threadpool, via put()
+    * unregister: selector thread only
+
+    Notably, this means _ThreadsafeSelector never needs to worry
```
webknjaz, line 65:
```suggestion
    Notably, this means :py:class:`_ThreadsafeSelector` never needs to worry
```

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+    that connections will be removed behind its back.
+
+    The lock is held when iterating or modifying the selector,
+    but is not required when select()ing on it.
```
webknjaz, line 69:
```suggestion
    The lock is held when iterating or modifying the selector but is not
    required when :py:meth:`~selectors.BaseSelector.select`-ing on it.
```

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+            for _, (_, sock_fd, _, conn) in mapping.items():
+                yield (sock_fd, conn)
+
+    def register(self, fileobj, events, data=None):
+        """Register fileobj with the selector."""
+        with self._lock:
+            return self._selector.register(fileobj, events, data)
+
+    def unregister(self, fileobj):
+        """Unregister fileobj from the selector."""
```
webknjaz, line 97:
```suggestion
        """Unregister ``fileobj`` from the selector."""
```

webknjaz, line 97: And "unregister" should probably be whitelisted @ https://github.com/cherrypy/cheroot/blob/master/docs/spelling_wordlist.txt.

cheroot/connections.py
```diff
@@ -49,6 +50,71 @@ def prevent_socket_inheritance(sock):
...
+    def __init__(self):
+        self._selector = selectors.DefaultSelector()
+        self._lock = threading.Lock()
+
+    def __len__(self):
+        # NOTE: the _SelectorMapping returned by get_map()
+        # implements its own __len__ routine which does not
+        # iterate the map, so it is not required
+        # to hold the lock here.
+        return len(self._selector.get_map())
```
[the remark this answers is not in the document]
liamstask: in this case, it's just accessing the len on the underlying dict as opposed to iterating, so should be covered by the GIL, but I added a lock here just so it's less confusing

cheroot/test/test_server.py
jaraco: Is this change in timeout related to the PR or incidental? If it's incidental, I'd rather see a separate issue describing the aberrant behavior (traceback) and a separate PR to work around or correct it.

cheroot/server.py
```diff
@@ -1353,7 +1354,27 @@ def close(self):
         self.rfile.close()
 
         if not self.linger:
-            self._close_kernel_socket()
+            # Make sure we terminate the connection at the transport level.
```
jaraco, line 1357: This block reads easier when there's a method encapsulating the purpose. I recommend bringing back `_close_kernel_socket` but just change its docstring to `Terminate the connection at the transport level.`

jaraco:
Overall, this looks good to me. I only wonder, since it doesn't have any tests, do we have any confidence that the issue is resolved? Without having a reproducer, my only option is to roll it out and see if users can detect the improvement. So I'll do that.

The test suite seems to be failing approximately as much here as in master, so I'm going to assume the failures are unrelated.

I'll add a couple of commits to address my concerns and then push forward.

cheroot/server.py
jaraco: Fixed in 8664cd2e.

cheroot/connections.py
jaraco: Looks like this has been inlined somewhere.

cheroot/test/test_server.py
jaraco: I've rebased the changes without 7c07fe8b. We can apply that later if needed.

cheroot/connections.py
jaraco: This suggestion was accepted inline.

cheroot/server.py
jaraco: This suggestion was accepted inline.

cheroot/connections.py
jaraco: This change was accepted inline.

cheroot/connections.py
jaraco: This change was accepted inline.

cheroot/connections.py
jaraco: This change was accepted inline. In the future, if you're accepting someone's change, consider also resolving the comment.

cheroot/connections.py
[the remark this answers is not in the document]
webknjaz: @liamstask oh, I didn't think about that. I was convinced that `len()` on a dict iterates over its elements for some reason. But it sounds like you are probably right about this...

cheroot/connections.py
webknjaz: @jaraco I usually recommend GH UI because it preserves the authorship and makes it easier to track what's accepted. Should we mention this in the contributing doc?

jaraco: I also recommend the GH UI, although that approach runs contrary to the other recommendation to 'squash related commits'. I personally have no objection to contributors amending commits with suggested changes, especially nitpicks or other minor edits. I mainly wished here to publish my surprise when I found that the comments were unresolved, but the changes were accepted. I'm okay without any guidance (just gentle nudges) or if you feel fervent about it, contributor guidance might be helpful. I do try to keep it short and sweet, as it's important to me to keep the barrier to contribution low.

liamstask: Happy to abide by whatever you prefer for this project - I found this unexpected in this case, though, since historically github has auto-collapsed (if not actually resolved) comments on code that has been modified in subsequent commits. Now they appear to only be labeled `Outdated`. Perhaps that has been changed in favor of explicit resolution of comments?

liamstask:
@jaraco wrt a reproducer - we don't have a concrete one, but this PR removes a suppressed error [here](https://github.com/cherrypy/cheroot/pull/337/files#diff-9a0bf410a5532a2425a663accbf6260f52b70fa9894d0752ce0671d8c7fb79afL1155) that was previously masking this failure in the test suite.

with that suppression removed & the fix _not_ applied, it was locally observed to occur frequently enough in the existing test suite to get at least some confidence that this change addressed it.

cheroot/connections.py
webknjaz: @liamstask they appear outdated if changed in the branch. But they'd appear resolved if the UI button was used. Also, if you accept changes from the UI, it preserves the authorship of the person who submitted the suggested change. I usually accept suggestions via UI, then do pull to my local repo, and then squash or rearrange things as necessary.

PR #340, opened 2020-11-30
server: raise exceptions from worker threads on serve() thread

liamstask:
  - [x] 🐞 bug fix
  - [ ] 🐣 feature
  - [ ] 📋 docs update
  - [x] 📋 refactoring
  - [ ] 💥 other

https://github.com/cherrypy/cherrypy/issues/1873

Exceptions from worker threads are raised in the context that `stop()` is called, rather than the thread that `serve()` runs on - this breaks a test in cherrypy.

This was inadvertently changed in https://github.com/cherrypy/cheroot/pull/309 - this reverts the behavior to raise the exception from the `serve()` thread as it was previously.

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/340)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -162,6 +162,9 @@ QUOTED_SLASH = b'%2F'
 QUOTED_SLASH_REGEX = re.compile(b''.join((b'(?i)', QUOTED_SLASH)))
 
 
+_STOPPING_FOR_INTERRUPT = object()  # sentinel used during shutdown
+
+
 comma_separated_headers = [
     b'Accept', b'Accept-Charset', b'Accept-Encoding',
     b'Accept-Language', b'Accept-Ranges', b'Allow', b'Cache-Control',
@@ -1802,7 +1805,7 @@ class HTTPServer:
 
     def serve(self):
         """Serve requests, after invoking :func:`prepare()`."""
-        while self.ready:
+        while self.ready and not self.interrupt:
             try:
                 self._connections.run(self.expiration_interval)
             except (KeyboardInterrupt, SystemExit):
@@ -1812,6 +1815,14 @@ class HTTPServer:
                     'Error in HTTPServer.serve', level=logging.ERROR,
                     traceback=True,
                 )
+
+        # raise exceptions reported by any worker threads,
+        # such that the exception is raised from the serve() thread.
+        if self.interrupt:
+            while self._stopping_for_interrupt:
+                time.sleep(0.1)
+            if self.interrupt:
+                raise self.interrupt
 
     def start(self):
         """Run the server forever.
@@ -2053,14 +2064,23 @@ class HTTPServer:
         """Flag interrupt of the server."""
         return self._interrupt
 
+    @property
+    def _stopping_for_interrupt(self):
+        """Return whether the server is responding to an interrupt."""
+        return self._interrupt is _STOPPING_FOR_INTERRUPT
+
     @interrupt.setter
     def interrupt(self, interrupt):
-        """Perform the shutdown of this server and save the exception."""
-        self._interrupt = True
+        """Perform the shutdown of this server and save the exception.
+
+        Typically invoked by a worker thread in
+        :py:mod:`~cheroot.workers.threadpool`, the exception is raised
+        from the thread running :py:meth:`serve` once :py:meth:`stop`
+        has completed.
+        """
+        self._interrupt = _STOPPING_FOR_INTERRUPT
         self.stop()
         self._interrupt = interrupt
-        if self._interrupt:
-            raise self.interrupt
 
     def stop(self):  # noqa: C901  # FIXME
         """Gracefully shutdown a server that is serving forever."""
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -119,6 +119,7 @@ def test_stop_interrupts_serve():
 
 def test_server_interrupt():
     """Check that assigning interrupt stops the server."""
+    interrupt_msg = 'should catch {uuid!s}'.format(uuid=uuid.uuid4())
     raise_marker_sentinel = object()
 
     httpserver = HTTPServer(
@@ -133,7 +134,7 @@ def test_server_interrupt():
         try:
             httpserver.serve()
         except RuntimeError as e:
-            if str(e) == 'should catch':
+            if str(e) == interrupt_msg:
                 result_q.put(raise_marker_sentinel)
 
     httpserver.prepare()
@@ -145,7 +146,7 @@ def test_server_interrupt():
 
     # this exception is raised on the serve() thread,
     # not in the calling context.
-    httpserver.interrupt = RuntimeError('should catch')
+    httpserver.interrupt = RuntimeError(interrupt_msg)
 
     serve_thread.join(0.5)
     assert not serve_thread.is_alive()
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -119,6 +119,8 @@ def test_stop_interrupts_serve():
 
 def test_server_interrupt():
     """Check that assigning interrupt stops the server."""
+    raise_marker_sentinel = object()
+
     httpserver = HTTPServer(
         bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT),
         gateway=Gateway,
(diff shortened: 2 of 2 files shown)


liamstask:
marking as draft pending a test

cheroot/server.py
```diff
@@ -1489,6 +1489,7 @@ class HTTPServer:
 
     _bind_addr = '127.0.0.1'
     _interrupt = None
+    _STOPPING_FOR_INTERRUPT = object()  # sentinel used during shutdown
```
webknjaz, line 1492: I think it's better to put constants like this to the module level.

cheroot/server.py
```diff
@@ -1810,6 +1811,14 @@ def serve(self):
                     traceback=True,
                 )
 
+            # raise exceptions reported by any worker threads,
+            # such that the exception is raised from the serve() thread.
+            if self.interrupt:
+                while self._stopping_for_interrupt:
+                    time.sleep(0.1)
+                if self.interrupt:
+                    raise self.interrupt
```
webknjaz, line 1820: Let's avoid repeating the recent regression:
```suggestion
                    self.serving = False
                    raise self.interrupt
```

cheroot/server.py
```diff
@@ -2056,14 +2068,22 @@ def interrupt(self):
...
+
     @interrupt.setter
     def interrupt(self, interrupt):
-        """Perform the shutdown of this server and save the exception."""
-        self._interrupt = True
+        """Perform the shutdown of this server and save the exception.
+
+        Typically invoked by a worker thread in threadpool.py,
+        the exception is raised from the thread running serve()
+        once self.stop() has completed.
```
webknjaz, line 2082: Let's reference the actual method docs in sphinx:
```suggestion
        the exception is raised from the thread running :py:meth:`serve`
        once :py:meth:`stop` has completed.
```

cheroot/server.py
```diff
@@ -2056,14 +2068,22 @@ def interrupt(self):
...
+        """Return whether the server is responding to an interrupt."""
+        return self._interrupt is _STOPPING_FOR_INTERRUPT
+
     @interrupt.setter
     def interrupt(self, interrupt):
-        """Perform the shutdown of this server and save the exception."""
-        self._interrupt = True
+        """Perform the shutdown of this server and save the exception.
+
+        Typically invoked by a worker thread in threadpool.py,
```
webknjaz, line 2080: Here's how you can reference the module in Sphinx
```suggestion
        Typically invoked by a worker thread in
        :py:mod:`~cheroot.workers.threadpool`,
```
(there's pages autogenerated from docstrings @ https://cheroot.cherrypy.org/en/latest/pkg/cheroot.server.html#cheroot.server.HTTPServer.interrupt)

webknjaz:
The linter failures are not fixed on master.

liamstask:
accepted suggestions, rebased on master to include most recent linting fixes

liamstask:
updated, and tested successfully locally against cherrypy/cherrypy master by:
* removing the `xfail` [here](https://github.com/cherrypy/cherrypy/blob/2a8aaccd649eb1011382c39f5cd93f76f980c0b1/cherrypy/test/test_states.py#L165)
* editing the [cheroot-master testenv](https://github.com/cherrypy/cherrypy/blob/2a8aaccd649eb1011382c39f5cd93f76f980c0b1/tox.ini#L30) to point to this PR:  `deps = git+git://github.com/liamstask/cheroot.git@dcd71fdc2af31910b3e866ae25e2438903294888#egg=cheroot`
* running `python3 -m tox -e cheroot-master`

webknjaz:
@liamstask do you think you could come up with a regression test for this?

liamstask:
test added

webknjaz:
@liamstask this seems to break some tests on 2.7 + needs a rebase (I've dropped the redundant CIs on master)

liamstask:
@webknjaz rebased - i am not sure i see any failures on this branch that are not also occurring on master, can you point to the failures you have in mind?

webknjaz:
Previous test runs had some extra failing tests but now those jobs get canceled too early because of the other failures... I'll try to do a few extra restarts to catch those.

webknjaz:
This job is where I saw more failing tests but it seems like they are not related:
https://github.com/cherrypy/cheroot/pull/340/checks?check_run_id=1514638021

webknjaz:
@liamstask by the way, do you have any ideas on how to stabilize the flaky tests that break master?

liamstask:
I am not totally sure - i have not been able to reproduce them locally yet to better understand what is actually failing.

Fundamentally, that specific test `test_keepalive_conn_management()` is not synchronized with the internal state of the server - ie, there are no specific guarantees that by the time the client makes the next request that a given `conn` has actually been re-registered with the selector, such that `server._connections.can_add_keepalive_connection` will return the expected value.

Granted, I don't see anything that should really account for substantial delays here, but I think the test is racy in any event.

we could try something like the following to synchronize this (quick sketch that passes locally):
```diff
diff --git a/cheroot/test/test_conn.py b/cheroot/test/test_conn.py
index 94fb878d..1f9b90f0 100644
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -494,6 +494,16 @@ def test_keepalive_conn_management(test_client):
             assert not header_exists('Connection', actual_headers)
             assert not header_exists('Keep-Alive', actual_headers)
 
+    def check_server_idle_conn_count(count, timeout=1.0):
+        deadline = time.time() + timeout
+        num_conns = -1
+        while time.time() < deadline:
+            num_conns = test_client.server_instance._connections._num_connections
+            if num_conns == count:
+                return
+        assert False, "idle conn count mismatch, wanted {}, got {}".format(
+            count, num_conns)
+
     disconnect_errors = (
         http_client.BadStatusLine,
         http_client.CannotSendRequest,
@@ -503,39 +513,48 @@ def test_keepalive_conn_management(test_client):
     # Make a new connection.
     c1 = connection()
     request(c1)
+    check_server_idle_conn_count(1)
 
     # Make a second one.
     c2 = connection()
     request(c2)
+    check_server_idle_conn_count(2)
 
     # Reusing the first connection should still work.
     request(c1)
+    check_server_idle_conn_count(2)
 
     # Creating a new connection should still work, but we should
     # have run out of available connections to keep alive, so the
     # server should tell us to close.
     c3 = connection()
     request(c3, keepalive=False)
+    check_server_idle_conn_count(2)
 
     # Show that the third connection was closed.
     with pytest.raises(disconnect_errors):
         request(c3)
+    check_server_idle_conn_count(2)
 
     # Wait for some of our timeout.
     time.sleep(1.2)
 
     # Refresh the second connection.
     request(c2)
+    check_server_idle_conn_count(2)
 
     # Wait for the remainder of our timeout, plus one tick.
     time.sleep(1.2)
+    check_server_idle_conn_count(1)
 
     # First connection should now be expired.
     with pytest.raises(disconnect_errors):
         request(c1)
+    check_server_idle_conn_count(1)
 
     # But the second one should still be valid.
     request(c2)
+    check_server_idle_conn_count(1)
 
     # Restore original timeout.
     test_client.server_instance.timeout = timeout
```

it's still not 100% synchronized since the server can still expire connections between the call to `check_server_idle_conn_count()` and the next step within the test, but it may be more reliable than the current test.

liamstask:
giving it a shot in https://github.com/cherrypy/cheroot/pull/349 - any recommendations on how best to 'soak test' this, or otherwise run the test repeatedly to check for flakiness?

webknjaz:
@liamstask thanks for the PR! I noticed that multiple restarts of the original test on master cause this race condition under Python 2 sometimes and with your patch the same procedure seems to be more stable on my laptop so it's definitely a good first step.
Going further, we may want to integrate some pytest plugin that would restart the test a few times and set the successful status only if a good percentage of restarts are good. Anyway, it seems to be way better with your patch than without it. There's other flaky tests that need separate attention, although they cause failures less frequently.

webknjaz:
Rebasing this PR now

webknjaz:
`test_connection_keepalive` is still unstable but overall this looks quite good. I'll try to push some small tweaks to the new test in this PR before mergning.

webknjaz:
I wanted to release this as v8.5.1 but after tens of restarts, the CI kept crashing. I guess I won't make any more attempts now that we could also include #350 into this release. On the bright note, changing the server fixture to be test function scoped instead of per-module seems to have stabilized the tests.

Issue #384, opened 2021-09-10
Add type hints


kasium:
❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [X] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐣 **Describe the solution you'd like**
It would be great to see the python 3.5+ type hints in the repository, especially for the ``Server`` api.

📋 **Describe alternatives you've considered**
Adding type stubs to typeshed (https://github.com/python/typeshed/tree/master/stubs).
If you think this makes more sense, that's possible. But I first wanted to ask here.

webknjaz:
Is love to see type hints here in Cheroot and in CherryPy. The next major release is supposed to be dropping py2 so it's possible to use native syntax. OTOH, we could also try pyi files or comment-based hints for something to be available in the 8.x stream. PR is welcome :)

kasium:
@webknjaz Let me try to add some more context and ideas
- type comments: very limited syntax but python 2 compatible. I guess it's not worth the effort due the deprecation of python 2
- pyi files: would then be part of typeshed I guess? However, I prefer native support, because you don't forget to adapt the types when you code. In addition, mypy can then also help you during development.

I'm happy to file a PR adding mypy and some utility stuff. I can also try to add some types by my own but I guess help of the actual core developers will be needed. In addition it should be decided, if you would like to type your whole codebase or only the public API. 

That's btw just my view, so please feel free to add yours 😄

webknjaz:
> - type comments: very limited syntax but python 2 compatible. I guess it's not worth the effort due the deprecation of python 2

There's more to it. We were thinking about keeping the 8.x for bugfixes. This means that we'd need to backport things from the newer versions. Active use of the newer syntax would mean the need to manually resolve conflicts, which adds to the maintainance burden. OTOH if the code base is similar enough, `git cherry-pick` will be able to apply patches cleanly, meaning that we could just rely on the backporting bot to do this.
So from this perspective, it's totally worth the effort. This is the main reason I proposed this.

> - pyi files: would then be part of typeshed I guess? However, I prefer native support, because you don't forget to adapt the types when you code. In addition, mypy can then also help you during development.

No, pyi files can also live in the project next to their counterparts.

> I'm happy to file a PR adding mypy and some utility stuff. I can also try to add some types by my own but I guess help of the actual core developers will be needed. In addition it should be decided, if you would like to type your whole codebase or only the public API. 

[...]

PR #385, opened 2021-09-24
Add basic type hints

kasium:
  - [ ] 🐞 bug fix
  - [X] 🐣 feature
  - [ ] 📋 docs update
  - [ ] 📋 refactoring
  - [ ] 💥 other

#384

#384

no new behavior 

📋 **Other information**:

📋 **Checklist**:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/385)


kasium:
So I gave some simple type hints a try. However I cannot active them in CI yet, since mypy will check the ,py files as well as the pyi ones.
Do you think it make sense to only let mypy run on pyi files?

kasium:
@webknjaz can you please have a look?

pyproject.toml
```diff
@@ -43,3 +43,18 @@ flake8-eradicate = ["-*"]
...
+show_error_codes = true
+warn_no_return = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "cheroot.test.*"
+ignore_errors = true
```
webknjaz, line 60: This should go to `mypy.ini` because it's hard to maintain a huge dump of different configs in a single file.

setup.cfg
```diff
@@ -143,6 +143,17 @@ testing =
...
 
+    # mypy is a static type checker
+    mypy==0.910
+
+    # type stubs for mypy
+    types-backports
+    types-setuptools
+    types-six
+    types-requests
+    types-pyOpenSSL
```
webknjaz, line 154: Type checking is not the same as running tests and has different dependencies. This is why it needs to live in a different requirements file. Also, I've started moving away from having the extras that are useless in runtime, env pins should live in dedicated constraints files.

Besides, you didn't add this to the CI and it should be present for this PR to make sense to be merged. I don't think that it's worth setting it up separately but we have pre-commit integrated so it'd be fine if you just move this to `.pre-commit-config.yaml`. This way, you won't need to add any other integration manually.

pyproject.toml
```diff
@@ -43,3 +43,18 @@ flake8-eradicate = ["-*"]
 flake8-isort = ["+*", "-I001", "-I003", "-I004"]
 # don't fail on unknown docutils roles because sphinx knows them
 flake8-rst-docstrings = ["+*", "-RST304", "-RST303", "-RST301", "-RST203", "-RST201"]
+
+[tool.mypy]
+pretty = true
+python_version = 3.6
```
webknjaz, line 49: Why 3.6? I'd start with 2.7 for the 8.x branch. Besides, it probably makes sense to test under different Pythons. WDYT?

webknjaz (changes requested):
This PR needs rebasing and I've also added a few notes on things that need to be fixed. Could you look into it?

setup.cfg
kasium: any suggestions how to call that file?

pyproject.toml
kasium: yes, mypy will just pickup the local interpreter

webknjaz: I thought maybe we should run under multiple interpreters/platforms to cover the gaps.

setup.cfg
webknjaz: Probably `requirements/mypy.in` or `requirements/types.in`. Just make sure to use `.in` as I'm planning to integrate pip-tools later and Dependabot only supports pip-tools setups with pairs of `.in` and `.txt` files with the same base name.

pyproject.toml
webknjaz: I think this is achievable via something like https://github.com/pypa/pip/blob/20.3.4/.pre-commit-config.yaml#L71-L80, at least partially.

setup.cfg
kasium: I never worked with pip-tools so can you please just tell me what to do here?

webknjaz: Oh, you don't need to integrate it right now, it's out of the scope of the current PR. I'm planning to do that, I wouldn't request you to integrate something that is not yet streamlined in the repo. I was just trying to explain that you need to create a file called `requirements/mypy.in` following `requirements.txt` syntax.

webknjaz: OTOH, if we're going to integrate this via pre-commit per https://github.com/cherrypy/cheroot/pull/385#discussion_r759814196, this file is not needed — the dependencies must be listed in `.pre-commit-config.yaml` as `additional_dependencies` under the respective mypy hooks.

pyproject.toml
kasium: The issue is, how should we type then? The typing module is only there in 3.6 and even typing-extensions are only 3.6. Therefore it makes no sense to check for python 2. Or did I miss something?

setup.cfg
kasium: okay, as soon as the other discussion is done, I'll do that

webknjaz: I know there was some python2 backport. `types`? Need to search for it but it's totally possible.

pyproject.toml
webknjaz: Use https://pypi.org/p/typing under Python 2.

kasium: Thanks, I'll try that

kasium:
So, running mypy will now fail:

```
# pre-commit run mypy -a
mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

cheroot/__init__.pyi: error: Duplicate module named "cheroot" (also at
"cheroot/__init__.py")
cheroot/__init__.pyi: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
Found 1 error in 1 file (errors prevented further checking)
```
Either we exclude the py files or we use stubtest to check them against the real files (https://stackoverflow.com/questions/51716200/how-do-you-check-if-a-typeshed-stub-pyi-file-matches-the-implementation)

.pre-commit-config.yaml
```diff
@@ -91,3 +91,10 @@ repos:
     - requests_unixsocket
     - six
     - Sphinx
+
```
webknjaz, line 94: plz follow the same style as the rest of the file — with a single line separator.
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -91,3 +91,10 @@ repos:
...
+
+-   repo: https://github.com/pre-commit/mirrors-mypy
```
webknjaz, line 96: Could you move this before `pylint`? It's faster and pylint is quite slow, this is why it needs to be the last one for best linting responsiveness.
Also, use the same indentation style as the entries above — with just one space after `-`, not 3.

webknjaz:
> So, running mypy will now fail:
> 
> ```
> # pre-commit run mypy -a
> mypy.....................................................................Failed
> - hook id: mypy
> - exit code: 2
> 
> cheroot/__init__.pyi: error: Duplicate module named "cheroot" (also at
> "cheroot/__init__.py")
> cheroot/__init__.pyi: note: Are you missing an __init__.py? Alternatively, consider using --exclude to avoid checking one of them.
> Found 1 error in 1 file (errors prevented further checking)
> ```
> 
> Either we exclude the py files or we use stubtest to check them against the real files ([stackoverflow.com/questions/51716200/how-do-you-check-if-a-typeshed-stub-pyi-file-matches-the-implementation](https://stackoverflow.com/questions/51716200/how-do-you-check-if-a-typeshed-stub-pyi-file-matches-the-implementation))

It's a side-effect of passing files as individual args to `mypy`. This way it works with separate contexts not realizing that the whole package should be treated as one whole.
The solution is to tell `pre-commit` not to pass those files but lint the whole thing. Here's a few more examples — https://github.com/sanitizers/octomachinery/blob/be18b54/.pre-commit-config.yaml#L60-L79 and https://github.com/abhinavsingh/proxy.py/blob/1eeed91/.pre-commit-config.yaml#L159-L183. Please copy-paste and edit the latter — it already has the style I want (no unnecessary quotes, right+consistent indentation, multiline sequences). But the key part is that `pass_filenames: false` combined with `-p <package name>`.
Once you do this, you can verify the check locally with something like `tox -e pre-commit -- mypy --all-files`.

When this is ready and works well, you'll need to duplicate the mypy entry (the hook part inside the repo) and invoke it with Python 2 while the first one would be invoked with Python 3.

P.S. You may keep `--install-types` but you'll still need to list them in `additional_dependencies` for pre-commit.ci integration to work because that env runs the checks in a disconnected from the network environment.

kasium:
> It's a side-effect of passing files as individual args to `mypy`. This way it works with separate contexts not realizing that the whole package should be treated as one whole. The solution is to tell `pre-commit` not to pass those files but lint the whole thing. Here's a few more examples — https://github.com/sanitizers/octomachinery/blob/be18b54/.pre-commit-config.yaml#L60-L79 and https://github.com/abhinavsingh/proxy.py/blob/1eeed91/.pre-commit-config.yaml#L159-L183. Please copy-paste and edit the latter — it already has the style I want (no unnecessary quotes, right+consistent indentation, multiline sequences). But the key part is that `pass_filenames: false` combined with `-p <package name>`. Once you do this, you can verify the check locally with something like `tox -e pre-commit -- mypy --all-files`.

done

> When this is ready and works well, you'll need to duplicate the mypy entry (the hook part inside the repo) and invoke it with Python 2 while the first one would be invoked with Python 3.

Sorry, I don't get this point

> P.S. You may keep `--install-types` but you'll still need to list them in `additional_dependencies` for pre-commit.ci integration to work because that env runs the checks in a disconnected from the network environment.

I added what's possible but for some dependencies there are no types...

.pre-commit-config.yaml
```diff
@@ -75,6 +75,20 @@ repos:
...
+  - id: mypy
+    additional_dependencies:
+    - typing
+    - types-six
+    args:
+    - --ignore-missing-imports
+    - --install-types
+    - --non-interactive
+    - cheroot/
+    pass_filenames: false
```
webknjaz, line 90: > Sorry, I don't get this point

I meant something like

```suggestion
- repo: https://github.com/pre-commit/mirrors-mypy
  rev: v0.910-1
  hooks:
  - id: mypy
    additional_dependencies:
    - typing
    - types-six
    args:
    - --ignore-missing-imports
    - --install-types
    - --non-interactive
    - --python-version
    - 3.5
    - cheroot/
    pass_filenames: false
- repo: https://github.com/pre-commit/mirrors-mypy
  rev: v0.910-1
  hooks:
  - id: mypy
    additional_dependencies:
    - typing
    - types-six
    args:
    - --ignore-missing-imports
    - --install-types
    - --non-interactive
    - --python-version
    - 2.7
    - cheroot/
    pass_filenames: false
```

mypy.ini
```diff
@@ -0,0 +1,12 @@
...
+pretty = true
+show_column_numbers = true
+show_error_codes = true
+warn_no_return = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
+ignore_missing_imports = true
+
+[mypy-cheroot.test.*]
```
webknjaz, line 11: Why is this ignored?

kasium: valid point

kasium:
Okay, so mypy run now fails with

```
cheroot/_compat.py:16:5: error: Name "selectors" already defined (by an import)
 [no-redef]
        import selectors2 as selectors  # noqa: F401  # lgtm [py/unused-im...
        ^
cheroot/_compat.py:23:26: error: Incompatible types in assignment (expression
has type "None", variable has type "bool")  [assignment]
        IS_ABOVE_OPENSSL10 = None
                             ^
cheroot/_compat.py:31:6: error: Name "suppress" already defined (possibly by an
import)  [no-redef]
        @contextmanager
         ^
cheroot/test/webtest.py:583:1: error: Need type comment for
"ignored_exceptions" (hint: "ignored_exceptions = ...  # type: List[<type>]") 
[var-annotated]
    ignored_exceptions = []
    ^
cheroot/makefile.py:13:5: error: Name "io" already defined (possibly by an
import)  [no-redef]
        import io
        ^
cheroot/makefile.py:51:20: error: Unsupported dynamic base class "getattr" 
[misc]
    class MakeFile_PY2(getattr(socket, '_fileobject', object)):
                       ^
cheroot/connections.py:23:9: error: Module "ctypes" has no attribute "windll" 
[attr-defined]
            from ctypes import windll, WinError
            ^
cheroot/connections.py:23:9: error: Module "ctypes" has no attribute "WinError"
 [attr-defined]
            from ctypes import windll, WinError
            ^
cheroot/server.py:87:5: error: Name "lru_cache" already defined (possibly by an
import)  [no-redef]
        from backports.functools_lru_cache import lru_cache
        ^
cheroot/server.py:136:20: error: Incompatible types in assignment (expression
has type "None", variable has type Module)  [assignment]
            grp, pwd = None, None
                       ^
cheroot/server.py:183:5: error: Module has no attribute "statistics" 
[attr-defined]
        logging.statistics = {}
        ^
cheroot/server.py:644:5: error: Need type comment for "inheaders" (hint:
"inheaders = ...  # type: Dict[<type>, <type>]")  [var-annotated]
        inheaders = {}
        ^
cheroot/server.py:647:5: error: Need type comment for "outheaders" (hint:
"outheaders = ...  # type: List[<type>]")  [var-annotated]
        outheaders = []
        ^
cheroot/server.py:2076:6: error: Name "interrupt" already defined on line 2066 
[no-redef]
        @interrupt.setter
         ^
cheroot/server.py:2076:6: error: "Callable[[HTTPServer], Any]" has no attribute
"setter"  [attr-defined]
        @interrupt.setter
         ^
cheroot/wsgi.py:329:15: error: Incompatible types in assignment (expression has
type "Tuple[str, int]", base class "Gateway_10" defined the type as
"Tuple[int, int]")  [assignment]
        version = 'u', 0
                  ^
cheroot/ssl/pyopenssl.py:71:9: error: Cannot assign multiple types to name
"ssl_conn_type" without an explicit "Type[...]" annotation  [misc]
            ssl_conn_type = SSL.ConnectionType
            ^
cheroot/ssl/pyopenssl.py:71:25: error: Module has no attribute
"ConnectionType"; maybe "Connection"?  [attr-defined]
            ssl_conn_type = SSL.ConnectionType
                            ^
cheroot/ssl/pyopenssl.py:73:11: error: Incompatible types in assignment
(expression has type "None", variable has type Module)  [assignment]
        SSL = None
              ^
cheroot/ssl/pyopenssl.py:232:20: error: Metaclasses not inheriting from "type"
are not supported  [misc]
    @six.add_metaclass(SSLConnectionProxyMeta)
                       ^
cheroot/ssl/builtin.py:20:11: error: Incompatible types in assignment
(expression has type "None", variable has type Module)  [assignment]
        ssl = None
              ^
cheroot/test/test_ssl.py:50:16: error: Argument 1 to "startswith" of "str" has
incompatible type "bytes"; expected "Union[str, Tuple[str, ...]]"  [arg-type]
        startswith(b'OpenSSL 1.0.')
                   ^
cheroot/test/test_core.py:66:16: error: Argument 1 to "_munge" has incompatible
type "str"; expected "HelloController"  [arg-type]
            _munge('/привіт'): hello,
                   ^
cheroot/test/test_core.py:67:16: error: Argument 1 to "_munge" has incompatible
type "str"; expected "HelloController"  [arg-type]
            _munge('/Юххууу'): hello,
                   ^
Found 24 errors in 10 files (checked 30 source files)

mypy.....................................................................Failed
- hook id: mypy
- exit code: 2

mypy: can't decode file 'cheroot/test/test_core.py': 'ascii' codec can't decode byte 0xd0 in position 1799: ordinal not in range(128)
```

I can try to fix them, but the decode error is weird...

kasium:
@webknjaz okay, config is now ready. However, do you really want to check the .py files if there is no pyi file?

webknjaz:
> @webknjaz okay, config is now ready. However, do you really want to check the .py files if there is no pyi file?

I think so. It should ease the further mypy adoption, right?

> What do you think about just to add all missing types (with first a lot of Any's) by using mypy stubgen?

You could try that.

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
   hooks:
   - id: forbid-html-img-without-alt-text
 
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: 'v0.910-1'
```
webknjaz, line 79:
```suggestion
  rev: v0.910-1
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+    args:
+    - --install-types
+    - --non-interactive
+    - --python-version
+    - '3.5'
+    - cheroot/
+    pass_filenames: false
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: 'v0.910-1'
```
webknjaz, line 96:
```suggestion
  rev: v0.910-1
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+    - types-pyOpenSSL
+    - types-requests
+    args:
+    - --install-types
+    - --non-interactive
+    - --python-version
+    - '3.5'
+    - cheroot/
+    pass_filenames: false
+
```
webknjaz, line 94: These two invocations are related so I wouldn't have any spacing between them.
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: 'v0.910-1'
+  hooks:
```
webknjaz, line 97: I just realized that two separate repo entries are unnecessary, just two separate hook entries needed.
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+    additional_dependencies:
+    - types-six
+    - types-backports
+    - types-pyOpenSSL
+    - types-requests
+    args:
+    - --install-types
+    - --non-interactive
+    - --python-version
+    - '3.5'
```
webknjaz, line 91: I wonder if this would work:
```suggestion
    - --python-version=3.5
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: 'v0.910-1'
+  hooks:
+  - id: mypy
+    additional_dependencies:
+    - typing
+    - types-six
+    - types-backports
+    - types-pyOpenSSL
+    - types-requests
```
webknjaz, line 104: Let's sort these
```suggestion
    - types-backports
    - types-pyOpenSSL
    - types-six
    - types-requests
    - typing
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
 
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: 'v0.910-1'
+  hooks:
+  - id: mypy
+    additional_dependencies:
+    - types-six
+    - types-backports
+    - types-pyOpenSSL
+    - types-requests
```
webknjaz, line 86: same sorting
```suggestion
    - types-backports
    - types-pyOpenSSL
    - types-requests
    - types-six
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,41 @@ repos:
...
+    args:
+    - --install-types
```
webknjaz, line 88: Can this option be moved to the config?

webknjaz:
Re: `type[no-redef]` — this is a known bug in mypy with try/except import fallbacks style. It can be sorted by replacing that with if/else. There is a bug on their tracker, I don't remember the exact number.

.pre-commit-config.yaml
kasium: nope, install types is CLI only

mypy.ini
```diff
@@ -0,0 +1,10 @@
+[mypy]
+pretty = true
+show_column_numbers = true
+show_error_codes = true
+warn_no_return = true
+warn_redundant_casts = true
+warn_unused_configs = true
+warn_unused_ignores = true
+ignore_missing_imports = true
```
webknjaz, line 9: This will need to be addressed (maybe in a separate PR)

mypy.ini
```diff
@@ -0,0 +1,10 @@
...
+exclude = cheroot/test
```
webknjaz, line 10: Why is this excluded? Linting tests is as important as linting the rest of the code.

.pre-commit-config.yaml
```diff
@@ -75,6 +75,35 @@ repos:
...
+    - types-six
+    args:
+    - --install-types
+    - --non-interactive
+    - --python-version=3.5
+    - cheroot/
+    pass_filenames: false
+  - id: mypy
+    additional_dependencies:
+    - types-backports
```
webknjaz, line 95: The error message at https://results.pre-commit.ci/run/github/16620627/1638950716.FddrDhgSRnSAcxDS8WIZng says that this is necessary under Python 2:
```suggestion
    - typed-ast
    - types-backports
```

webknjaz:
@kasium thanks again for working on this!
It seems like there are a few failures that are visible at https://results.pre-commit.ci/run/github/16620627/1638950716.FddrDhgSRnSAcxDS8WIZng. I mentioned earlier (https://github.com/cherrypy/cheroot/pull/385#issuecomment-988225111) that the resolution of one of them is changing try/except to if/else. I haven't checked closely but the others may need type-ignores added (unless there's a way to come up with a proper solution, that is).

Also, I wanted to ask if you were planning to apply interactive rebase to your PR branch when you're done to clean it up a bit? If not, I'll likely do this myself before merging. Please tell me what your preference is.

mypy.ini
kasium: I was not aware of that. Not everybody likes types tests...

mypy.ini
kasium: Makes sense

kasium:
@webknjaz I'll take a look into the issue. If it's fine´, I'll leave the rebase up tp you (however, a squash also helps).
One other thing: Currently mypy just checks the stubs, but not the stubs against the code base (stubtest). I guess we should include that, right?

webknjaz:
Yes, it would be great to include that type of testing as well. I just want to be conscious about the PR size and the scope creep. It's probably easier to have a series of small PRs rather than a big one — it's easier to digest from the reviewer's perspective and easier to merge complete changes faster rather than having some finished parts of the change wait for one unfinished one. It'd also help for GHA to run automatically on your PRs (GitHub requires manual approval from me for the first-time contributor PRs but post-merge, the new PRs won't need that).

mypy.ini
kasium: @webknjaz are you sure, that we should add stubs for tests? That sounds weird to me

kasium:
@webknjaz I hope, I now solved the missing pieces. I guess it makes sense to have a general issue with some tasks

- Add typing to _compat
- Add typing to tests (?)
- Activate "warn_usused_ignore" (needed because of differences in py2 and py3)
- Add stubtest
- Add py.typed file and add it to the distribution
- Improve the stubs (add missing types and remove Any's)

This all can be done in separate PRs

mypy.ini
webknjaz: Usually, the whole project needs to be tested. But maybe you're right — it's probably more useful to have in-source annotations for tests rather than in separate files.

.pre-commit-config.yaml
```diff
@@ -75,6 +75,36 @@ repos:
   hooks:
   - id: forbid-html-img-without-alt-text
 
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v0.910-1
+  hooks:
+  - id: mypy
```
webknjaz, line 81:
```suggestion
  - id: mypy
    name: MyPy, for Python 3.5
```

.pre-commit-config.yaml
```diff
@@ -75,6 +75,36 @@ repos:
...
+    - types-pyOpenSSL
+    - types-requests
+    - types-six
+    args:
+    - --install-types
+    - --non-interactive
+    - --python-version=3.5
+    - cheroot/
+    pass_filenames: false
+  - id: mypy
```
webknjaz, line 93:
```suggestion
  - id: mypy
    name: MyPy, for Python 2.7
```

webknjaz (approved):
Great job! I'll merge this soon.

As for the tracking issue, that sounds reasonable, feel free to file it.

webknjaz:
Looks like GHA has broken one of their images so I'll have to merge without green CI.

PR #500, opened 2022-03-03
Use importlib.metadata instead of deprecated pkg_resources

wylfen:
pkg_resources is deprecated upstream and does not handle modern .dist-info directories correctly. Specifically it does not honor the [new replacement rules](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode). As a result, it can't find some of cheroot's dependencies on modern Gentoo systems and ultimately [fails to load](https://bugs.gentoo.org/834522) with a DistributionNotFound error.

pkg_resources is only used to get the cheroot version and can easily be replaced with [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html), the intended replacement for this specific use-case. For python versions before 3.8, attempt to use the importlib_metadata backport instead.

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

📋 **Contribution checklist:**

I do not know how to add tests for this specific failure case, so I omitted them.

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/500)

### DIFF
--- a/cheroot/__init__.py
+++ b/cheroot/__init__.py
@@ -4,12 +4,15 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 try:
-    import pkg_resources
+    from importlib.metadata import version
 except ImportError:
-    pass
+    try:
+        from importlib_metadata import version
+    except ImportError:
+        pass
 
 
 try:
-    __version__ = pkg_resources.get_distribution('cheroot').version
+    __version__ = version('cheroot')
 except Exception:
     __version__ = 'unknown'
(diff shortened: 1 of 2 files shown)


wylfen:
Let me know what to do about the nested try block, flake8 didn't trigger on my local run of pre-commit.

jaraco (approved):
I've started recommending against supplying a `__version__` at all and instead directing consumers to query for `importlib.metadata.version("cheroot")` if they want to know the version. However, for compatibility, it's probably better to go with this approach for now.

PR #505, opened 2022-05-02
Make it possible to set SO_REUSEPORT to the server's socket

### LINKED ISSUE #504: Add an option for port re-using
❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [X] 🐣 feature request

We have a process that automatically distributes available ports across multiple cheroot processes and keeps sockets on those ports opened until cheroot processes take control on them. Cheroot right now cannot reuse those opened ports.

🐣 **Describe the solution you'd like**
Add an option to say cheroot to set SO_REUSEPORT socket option.

📋 **Describe alternatives you've considered**
Passing socket object to cheroot instead of bind address, but that would require some refactoring and I don't think it would follow Zen of Cherrypy if this solution would be ported to cherrypy.

VladimirKuzmin:

* [ ] 🐞 bug fix
* [x] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves #504

Cheroot cannot be used in infrastructure where bind address for it allocated and kept by some different process.

*main.py:*
```
s = socket.socket(family)
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
s.bind(('127.0.0.1', 1234))
subprocess.run(['python3', 'server.py', '127.0.0.1', '1234'])
```
*server.py:*
```
addr = (argv[0], int(argv[1]))
server = wsgi.Server(addr, wsgi_app)
server.start()  # fails, socket already in use
```

Cheroot can optionally re-use port:

*main.py:*
```
s = socket.socket(family)
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
s.bind(('127.0.0.1', 1234))
subprocess.run(['python3', 'server.py', '127.0.0.1', '1234'])
```
*server.py:*
```
addr = (argv[0], int(argv[1]))
server = wsgi.Server(addr, wsgi_app, reuse_port=True)  # a new option reuse_port
server.start()  # works, SO_REUSEPORT has been set
```

📋 **Contribution checklist:**

[related squash]:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/505)


cheroot/test/test_server.py
```diff
@@ -363,6 +363,33 @@ def native_process_conn(conn):
...
+    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+    s.bind((ip_addr, EPHEMERAL_PORT))
+    server = HTTPServer(
+        bind_addr=s.getsockname()[:2], gateway=Gateway, reuse_port=True,
+    )
+    try:
+        server.prepare()
+    finally:
+        server.stop()
+        s.close()
```
webknjaz, line 390: There should be an assertion of some kind here. Maybe with a `mocker.spy()` or something like that.

webknjaz (commented):
I'm hesitant to accept this as is right now. I wasn't aware of this option, and it'll need to be researched closer so that the PR is reviewed consciously. But I'm leaving a few notes for the future reference.

One thought is whether we really need this to be toggleable. `SO_REUSEADDR` cannot be switched on or off, for example. Why would it be any different for `SO_REUSEPORT` which is essentially a feature of the same category?
Are there any cases when it'd be undesired to have `SO_REUSEADDR` on by default?

Some googling suggests that it may be a good idea to have both of these options set on the socket simultaneously.

Extra reading for maintainers:
* https://www.unixguide.net/network/socketfaq/4.11.shtml
* https://lwn.net/Articles/542629/
* https://stackoverflow.com/a/14388707/595220

Extra questions:
* How does this affect the ephemeral sockets?
* How does this impact the abstract sockets?

VladimirKuzmin:
In my opinion here is the key logical difference between `SO_REUSEADDR` and `SO_REUSEPORT` (with exception to Windows):
* `SO_REUSEADDR` is used to resolve conflict with some unrelated processes, and can be set by a process listening to non-wildcard addresses;
* `SO_REUSEPORT` is used as an tool for system design (load balancing, assigning ports by some coordinating process, picking up port from another stuck process) and it's set intentionally by every process using the address.

Turning it on by default can lead to a situation when two independent processes (e.g two different cherrypy/cheroot servers) accidentally open connections to the same port and address, set `SO_REUSEPORT` just in case and get only part of their requests. It might be a rare case, but it would be hard to debug. So I am still inclined to use the extra parameter

I will add better handling of some special cases:
* Windows can use `SO_REUSEADDR` instead of `SO_REUSEPORT`;
* Ephemeral socket should not be used together with reuse_port, will add a check or switch off `SO_REUSEPORT` if we go with the solution without the extra parameter;
* Will do research on abstract sockets.

VladimirKuzmin:
Sorry for the long delay.  I updated my PR and resolved previously discussed issues.

Also, here are some examples of port reuse implementations in Python web servers:
https://github.com/benoitc/gunicorn/blob/20.1.0/gunicorn/sock.py#L41
https://github.com/unbit/uwsgi/blob/2.0.21/core/socket.c#L699
https://github.com/jonashaag/bjoern/blob/3.2.2/bjoern.py#L24

Alternatively, if this feature is too expensive to maintain, I can send PR making `prepare_socket` a class method instead of a static method, so that method can be overridden.

cheroot/test/test_server.py
VladimirKuzmin: Thank you for the hint, done!

cheroot/server.py
```diff
@@ -1968,14 +1977,41 @@ def bind_unix_socket(self, bind_addr):  # noqa: C901  # FIXME
...
+            socket_.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        else:
+            raise NotImplementedError(
+                'Current platform does not support port reuse',
+            )
+
+    @classmethod
+    def prepare_socket(
+            cls, bind_addr, family, type, proto, nodelay, ssl_adapter,
+            reuse_port=False,
```
jaraco, line 1999: I don't love how much plumbing was required for this change. It's also technically backward-incompatible in the case where a subclass has overridden `prepare_socket` but isn't accepting arbitrary `args` and `kwargs`.

That's not your fault, of course, but a problem with the current conventions/approach.

I'm tempted to say we should first (or after) refactor to make `prepare_socket` an instance method and remove the need to pass instance attributes through to it.

I'll plan to deal with that separately.

cheroot/server.py
```diff
@@ -1968,14 +1977,41 @@ def bind_unix_socket(self, bind_addr):  # noqa: C901  # FIXME
...
         connections.prevent_socket_inheritance(sock)
 
         host, port = bind_addr[:2]
         IS_EPHEMERAL_PORT = port == 0
 
+        if reuse_port:
+            if sock.family not in (socket.AF_INET, socket.AF_INET6):
+                raise ValueError('Cannot reuse a non-IP socket')
+            if IS_EPHEMERAL_PORT:
+                raise ValueError('Cannot reuse an ephemeral port (0)')
```
jaraco, line 2012: I think I'd like to see this code moved to `_make_socket_reusable`, something like this:

```diff
diff --git a/cheroot/server.py b/cheroot/server.py
index 4ffeb69c..16fd9ac6 100644
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1977,7 +1977,16 @@ class HTTPServer:
         return sock
 
     @staticmethod
-    def _make_socket_reusable(socket_):
+    def _make_socket_reusable(socket_, bind_addr):
+        host, port = bind_addr[:2]
+        IS_EPHEMERAL_PORT = port == 0
+
+        if socket_.family not in (socket.AF_INET, socket.AF_INET6):
+            raise ValueError('Cannot reuse a non-IP socket')
+
+        if IS_EPHEMERAL_PORT:
+            raise ValueError('Cannot reuse an ephemeral port (0)')
+
         # Most BSD kernels implement SO_REUSEPORT the way that only the
         # latest listener can read from socket. Some of BSD kernels also
         # have SO_REUSEPORT_LB that works similarly to SO_REUSEPORT
@@ -2006,11 +2015,7 @@ class HTTPServer:
         IS_EPHEMERAL_PORT = port == 0
 
         if reuse_port:
-            if sock.family not in (socket.AF_INET, socket.AF_INET6):
-                raise ValueError('Cannot reuse a non-IP socket')
-            if IS_EPHEMERAL_PORT:
-                raise ValueError('Cannot reuse an ephemeral port (0)')
-            cls._make_socket_reusable(sock)
+            cls._make_socket_reusable(bind_addr, sock)
 
         if not (IS_WINDOWS or IS_EPHEMERAL_PORT):
             """Enable SO_REUSEADDR for the current socket.
```

I know there's some repetition around `IS_EPHEMERAL_PORT`, but that seems preferable to having the re-use port logic in two places.

WDYT?

jaraco (commented):
This looks good to me. It seems to follow similar conventions of other servers and unblocks an important use case.

I find it a little awkward that `REUSE_ADDR` is now used in two different locations (under re-use port for Windows and by default for non-ephemeral addresses on other platforms). 

> * Windows can use `SO_REUSEADDR` instead of `SO_REUSEPORT`;

I'm not sure I understand this motivation. Can you provide a reference for why `SO_REUSEADDR` is appropriate on Windows when the user is requesting `SO_REUSEPORT`?

VladimirKuzmin:
> I'm not sure I understand this motivation. Can you provide a reference for why SO_REUSEADDR is appropriate on Windows when the user is requesting SO_REUSEPORT?

Historically SO_REUSEADDR in Windows [works](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse) as a combination of SO_REUSEADDR and SO_REUSEPORT. It's the reason this flag is skipped for Windows in [IS_EPHEMERAL_PORT section](https://github.com/cherrypy/cheroot/blob/580e296b0cf201f02523e539ff1b200773ccfc36/cheroot/server.py#L1979).

At the same time Windows has a flag SO_EXCLUSIVEADDRUSE which apparently is ["anti-SO_REUSEPORT"](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse#using-so_exclusiveaddruse). That special case can probably be handled in the IS_EPHEMERAL_PORT section, but that would spread reuse port logic even more:
```python
    if not IS_EPHEMERAL_PORT:
        """Enable SO_REUSEADDR for the current socket.
        Skip for Windows (has different semantics)
        or ephemeral ports (can steal ports from others).
        Refs:
        * https://msdn.microsoft.com/en-us/library/ms740621(v=vs.85).aspx
        * https://github.com/cherrypy/cheroot/issues/114
        * https://gavv.github.io/blog/ephemeral-port-reuse/
        """
        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        if IS_WINDOWS and not reuse_port and hasattr(socket, "SO_EXCLUSIVEADDRUSE"):
            sock.setsockopt(socket.SOL_SOCKET, socket.SO_EXCLUSIVEADDRUSE, 1)
```

jaraco (approved):

jaraco:
```
Commit 081aa090ac:
3: B6 Body message is missing

Commit 78d59c73d6:
3: B6 Body message is missing
```

I'm not fixing that.

PR #507, opened 2022-05-13
__init__: generate correct version number

### LINKED ISSUE #506: BUG: incorrect version number generated in distributions
Debian and Ubuntu user package versions like 8.6.0-ds1 which are fine for *.deb packages but not legal according to PEP440.

In cheroot/__init__.py:13: the Debian package number is misused as Python package __version__.

```
 __version__ = pkg_resources.get_distribution('cheroot').version
```

This leads to errors like the following

```
$ python3
>>> import cheroot
>>> print cheroot.__version__
8.6.0+ds1
>>> from distutils.version import StrictVersion
>>> StrictVersion(cheroot.__version__)
  File "/usr/lib/python3.10/distutils/version.py", line 40, in __init__
    self.parse(vstring)
  File "/usr/lib/python3.10/distutils/version.py", line 137, in parse
    raise ValueError("invalid version number '%s'" % vstring)
ValueError: invalid version number '8.6.0+ds1'
>>>
```

It would be preferable to sanitize the string returned by pkg_resources.get_distribution('cheroot').version.

Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010894
Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/ceph/+bug/1967139

Best regards

Heinrich

xypron:
Closes #506

The distribution package number is reused as Python package version.

This leads to errors in distutils.version.StrictVersion like

    ValueError: invalid version number '8.6.0+ds1'

Use a regular expression to extract a valid version number.

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/507)

### DIFF
--- a/cheroot/__init__.py
+++ b/cheroot/__init__.py
@@ -1,6 +1,7 @@
 """High-performance, pure-Python HTTP server used by CherryPy."""
 
 from __future__ import absolute_import, division, print_function
+import re
 __metaclass__ = type
 
 try:
@@ -10,6 +11,9 @@ except ImportError:
 
 
 try:
-    __version__ = pkg_resources.get_distribution('cheroot').version
+    __version__ = re.search(
+        r'(?x)^(\d+)\.(\d+)(\.(\d+))?([ab](\d+))?',
+        pkg_resources.get_distribution('cheroot').version,
+    ).expand(r'\1.\2\3\5')
 except Exception:
     __version__ = 'unknown'


wolsen (commented):
I might suggest instead using the packaging.version from the setuptools module. It's already PEP 440 compliant and tolerates the local deviations from downstream distros and such following this model. I think the public version will give you what you want here.

e.g.

```console
$ python3
Python 3.10.4 (main, Apr  2 2022, 09:04:19) [GCC 11.2.0] on linux
>>> from packaging import version
>>> v = version.parse('8.6.0+ds1')
>>> v.public
'8.6.0'
>>> v.local
'ds1'
>>> v
<Version('8.6.0+ds1')>
>>> 
```

PR #516, opened 2022-07-08
Deduplicate `ThreadPool` worker thread creation

pR0Ps:

* [ ] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [x] 📋 refactoring
* [ ] 💥 other

N/A

The `ThreadPool.start` and `ThreadPool.grow` functions both add threads to the pool, but both implement their own version that is similar, but not the same.

This commit refactors the `ThreadPool.start` function to use `ThreadPool.grow` to create the initial workers, reducing code duplication. It also adds some error checking that was not previously there (and some unit tests to exercise them)

📋 **Other information**:
In addition to the slight refactor, this PR used to fix a bug where due to a trailing comma, the names of the `ThreadPool`'s `WorkerThread`s were being set to the tuple `("CP Server Thread-<id>",)` instead of just a string. #503 was merged so this PR was updated to only include the refactor.

📋 **Contribution checklist:**

* [x] I have [squashed related commits together][related squash] after the changes have been approved
* [x] The PR relates to *only* one subject with a clear title and description in grammatically correct, complete sentences

[related squash]:

---
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/516)


pR0Ps:
Sorry, just noticed that the primary issue this PR fixes already had a PR submitted for it (#503). This PR also tidies up the `ThreadPool` initialization code though so it's not a full duplicate.

webknjaz:
I merged #503 because it's atomic and doesn't mix up refactoring with bugfixing. I've then rebased this branch and will be looking at the refactoring part separately.

cheroot/workers/threadpool.py
```diff
@@ -173,17 +173,7 @@ def __init__(
...
-        for worker in self._threads:
-            worker.name = (
-                'CP Server {worker_name!s}'.
-                format(worker_name=worker.name)
-            )
-            worker.start()
-        for worker in self._threads:
-            while not worker.ready:
-                time.sleep(.1)
+        self.grow(self.min)
```
webknjaz, line 176: `grow()`'s logic doesn't always spawn the same amount of threads as the current implementation.
With the default `max = -1`, it calculates the same amount of threads as the current logic. But if one were to set like `max = 2`, then the `budget` would be calculated as `2` and not `min = 10` (the current default).

So the difference is that the current implementation spawns exactly the amount of threads requested by the `min` argument, while this PR's implementation may spawn less in case of `max < min`.

cheroot/workers/threadpool.py
```diff
@@ -173,17 +173,7 @@ def __init__(
...
-            self._threads.append(WorkerThread(self.server))
-        for worker in self._threads:
-            worker.name = (
-                'CP Server {worker_name!s}'.
-                format(worker_name=worker.name)
-            )
-            worker.start()
-        for worker in self._threads:
-            while not worker.ready:
-                time.sleep(.1)
```
webknjaz, line 186: The waiting implementation here, in the original code, seems a bit more efficient, while `grow()` has more looping over the data as it seems.

Have you compared the performance of both?

webknjaz (dismissed):
🙏 Thank you for your PR!

ℹ️ I've noticed that there are some functional differences being introduced by this refactoring attempt. I've added more details in the comments below.

✨ Currently, I'm not sure if it's acceptable to have these changes as a part of this PR, but if only spawning and waiting logic was put into a helper method, it'd be easier to accept it. Plus, performance-wise, we'd need to select one of the implementations that is faster.

❓ Would you mind working on addressing these concerns?

cheroot/workers/threadpool.py
pR0Ps: I guess the real question is what should the code do if the user runs something like `ThreadPool(max=2)`?
 - Spawn 10 threads (current)
 - Spawn 2 threads (this PR)
 - `raise ValueError("min of 10 cannot be greater than max of 2")`

cheroot/workers/threadpool.py
pR0Ps: I have not compared the performance between using the original code from `start()` and the code from `grow()` but have some thoughts:
 - Both the current implementation and `grow()` wait for all the threads to be started and ready with a spinlock delay of 100ms. Compared with variance introduced by that, I would imagine that any difference in performance of looping over a datastructure of this size a few times to be negligible. With that being said, I've changed `grow()` to use the structure from `start()` since I agree with you that it should be faster.
 - This code is run in the `__init__` of a `ThreadPool` which is only initialized twice in the entire codebase, both times in the `__init__` of long-lived objects (HTTP servers). It's not on the critical path or I would've paid more attention to micro-optimizing it.

cheroot/workers/threadpool.py
```diff
@@ -173,17 +173,7 @@ def __init__(
 
     def start(self):
         """Start the pool of threads."""
-        for _ in range(self.min):
-            self._threads.append(WorkerThread(self.server))
```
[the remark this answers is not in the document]
pR0Ps: I've gone over it a few times and I can't see any issues with changing the order. In fact, not ever having unstarted workers in `self._threads` strikes me as a more atomic implementation. The threads are there and running or not at all.

With that being said, `start()` shouldn't ever be called multiple times concurrently so adding then starting should be functionally the same as starting then adding in this case.

cheroot/workers/threadpool.py
webknjaz: That's an interesting observation too. Usually, in such cases I go for Git archeology (inspecting what added this behavior / commits / blame / motivation). Do you think you could look into it and attempt to deduce what the original motivation was?

cheroot/workers/threadpool.py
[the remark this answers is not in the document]
webknjaz: But what if starting a portion of those pre-allocated threads succeeds while others fail, causing the exception to bubble up the stack leaving the started threads running (hence consuming some system resources) without any references to them?

cheroot/workers/threadpool.py
webknjaz: Okay, that sounds reasonable. Marking this thread as resolved.

cheroot/workers/threadpool.py
webknjaz: Following up on "which behavior of double-start is more correct", I think it could follow stdlib's Thread behavior and throw a runtime error:
```python-traceback
$ python 
Python 3.11.0rc2 (main, Sep 24 2022, 17:37:14) [GCC 11.2.1 20220115] on linux
>>> from threading import Thread
>>> thread = Thread()
>>> thread.start()
>>> thread.start()
  File "~/.pyenv/versions/3.11.0rc2/lib/python3.11/threading.py", line 952, in start
    raise RuntimeError("threads can only be started once")
RuntimeError: threads can only be started once
```

And re:input validation — it sounds like throwing an exception is the right thing to do here as well. But in an attempt to avoid https://xkcd.com/1172/ I'd like to understand how this was being designed originally.

pR0Ps: Did some digging:

The original implementation of the class in e8127b98 would not raise any errors when calling `start()` twice and just add `self.min` more threads each time. `.grow()` would refuse to add more than `.max` threads. `.start()` would add all the threads to `._threads`, start them all, then wait for them all to be ready. `.grow()` would create the new thread, add it to `._threads`, then start it without waiting for it to be ready.

47844dab changed `.grow()` to spawn all the workers before adding them to `._threads`, still not waiting for them to be ready. I couldn't figure out why this was changed.

1d70c4b1 changed `.grow()` to wait for threads to become ready before adding them to `._threads`. The commit message mentioned it fixed issue #1205 which seems to be a reference to an old bug tracker.

cheroot/workers/threadpool.py
```diff
@@ -157,6 +157,11 @@ def __init__(
             accepted_queue_timeout (int): timeout for putting request
                 into queue
         """
+        if min < 1:
+            raise ValueError('min must be > 0')
```
webknjaz, line 161: It'd be nice to add tests for the invalid inputs too. Could even publish all the input validation additions as a separate PR, optionally.

webknjaz:
@pR0Ps could you address those last two comments? I think they are the only ones necessary for this to be mergable.

cheroot/workers/threadpool.py
pR0Ps: tests added

cheroot/workers/threadpool.py
```diff
@@ -157,6 +157,11 @@ def __init__(
...
+        if max > 0 and min < max:
```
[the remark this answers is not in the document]
pR0Ps: I ended up changing it slightly (to `-1 < max < min`) after writing the input validation tests

pR0Ps:
@webknjaz I addressed the last 2 comments, let me know if there's anything else you'd like added/changed

cheroot/test/test_server.py
```diff
@@ -439,3 +441,52 @@ def many_open_sockets(request, resource_limit):
...
+        (2, 2, None),
+    ),
+)
+def test_threadpool_threadnum_validation(minthreads, maxthreads, error):
+    kwargs = dict(
+        server=None,
+        min=minthreads,
+        max=maxthreads,
+    )
+    if error is None:
```
webknjaz, line 473: The tests shouldn't be branchy, logic in tests should be avoided as much as possible.
Besides, I see two distinct tests here — negative and positive. This is a clear hint that they should reside in separate test functions with separate names clearly stating what they are testing — this is also useful in the test report output.

cheroot/test/test_server.py
```diff
@@ -439,3 +441,52 @@ def many_open_sockets(request, resource_limit):
...
+        (1, 1, None),
+        (1, 2, None),
+        (2, -1, None),
+        (2, 0, ValueError),
+        (2, 1, ValueError),
+        (2, 2, None),
+    ),
+)
+def test_threadpool_threadnum_validation(minthreads, maxthreads, error):
+    kwargs = dict(
```
webknjaz, line 468: I think this extra var adds mental overhead because it is detached from a single tested initializer invocation.

cheroot/test/test_server.py
```diff
@@ -439,3 +441,52 @@ def many_open_sockets(request, resource_limit):
...
+        (0, 2, ValueError),
+        (1, -1, None),
+        (1, 0, ValueError),
+        (1, 1, None),
+        (1, 2, None),
+        (2, -1, None),
+        (2, 0, ValueError),
+        (2, 1, ValueError),
+        (2, 2, None),
+    ),
```
webknjaz, line 465: Could you also add an `ids` function for creating a nicer representation of each test factor in the pytest output?

Maybe, something like this would do:
```suggestion
    ),
    ids=lambda *params: '...'.join(params[:2]),
```

cheroot/test/test_server.py
```diff
@@ -439,3 +441,52 @@ def many_open_sockets(request, resource_limit):
...
+def test_threadpool_threadnum_validation(minthreads, maxthreads, error):
+    kwargs = dict(
+        server=None,
+        min=minthreads,
+        max=maxthreads,
+    )
+    if error is None:
+        ThreadPool(**kwargs)
+    else:
+        with pytest.raises(error):
```
webknjaz, line 476: `pytest.raises()` should always have a `match='^.........$'` argument passed.

webknjaz:
@pR0Ps looks like it's also time to update the PR description, removing mentions of the changes that are no longer included.

cheroot/test/test_server.py
pR0Ps: I pulled the test apart into 2 tests - one for the errors (`test_threadpool_invalid_threadrange`), and one that ensures that valid min/max values are properly set (`test_threadpool_threadrange_set`). This removed the branching, the extra variable, and the need for an `ids` function.

cheroot/test/test_server.py
pR0Ps: see other comment

cheroot/test/test_server.py
pR0Ps: see other comment

cheroot/test/test_server.py
```diff
@@ -439,3 +441,52 @@ def many_open_sockets(request, resource_limit):
...
+            ThreadPool(**kwargs)
+
+
+def test_threadpool_multistart_validation(monkeypatch):
```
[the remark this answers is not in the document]
pR0Ps: done!

cheroot/test/test_server.py
webknjaz: It's still relevant because pytest joins parameter segments with dashes. `test_name[1...-2]` reads as a range, while `test_name[1--2]` is ambiguous/confusing.

cheroot/test/test_server.py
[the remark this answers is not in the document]
webknjaz: FYI the linters check for PEP 257.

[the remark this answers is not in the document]
pR0Ps: I think I fixed it now with the closing quote on a new line

cheroot/test/test_server.py
pR0Ps: I had another look and I don't think this is possible. It looks like `ids` is called once per value, not once per test. As a test, I ran the tests with `ids=lambda *a, **kw: "{} {}".format(a, kw)` and it generated names like `test_threadpool_invalid_threadrange[(0,) {}-(2,) {}-('min must be > 0',) {}]`.

cheroot/test/test_server.py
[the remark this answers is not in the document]
pR0Ps: Adding the "Raises:" in the same format as I saw elsewhere in the code didn't seem to placate the linter for some reason but all it;s other complaints about the docstrings have been resolved.

cheroot/test/test_server.py
webknjaz: Right, I keep forgetting that the `ids` behaves differently from `pytest.param(..., id=...)`. It'd make sense to address this on another level, but for now, let's drop this, so it doesn't block the merge further. It's one of those things that can be improved in many tests and could benefit from a more generic approach. This can be done separately, when somebody has inspiration for mass-refactoring...

cheroot/test/test_server.py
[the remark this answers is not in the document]
webknjaz: I don't see the related error in CI so you probably shouldn't worry about that. `Raises:` is so-called “napoleon” format for docstring metadata but the `darglint` plugin enforces the Sphinx-native style which needs `:raises:`.
FWIW it's not enabled everywhere because of the inconsistency in use. Those napoleon-style definitions should be converted to the native sphinx style over time.

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,20 @@ def __init__(
...
             min (int): minimum number of worker threads
-            max (int): maximum number of worker threads
+            max (int): maximum number of worker threads (-1 for no max)
             accepted_queue_size (int): maximum number of active
                 requests in queue
             accepted_queue_timeout (int): timeout for putting request
                 into queue
+
+        Raises:
+            ValueError: if the min/max values are invalid
```
webknjaz, line 161:
```suggestion
        :raises ValueError: if the min/max values are invalid
```

cheroot/workers/threadpool.py
```diff
@@ -167,18 +175,14 @@ def __init__(
...
-                format(worker_name=worker.name)
-            )
-            worker.start()
-        for worker in self._threads:
-            while not worker.ready:
-                time.sleep(.1)
+        """Start the pool of threads.
+
+        Raises:
+            RuntimeError: if the pool is already started
```
webknjaz, line 181:
```suggestion
        :raises RuntimeError: if the pool is already started
```

cheroot/workers/threadpool.py
```diff
@@ -167,18 +175,14 @@ def __init__(
...
+        """
+        if self._threads:
+            raise RuntimeError('threadpools can only be started once')
```
webknjaz, line 184: I think exception messages usually start with a capital letter.
```suggestion
            raise RuntimeError('Threadpools can only be started once')
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,20 @@ def __init__(
...
             accepted_queue_size (int): maximum number of active
                 requests in queue
             accepted_queue_timeout (int): timeout for putting request
                 into queue
+
+        Raises:
+            ValueError: if the min/max values are invalid
         """
+        if min < 1:
+            raise ValueError('min must be > 0')
```
webknjaz, line 164: The WPS226 linting violation (https://results.pre-commit.ci/run/github/16620627/1670250782.Xd5wEFEFTPKTk0zOvjQ30Q) hinted me that it could be useful to have the broken values presented to the users:
```suggestion
            raise ValueError(f'{min=} must be > 0')
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,20 @@ def __init__(
...
+        if -1 < max < min:
+            raise ValueError('max must be > min (or -1 for no max)')
```
webknjaz, line 166:
```suggestion
            raise ValueError(f'{max=} must be > {min=} (or -1 for no max)')
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,20 @@ def __init__(
...
                 requests in queue
             accepted_queue_timeout (int): timeout for putting request
                 into queue
+
+        Raises:
+            ValueError: if the min/max values are invalid
         """
+        if min < 1:
+            raise ValueError('min must be > 0')
+        if -1 < max < min:
```
webknjaz, line 165: I've been staring at this for a while and realized that the `-1` value is just a dirty hack. It would be more fitting for the low-level ecosystems like C as a magical constant.
In Python, though, we can use the actual infinity, which is semantically closer to what it attempts to represent.
For compatibility, though, we could internally convert `-1` into infinity.
```suggestion
        if max < 0:
            max = float('inf')
        if max < min:
```
(`float('inf') is math.inf`)

cheroot/workers/threadpool.py
pR0Ps: This `{x=}` f-string syntax is only available in Python3.8+ and this package supports v3.6 - I've used the more verbose `x={x}`syntax instead.

cheroot/workers/threadpool.py
pR0Ps: This `{x=}` f-string syntax is only available in Python3.8+ and this package supports v3.6 - I've used the more verbose `x={x}`syntax instead.

pR0Ps:
@webknjaz addressed all the latest comments

cheroot/test/test_server.py
```diff
@@ -439,3 +441,88 @@ def many_open_sockets(request, resource_limit):
...
+        (-1, 1, 'min=-1 must be > 0'),
+        (-1, 2, 'min=-1 must be > 0'),
+        (0, -1, 'min=0 must be > 0'),
+        (0, 0, 'min=0 must be > 0'),
+        (0, 1, 'min=0 must be > 0'),
+        (0, 2, 'min=0 must be > 0'),
+        (1, 0, 'Expected an integer or the infinity value for the `max` argument but got 0.'),
+        (1, 0.5, 'Expected an integer or the infinity value for the `max` argument but got 0.5.'),
+        (2, 0, 'Expected an integer or the infinity value for the `max` argument but got 0.'),
+        (2, '1', 'Expected an integer or the infinity value for the `max` argument but got \'1\'.'),
```
webknjaz, line 493:
```suggestion
        (2, '1', "Expected an integer or the infinity value for the `max` argument but got '1'."),
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,34 @@ def __init__(
...
             accepted_queue_size (int): maximum number of active
                 requests in queue
             accepted_queue_timeout (int): timeout for putting request
                 into queue
+
+        :raises ValueError: if the min/max values are invalid
+        :raises TypeError: if the max is not an integer or inf
         """
+        if min < 1:
+            raise ValueError('min={} must be > 0'.format(min))
```
webknjaz, line 164: Now that I've allowed f-strings in the linter, it's possible to get bet to them. I'm not against `str.format()` but the placeholders should be named for it to be readable. Just `{}` is rather confusing.
```suggestion
            raise ValueError(f'min={min!s} must be > 0')
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,34 @@ def __init__(
...
+
+        if max == float('inf'):
+            pass
+        elif not isinstance(max, int) or max == 0:
+            raise TypeError(
+                'Expected an integer or the infinity value for the `max` '
+                'argument but got {!r}.'.format(max),
```
webknjaz, line 171:
```suggestion
                f'argument but got {max!r}.',
```

cheroot/workers/threadpool.py
```diff
@@ -151,12 +151,34 @@ def __init__(
...
+            )
+        elif max < 0:
+            max = float('inf')
+
+        if max < min:
+            raise ValueError(
+                'max={} must be > min={} (or infinity for no max)'
+                ''.format(max, min),
```
webknjaz, line 179:
```suggestion
                f'max={max!s} must be > min={min!s} (or infinity for no max)',
```

cheroot/test/test_server.py
```diff
@@ -439,3 +441,88 @@ def many_open_sockets(request, resource_limit):
...
+    tp = ThreadPool(
+        server=None,
+        min=minthreads,
+        max=maxthreads,
+    )
+    assert tp.min == minthreads
+    if maxthreads < 0:
+        assert tp.max == float('inf')
+    else:
+        assert tp.max == maxthreads
```
webknjaz, line 476: It's best to have a clear expectation regarding what the max is, instead of putting logic into the test.
```suggestion
@pytest.mark.parametrize(
    ('minthreads', 'maxthreads', 'inited_maxthreads'),
    (
        (
            # NOTE: The docstring only mentions -1 to mean "no max", but other
            # NOTE: negative numbers should also work.
            1,
            -2,
            float('inf'),
        ),
        (1, -1, float('inf')),
        (1, 1, 1),
        (1, 2, 2),
        (1, float('inf'), float('inf')),
        (2, -2, float('inf')),
        (2, -1, float('inf')),
        (2, 2, 2),
        (2, float('inf'), float('inf')),
    ),
)
def test_threadpool_threadrange_set(minthreads, maxthreads, inited_maxthreads):
    """Test setting the number of threads in a ThreadPool.

    The ThreadPool should properly set the min+max number of the threads to use
    in the pool if those limits are valid.
    """
    tp = ThreadPool(
        server=None,
        min=minthreads,
        max=maxthreads,
    )
    assert tp.min == minthreads
    assert tp.max == inited_maxthreads
```

cheroot/workers/threadpool.py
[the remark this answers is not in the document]
webknjaz: Looks like this has been addressed. Marking the conversation as resolved...

webknjaz (approved):

Issue #517, opened 2022-09-17
ssl.SSLZeroReturnError exception on startup with builtin ssl backend python 3.8 or above


toppk:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**
I get an exception on startup.  Everything seems to be working okay.
I have attached the exception, and the code used you will need to generate an ssl cert/key pair to test.

❓ **What is the motivation / use case for changing the behavior?**

💡 **To Reproduce**

python test.py

this is the exception so people don't have to download the attachment.  the code is a basic CherryPy startup that I grabbed from https://github.com/cherrypy/cheroot/issues/346

```python-traceback
[17/Sep/2022:01:17:23] ENGINE Started monitor thread 'Autoreloader'.
[17/Sep/2022:01:17:23] ENGINE Serving on https://0.0.0.0:4400
[17/Sep/2022:01:17:23] ENGINE Bus STARTED
[17/Sep/2022:01:17:23] ENGINE Error in HTTPServer.serve
Traceback (most recent call last):
  File "/home/toppk/.local/lib/python3.9/site-packages/cheroot/server.py", line 1823, in serve
    self._connections.run(self.expiration_interval)
  File "/home/toppk/.local/lib/python3.9/site-packages/cheroot/connections.py", line 203, in run
    self._run(expiration_interval)
  File "/home/toppk/.local/lib/python3.9/site-packages/cheroot/connections.py", line 246, in _run
    new_conn = self._from_server_socket(self.server.socket)
  File "/home/toppk/.local/lib/python3.9/site-packages/cheroot/connections.py", line 300, in _from_server_socket
    s, ssl_env = self.server.ssl_adapter.wrap(s)
  File "/home/toppk/.local/lib/python3.9/site-packages/cheroot/ssl/builtin.py", line 277, in wrap
    s = self.context.wrap_socket(
  File "/usr/lib64/python3.9/ssl.py", line 501, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/lib64/python3.9/ssl.py", line 1041, in _create
    self.do_handshake()
  File "/usr/lib64/python3.9/ssl.py", line 1310, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLZeroReturnError: TLS/SSL connection has been closed (EOF) (_ssl.c:1129)

```
[...]

PR #518, opened 2022-09-22
Handle `SSLZeroReturnError` exceptions in stdlib TLS adapter

toppk:
with python 3.8 and above ssl will generate SSLZeroReturnError exceptions for zero-bytes-send connections.  These connections are ignored in cheroot when the exception is SSLError with errno=ssl.SSL_ERROR_EOF
this is the cpython commit that introduced the change of behavior.

https://github.com/python/cpython/pull/18772

* [X] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves https://github.com/cherrypy/cheroot/issues/517

Annoying stack trace on startup, (see issue for full details), but cherrypy still operates normally

No exception is printed, no other change in behavior. 

📋 **Other information**:

📋 **Contribution checklist:**

* [ ] I wrote descriptive pull request text above
* [ ] I think the code is well written
* [ ] I wrote [good commit messages]
* [ ] I used the same coding conventions as the rest of the project
* [ ] The new code doesn't generate linter offenses
* [ ] The PR relates to *only* one subject with a clear title

[related squash]:

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/518)

### DIFF
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -30,8 +30,6 @@ from .. import errors
 from ..makefile import StreamReader, StreamWriter
 from ..server import HTTPServer
 
-generic_socket_error = OSError
-
 
 def _assert_ssl_exc_contains(exc, *msgs):
     """Check whether SSL exception contains either of messages provided."""
@@ -264,7 +262,6 @@ class BuiltinSSLAdapter(Adapter):
 
     def wrap(self, sock):
         """Wrap and return the given socket, plus WSGI environ entries."""
-        EMPTY_RESULT = None, {}
         try:
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
@@ -276,16 +273,25 @@ class BuiltinSSLAdapter(Adapter):
             raise errors.FatalSSLAlert(
                 *tls_connection_drop_error.args,
             ) from tls_connection_drop_error
-        except ssl.SSLError as ex:
-            if ex.errno == ssl.SSL_ERROR_SSL:
-                if _assert_ssl_exc_contains(ex, 'http request'):
-                    # The client is speaking HTTP to an HTTPS server.
-                    raise errors.NoSSLError
-        except generic_socket_error:
-            pass
-        else:
-            return s, self.get_environ(s)
-        return EMPTY_RESULT
+        except ssl.SSLError as generic_tls_error:
+            peer_speaks_plain_http_over_https = (
+                generic_tls_error.errno == ssl.SSL_ERROR_SSL and
+                _assert_ssl_exc_contains(generic_tls_error, 'http request')
+            )
+            if peer_speaks_plain_http_over_https:
+                reraised_connection_drop_exc_cls = errors.NoSSLError
+            else:
+                reraised_connection_drop_exc_cls = errors.FatalSSLAlert
+
+            raise reraised_connection_drop_exc_cls(
+                *generic_tls_error.args,
+            ) from generic_tls_error
+        except OSError as tcp_connection_drop_error:
+            raise errors.FatalSSLAlert(
+                *tcp_connection_drop_error.args,
+            ) from tcp_connection_drop_error
+
+        return s, self.get_environ(s)
 
     def get_environ(self, sock):
         """Create WSGI environ entries to be merged into each request."""
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -269,6 +269,13 @@ class BuiltinSSLAdapter(Adapter):
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
             )
+        except (
+            ssl.SSLEOFError,
+            ssl.SSLZeroReturnError,
+        ) as tls_connection_drop_error:
+            raise errors.FatalSSLAlert(
+                *tls_connection_drop_error.args,
+            ) from tls_connection_drop_error
         except ssl.SSLError as ex:
             if ex.errno == ssl.SSL_ERROR_SSL:
                 if _assert_ssl_exc_contains(ex, 'http request'):
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -270,6 +270,11 @@ class BuiltinSSLAdapter(Adapter):
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
             )
+        except ssl.SSLZeroReturnError:
+            # This is almost certainly due to the cherrypy engine
+            # 'pinging' the socket to assert it's connectable;
+            # the 'ping' isn't SSL.
+            return EMPTY_RESULT
         except ssl.SSLError as ex:
             if ex.errno == ssl.SSL_ERROR_EOF:
                 # This is almost certainly due to the cherrypy engine
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -27,7 +27,6 @@ except ImportError:
 
 from . import Adapter
 from .. import errors
-from .._compat import IS_ABOVE_OPENSSL10
 from ..makefile import StreamReader, StreamWriter
 from ..server import HTTPServer
 
@@ -270,63 +269,16 @@ class BuiltinSSLAdapter(Adapter):
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
             )
-        except ssl.SSLZeroReturnError:
-            # This is almost certainly due to the cherrypy engine
-            # 'pinging' the socket to assert it's connectable;
-            # the 'ping' isn't SSL.
-            return EMPTY_RESULT
         except ssl.SSLError as ex:
-            if ex.errno == ssl.SSL_ERROR_EOF:
-                # This is almost certainly due to the cherrypy engine
-                # 'pinging' the socket to assert it's connectable;
-                # the 'ping' isn't SSL.
-                return EMPTY_RESULT
-            elif ex.errno == ssl.SSL_ERROR_SSL:
+            if ex.errno == ssl.SSL_ERROR_SSL:
                 if _assert_ssl_exc_contains(ex, 'http request'):
                     # The client is speaking HTTP to an HTTPS server.
                     raise errors.NoSSLError
-
-                # Check if it's one of the known errors
-                # Errors that are caught by PyOpenSSL, but thrown by
-                # built-in ssl
-                _block_errors = (
-                    'unknown protocol', 'unknown ca', 'unknown_ca',
-                    'unknown error',
-                    'https proxy request', 'inappropriate fallback',
-                    'wrong version number',
-                    'no shared cipher', 'certificate unknown',
-                    'ccs received early',
-                    'certificate verify failed',  # client cert w/o trusted CA
-                    'version too low',  # caused by SSL3 connections
-                    'unsupported protocol',  # caused by TLS1 connections
-                )
-                if _assert_ssl_exc_contains(ex, *_block_errors):
-                    # Accepted error, let's pass
-                    return EMPTY_RESULT
-            elif _assert_ssl_exc_contains(ex, 'handshake operation timed out'):
-                # This error is thrown by builtin SSL after a timeout
-                # when client is speaking HTTP to an HTTPS server.
-                # The connection can safely be dropped.
-                return EMPTY_RESULT
-            raise
-        except generic_socket_error as exc:
-            """It is unclear why exactly this happens.
-
-            It's reproducible only with openssl>1.0 and stdlib
-            :py:mod:`ssl` wrapper.
-            In CherryPy it's triggered by Checker plugin, which connects
-            to the app listening to the socket port in TLS mode via plain
-            HTTP during startup (from the same process).
-
-
-            Ref: https://github.com/cherrypy/cherrypy/issues/1618
-            """
-            is_error0 = exc.args == (0, 'Error')
-
-            if is_error0 and IS_ABOVE_OPENSSL10:
-                return EMPTY_RESULT
-            raise
-        return s, self.get_environ(s)
+        except generic_socket_error:
+            pass
+        else:
+            return s, self.get_environ(s)
+        return EMPTY_RESULT
 
     def get_environ(self, sock):
         """Create WSGI environ entries to be merged into each request."""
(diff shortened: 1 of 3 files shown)


toppk:
not sure why it doesn't like my commit message.

webknjaz:
> not sure why it doesn't like my commit message.

Don't worry, that platisd/bad-commit-message-blocker check is broken, I haven't gotten to fixing it.

But the one in pre-commit (see https://results.pre-commit.ci/run/github/16620627/1663816002.9KC-J1hTS36wzU6ARo-C4w) where there's two flake8 violations introduced is a legit failure. The complexity has been increased past the acceptable limit and so it needs to be simplified for the linting to go green.

cheroot/ssl/builtin.py
```diff
@@ -277,6 +277,11 @@ def wrap(self, sock):
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
             )
+        except ssl.SSLZeroReturnError:
+            # This is almost certainly due to the cherrypy engine
+            # 'pinging' the socket to assert it's connectable;
+            # the 'ping' isn't SSL.
```
webknjaz, line 283: You said that you've disabled the Checker and it's still happening. Pretty sure it's unrelated, then.
It must be this https://github.com/cherrypy/cheroot/blob/3f9b1cb/cheroot/ssl/builtin.py#L55-L83.
By the way, maybe the `suppress()` in that helper function needs to be extended too.

cheroot/ssl/builtin.py
```diff
@@ -277,6 +277,11 @@ def wrap(self, sock):
...
+            return EMPTY_RESULT
         except ssl.SSLError as ex:
             if ex.errno == ssl.SSL_ERROR_EOF:
```
webknjaz, line 286: I wonder if we could declare an internal `CherootSSLZeroReturnError` in `cheroot._compat`, and then wrap the TLS calls to re-raise both `ssl.SSLZeroReturnError` and `ssl.SSLError` with `ssl.SSL_ERROR_EOF` as this internal exception. This would allow having a cleaner structural block for handling two incarnations of the same situation as one thing without having to care about two copy-pasted blocks are handled the same way.

webknjaz: Oh, I see, we already have `errors.NoSSLError` so maybe not `_compat` but `errors` would be a better place for the exception.

webknjaz: And maybe all these branchy cases could be treated the same with a shorter and better readable error handling on this layer of abstraction.

webknjaz:
Hey @mxii-ca, you've provided a lot of invaluable input on TLS in the past. Would you mind checking this patch? Am I right that it's related to https://github.com/cherrypy/cheroot/blob/3f9b1cb/cheroot/ssl/builtin.py#L55-L83?

cheroot/ssl/builtin.py
toppk: It isn't the checker, but cherrypy has a call to a module called portend which is used to move the lifecycle of the service to the active state.  This is a simple responds to TCP port check.  So it is a check but not the checker...

cheroot/ssl/builtin.py
toppk: in doing a full review of how this exception is handled, and comparing to the pyopenssl implementation, I would say my opinion is that we should change this entire exception code to just swallow the exception and return EMPTY_RESULT, except for the   http-on-https port condition.  

Something like this.
```
try:
            s = self.context.wrap_socket(
                sock, do_handshake_on_connect=True, server_side=True,
            )
            return s, self.get_environ(s)
except ssl.SSLError as ex:
            if ex.errno == ssl.SSL_ERROR_SSL:
                     if _assert_ssl_exc_contains(ex, 'http request'):
                           # The client is speaking HTTP to an HTTPS server.
                           raise errors.NoSSLError
except generic_socket_error:
           pass
except  ssl.SSLZeroReturnError:
           pass
return EMPTY_RESULT
```
If we ever wanted to log these bad sockets, we can add to this, but right now we dissect a bunch of exceptions all for it to be ignored anyway.

I would also get rid of the `except Exception:` in `server.serve()` as there shouldn't be any unknown/unexpected exceptions bubbling up there, the only unknown/unexpected exceptions should be coming from user code, and the workers are in different threads.

toppk:
I've added a second commit that attempts simplification of the code.  The big difference is that the earlier code would reraise some exceptions that wrap_socket raised, and this will just ignore all of them except for that only exception that cheroot handles (http-over-https aka errors.NoSSLError).

If this approach is satisfactory I can squash the commits for a cleaner commit log.

toppk:
> This pull request **introduces 2 alerts** when merging [6c98cef](https://github.com/cherrypy/cheroot/commit/6c98cef1e686f5dd38111175e543cab8d6243e9b) into [3915f35](https://github.com/cherrypy/cheroot/commit/3915f35fbfe34c7cd0e4ffe68aba092f750345ac) - [view on LGTM.com](https://lgtm.com/projects/g/cherrypy/cheroot/rev/pr-814ef145497e27488287d057ed4d6b19a972ff61)
> 
> **new alerts:**
thank you bot!

> * 1 for Unreachable 'except' block
yes, I didn't notice that, this code can get removed as well.
> * 1 for Unused import
if this approach is accepted, I will remove this code

cheroot/ssl/builtin.py
```diff
@@ -272,63 +272,20 @@ def bind(self, sock):
 
     def wrap(self, sock):
         """Wrap and return the given socket, plus WSGI environ entries."""
-        EMPTY_RESULT = None, {}
```
webknjaz, line 275: This should be kept. The variable name is a semantic label of the data it holds.

cheroot/ssl/builtin.py
```diff
@@ -272,63 +272,20 @@ def bind(self, sock):
...
         try:
             s = self.context.wrap_socket(
                 sock, do_handshake_on_connect=True, server_side=True,
             )
+            return s, self.get_environ(s)
```
webknjaz, line 279: There must be exactly one statement in a try-block. Otherwise, await surprises. For example, `self.get_environ()` call would get processed by the except blocks here, while one would only expect the `wrap_socket()`'s errors to be handled in this place.

cheroot/ssl/builtin.py
webknjaz: Oh, that's good to know. That would be useful to mention.

cheroot/ssl/builtin.py
webknjaz: My understanding is that the reason this is more complicated than pyopenssl is that pyopenssl does some processing internally and turns weird cases into specific exceptions. I'd love the same to happen with the `ssl` implementation. But it'd have to happen on our side since stdlib doesn't do this. And even if it did, we'd need to have something for compatibility for years.

> I would also get rid of the `except Exception:` in `server.serve()` as there shouldn't be any unknown/unexpected exceptions bubbling up there, the only unknown/unexpected exceptions should be coming from user code, and the workers are in different threads.

Agreed. Although, this deserves a separate discussion/PR.

webknjaz:
> I've added a second commit that attempts simplification of the code. The big difference is that the earlier code would reraise some exceptions that wrap_socket raised, and this will just ignore all of them except for that only exception that cheroot handles (http-over-https aka errors.NoSSLError).

Well, it may cause hard-to-debug states if the exceptions are just ignored silently. The currently listed errors were carefully curated, and I'm not sure if it's reasonable to throw it all away without a mechanism for inspection. It'd be good to at least have some logging there, I'm not sure.

I see your point that most of the exceptions would need to be ignored anyway, but I foresee difficulties for the end-users attempting to figure out what they've set up wrong in the TLS configuration, and it tends to be nontrivial when it comes to understanding crypto stuff like ciphers. How would one see that a client is sending requests with non-matching ciphers enabled, for example?

By the way, I'm not sure if you misunderstood my initial idea — I was hoping that the call to `wrap_socket()` could be somehow decorated with an exception transformer.

cheroot/ssl/builtin.py
toppk: revived.

cheroot/ssl/builtin.py
toppk: I moved it into a try/else

toppk:
> Well, it may cause hard-to-debug states if the exceptions are just ignored silently. The currently listed errors were carefully curated, and I'm not sure if it's reasonable to throw it all away without a mechanism for inspection. It'd be good to at least have some logging there, I'm not sure.

I can see that the curation was a hard fought for list, but when I looked at pyopenssl backend I chuckled a bit.    As for logging, it is a no win situation, as either it will annoy some people or not annoy others.  For an internet facing service, logs will get litter with the results of random port scans, for some development/production scenarios it may be desirable, but that might also include things that are on the curated list.   

> 
> I see your point that most of the exceptions would need to be ignored anyway, but I foresee difficulties for the end-users attempting to figure out what they've set up wrong in the TLS configuration, and it tends to be nontrivial when it comes to understanding crypto stuff like ciphers. How would one see that a client is sending requests with non-matching ciphers enabled, for example?

of course, tls communicates back to the client failure, so silently ignoring on the server side is not removing all information.  The two issues I have with current approach,

1. that if we are deciding to log an exception, the stack trace isn't the best format
2. if the user turns on some sort of error logging, we will be generating one on every startup due to the portend port check.

Another thing while the current blacklist approach of generating a stack trace on all unknown exceptions is very nice to send users your way (as it did me), it has the obvious downside where we are chasing new exceptions categories, and it is not obvious what do.  An actual whitelist of exceptions we want to log, and a facility for users to turn on/off said logging may be useful.  

> 
> By the way, I'm not sure if you misunderstood my initial idea — I was hoping that the call to `wrap_socket()` could be somehow decorated with an exception transformer.

Yeah, I was thinking about extending context, and then wrap socket can produce better errors, or a solution like the with supress() that exists elsewhere, but I think those would just obfuscate the execution from the code.  The wrap() function is the clearest way to express how we want to handle the wrap exceptions, and doesn't do much else.

It would be cool if except got some of the PEP 634/Structural Pattern Matching features, but maybe this use case is just going to look like a bunch of duct tape no matter what (bad exceptions being raised, supporting multiple python runtimes).   I may send something to python-ideas anyway.

webknjaz:
> As for logging, it is a no win situation, as either it will annoy some people or not annoy others. For an internet facing service, logs will get litter with the results of random port scans

Yeah. Although, I wouldn't bundle any behavior changes in this PR anyway — I think it should just contain a bugfix with minimum changes. And any refactoring needs to go elsewhere.
I was thinking that it'd be nice to have prod/dev/debug modes that would maybe control this but that's a feature. Also, I've been meaning to look into making use of `logging` consistently so that the end-users could suppress what they don't need using standard mechanisms. Plus, with such a framework in place, we could log TLS errors nicely, without tracebacks (unless in dev/debug mode), similar to what other web servers typically do.
With that in mind, implementing the use of `logging` across the project would be a blocker to having env-dependent toggles.

> Yeah, I was thinking about extending context, and then wrap socket can produce better errors, or a solution like the with supress() that exists elsewhere, but I think those would just obfuscate the execution from the code. The wrap() function is the clearest way to express how we want to handle the wrap exceptions, and doesn't do much else.

How about a method? `self.context` is already an instance attribute. So it wouldn't need to be accessed directly in `wrap()`.

```python
def _wrap_context_socket_with_consistent_errors(self, sock):
    try:
        return self.context.wrap_socket(
            sock, do_handshake_on_connect=True, server_side=True,
        )
    except ...:
        raise CherootTLSEOL
    except ...:
        raise CherootTLSEOL

def wrap(self, sock)
    try:
        return self._wrap_context_socket_with_consistent_errors(sock)
    except CherootTLSEOL:
        ...
    except ...:
        ...
```

> It would be cool if except got some of the PEP 634/Structural Pattern Matching features, but maybe this use case is just going to look like a bunch of duct tape no matter what (bad exceptions being raised, supporting multiple python runtimes). I may send something to python-ideas anyway.

That would be a no-go for us for another 3-4 years because of compat requirements. Libs don't have the same luxury that apps do.

toppk:
i'll take a look at your last recommendation and update this pr.  not sure it will pass the complexity review, but we'll see.

thezoggy (changes requested):
Part of the code removed is the usage of `IS_ABOVE_OPENSSL10` which is now an unused import. 
`from .._compat import IS_ABOVE_OPENSSL10, suppress`
--------------------------^^^^^^^^^^^^^
So that should also be cleaned up to make CI happy.

webknjaz:
> Part of the code removed is the usage of `IS_ABOVE_OPENSSL10` which is now an unused import. `from .._compat import IS_ABOVE_OPENSSL10, suppress` --------------------------^^^^^^^^^^^^^ So that should also be cleaned up to make CI happy.

Yes, that's what the LGTM service has been reporting, along with the linters. The author acknowledged this too.

webknjaz:
Ref: https://github.com/cherrypy/cheroot/issues/346

The-Compiler:
FWIW, I can confirm this fixes #517 for me.

satmandu:
Could we please get this merged? @toppk Are you still planning on pushing this PR to completion?

toppk:
> Could we please get this merged? @toppk Are you still planning on pushing this PR to completion?

I am willing to, but I'm having a hard time  incorporating the feedback into a new design.   I will revisit this week and create something so there can be some further discussion/action on this.

webknjaz:
Do you need any help with this?

webknjaz:
> with python 3.8 and above ssl will generate SSLZeroReturnError

@toppk could you clarify why you're talking about `ssl.SSLZeroReturnError` while the CPython PR you linked is about `ssl.SSLEOFError`? These are different exceptions. Also, AFAICS both are subclasses of `ssl.SSLError`.

thezoggy:
> > with python 3.8 and above ssl will generate SSLZeroReturnError
> 
> @toppk could you clarify why you're talking about `ssl.SSLZeroReturnError` while the CPython PR you linked is about `ssl.SSLEOFError`? These are different exceptions. Also, AFAICS both are subclasses of `ssl.SSLError`.

because its one and the same, openssl changed things and cpython reacted and such, 
https://github.com/python/cpython/commit/13df5d3497e6148b2b42bed7b5593144f8f6216c

webknjaz:
> because its one and the same, openssl changed things and cpython reacted and such,
> [python/cpython@13df5d3](https://github.com/python/cpython/commit/13df5d3497e6148b2b42bed7b5593144f8f6216c)

Oh, I didn't realize, thanks!

webknjaz:
Thanks @toppk!

PR #633, opened 2023-05-20
Drop support for Python 3.7 and earlier

### LINKED ISSUE #565: Drop support for Python 3.7
Python 3.6 has been EOL for some time now. Let's drop support for it.

jaraco:
Fixes #565.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/633)


.github/workflows/ci-cd.yml
```diff
@@ -674,46 +673,20 @@ jobs:
...
         os:
         - ubuntu-22.04
         - ubuntu-20.04
         - macos-11.0
         - macos-latest
         - windows-2019
         - windows-2022
-        exclude:
+        # exclude:
         # NOTE: Windows PyPy jobs are excluded to address the tox bug
```
webknjaz, line 684: These are comments referring to the removed lines, they need to go too.

.github/workflows/ci-cd.yml
```diff
@@ -817,16 +790,6 @@ jobs:
         --user
         '${{ env.TOX_VERSION }}'
 
-    - name: Patch tox.ini for Python 3.6 under Windows
-      if: >-
-        runner.os == 'Windows'
-        && matrix.python-version == '3.6'
-      run: >-
-        sed -i
-        's/^package_env\(\s\)\?=.*/package_env = py36-win-dummy/g'
```
webknjaz, line 826: There's a related hack in `tox.ini`. It can be removed now as well.

webknjaz (commented):
Besides those two notes, the rest looks good to me.

.github/workflows/ci-cd.yml
jaraco: I retained the comments because they don't refer to specific Python versions but to other factors (OS, PyPy) which seemingly apply to other Python versions and would be relevant should PyPy be added to the matrix.

.github/workflows/ci-cd.yml
jaraco: Done.

.github/workflows/ci-cd.yml
jaraco: Removed. We can come back here if we need it in the future.

Issue #631, opened 2023-05-20
tox -e pre-commit fails attempting to build cryptography < 3.3


jaraco:
Today I tried to run the pre-commits, but the build env fails:

```
 cheroot reuse-port $ tox -e pre-commit
pre-commit: install_deps> .tox/pre-commit/bin/python -bb -E -s /Users/jaraco/code/cherrypy/cheroot/bin/pip-wrapper pre-commit /Users/jaraco/code/cherrypy/cheroot/requirements/ install -r requirements/tox-pre-commit.in
.pkg: _optional_hooks> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_editable> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: build_editable> python /Users/jaraco/.local/pipx/venvs/tox/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta
pre-commit: install_package_deps> .tox/pre-commit/bin/python -bb -E -s /Users/jaraco/code/cherrypy/cheroot/bin/pip-wrapper pre-commit /Users/jaraco/code/cherrypy/cheroot/requirements/ install 'importlib-metadata; python_version < "3.8"' jaraco.functools 'more-itertools>=2.6'
pre-commit: install_package> .tox/pre-commit/bin/python -bb -E -s /Users/jaraco/code/cherrypy/cheroot/bin/pip-wrapper pre-commit /Users/jaraco/code/cherrypy/cheroot/requirements/ install --force-reinstall --no-deps /Users/jaraco/code/cherrypy/cheroot/.tox/.tmp/package/5/cheroot-9.0.1.dev771+g081aa090-0.editable-py3-none-any.whl
pre-commit: commands[0]> pre-commit run --all-files --show-diff-on-failure
[INFO] Initializing environment for https://github.com/asottile/add-trailing-comma.git.
[INFO] Initializing environment for https://github.com/Lucas-C/pre-commit-hooks.git.
[INFO] Initializing environment for https://github.com/pre-commit/pre-commit-hooks.git.
[...]

PR #635, opened 2023-05-21
Modernize the codebase with black and pyupgrade

jaraco:
- ⚫ Fade to black.
- Ran pyupgrade for Python 3.8+

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/635)


webknjaz:
I'm still quite against imposing black on a well-structured code base. I gravitate towards using autopep8 instead — the output is similar but less intrusive. It'd like the formatting setup be more contributor friendly in this regard.. Though, honestly, I think that current number of formatters that are integrated is about enough, maybe on the edge of the balance between being useful and annoying.

Also, pyupgrade seems to be making potential behavior changes that are hard to spot with multiple formatters applied at once. I'd recommend to run it separately.

cheroot/errors.py
```diff
@@ -36,19 +35,29 @@ def plat_specific_errors(*errnames):
 
 socket_errors_to_ignore = plat_specific_errors(
     'EPIPE',
-    'EBADF', 'WSAEBADF',
```
webknjaz, line 39: Things like this are paired because of their related semantic meaning. These lines mustn't be messed up into something that doesn't relay this connection. Alternatively, they'd have to be moved to separate variables with names and unpacked here.

docs/conf.py
```diff
@@ -103,15 +103,13 @@
     r'http://localhost:\d+/',  # local URLs
     r'https://codecov\.io/gh/cherrypy/cheroot/branch/master/graph/badge\.svg',
     r'https://github\.com/cherrypy/cheroot/actions',  # 404 if no auth
-
```
webknjaz, line 106: Removing separators between sections makes the grouping unobvious. If it's really desired, those bits of data should be put into named variables and merged here, to preserve relaying the original intent.

pyproject.toml
```diff
@@ -11,4 +11,8 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[tool.black]
```
webknjaz, line 14: I still think this is a bad idea. Unless, the bits that hugely break readability are fixed first which would make sense to be separate from pyupgrade.

cheroot/server.py
webknjaz, line 1: Could you only keep the smallest necessary changes in this file — I was hoping to resurrect the parser migration PR and having less merge conflicts would be very helpful..

cheroot/errors.py
```diff
@@ -58,7 +67,8 @@ def plat_specific_errors(*errnames):
 
 acceptable_sock_shutdown_error_codes = {
     errno.ENOTCONN,
-    errno.EPIPE, errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
+    errno.EPIPE,
```
webknjaz, line 70: This one also corresponds to BrokenPipeError which is lost with the contextless formatting

cheroot/ssl/__init__.pyi
```diff
@@ -8,7 +8,13 @@ class Adapter(metaclass=ABCMeta):
     ciphers: Any
     context: Any
     @abstractmethod
-    def __init__(self, certificate, private_key, certificate_chain: Any | None = ..., ciphers: Any | None = ...): ...
+    def __init__(
```
webknjaz, line 11: I wouldn't bother reformatting pyi stubs. I wanted them separately to potentially backport stuff to the older py2-supporting versions of Cheroot but now it'd be reasonable to just transfer them into the Python modules (as a separate change, of course).

jaraco:
Black conflicts in incompatible ways with some of the pre-commit checks, and I can't run pre-commit checks due to #631.

> I'm still quite against imposing black on a well-structured code base. I gravitate towards using autopep8 instead — the output is similar but less intrusive. It'd like the formatting setup be more contributor friendly in this regard.. Though, honestly, I think that current number of formatters that are integrated is about enough, maybe on the edge of the balance between being useful and annoying.

Yes, another autoformatter may be viable. The reason I've settled on black is (a) it was one of the first and seems to be one of the most popular autoformatters and (b) I've configured my dev setup to format on save for black (when configured in the project).

I've adopted black in jaraco/skeleton projects, so I'm invested in it and it's negligible effort for me to employ. Any other formatter, I'll need to learn how to use it and probably have to invest some time in trying to get my editor to autoformat for this project differently.

I thought I was motivated by a PR that I'd reviewed and closed recently suggesting to use black and pyupgrade, but I can't find that PR now.


> Also, pyupgrade seems to be making potential behavior changes that are hard to spot with multiple formatters applied at once. I'd recommend to run it separately.

I did run pyupgrade in a separate commit, making it easy to see that change.

I'm going to abandon this effort for now.

webknjaz:
> Any other formatter, I'll need to learn how to use it and probably have to invest some time in trying to get my editor to autoformat for this project differently.

Yes and no. I'd suggest that if you configure your editor to run pre-commit, it'll pick up any formatter integrated there, regardless of its name. I saw that you had problems with pre-commit and will try to improve its usability. Projects that have both formatters and pre-commit usually integrate it there. The pre-commit.ci integration also auto-pushes commits to PRs so you don't really have to care about that locally, unless you want to. pyupgrade usually goes into pre-commit too, by the way I meant that it needs to be submitted as a separate PR since it does syntax changes, this type of thing is better to review separately and PRs don't really let one do that with individual commits in a sane way — plus, it's undesired formatting bits would not block the syntax updates from being merged separately.

Regarding the popularity, yes, it might be. But the cost is too high for me — this thing tends to stand the way of making useful impact so much that in forced me to start avoiding projects forcing it upon the contributors.
OTOH, when I tested autopep8 last, I liked what it offered me in terms of being less intrusive and only changing things that should obviously change. I'm also looking into ways of keeping the usefulness of some linting violations that formatters hide from unexpected contributors.

PR #745, opened 2025-05-22
When connection queue is full, respond with a 503 error.

itamarst:

* [ ] 🐞 bug fix
* [x] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

When connections can't be handled, the connection is just closed.

A 503 error is returned, as suggested in TODO comment in relevant part of the code.

Since writing to the socket is blocking, this is done in a thread.

📋 **Other information**:

For some reason OpenStack Ironic needs this apparently, I didn't dig into why.

📋 **Contribution checklist:**

* [ ] I used the same coding conventions as the rest of the project
* [ ] The new code doesn't generate linter offenses
* [ ] The PR relates to *only* one subject with a clear title

[related squash]:

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/745)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -101,6 +101,36 @@ __all__ = (
 )
 
 
+if sys.version_info[:2] >= (3, 13):
+    from queue import (
+        Queue as QueueWithShutdown,
+        ShutDown as QueueShutDown,
+    )
+else:
+
+    class QueueShutDown(Exception):
+        """Queue has been shut down."""
+
+    class QueueWithShutdown(queue.Queue):
+        """Add shutdown() similar to Python 3.13+ Queue."""
+
+        _queue_shut_down: bool = False
+
+        def shutdown(self, immediate=False):
+            if immediate:
+                while True:
+                    try:
+                        self.get_nowait()
+                    except queue.Empty:
+                        break
+            self._queue_shut_down = True
+
+        def get(self, *args, **kwargs):
+            if self._queue_shut_down:
+                raise QueueShutDown
+            return super().get(*args, **kwargs)
+
+
 IS_WINDOWS = platform.system() == 'Windows'
 """Flag indicating whether the app is running under Windows."""
 
@@ -1658,6 +1688,8 @@ class HTTPServer:
         self.reuse_port = reuse_port
         self.clear_stats()
 
+        self._unservicable_conns = QueueWithShutdown()
+
     def clear_stats(self):
         """Reset server stat counters.."""
         self._start_time = None
@@ -1866,8 +1898,39 @@ class HTTPServer:
         self.ready = True
         self._start_time = time.time()
 
+    def _serve_unservicable(self):
+        """Serve connections we can't handle a 503."""
+        while self.ready:
+            try:
+                conn = self._unservicable_conns.get()
+            except QueueShutDown:
+                return
+            request = HTTPRequest(self, conn)
+            try:
+                request.simple_response('503 Service Unavailable')
+            except (OSError, errors.FatalSSLAlert):
+                # We're sending the 503 error to be polite, it it fails that's
+                # fine.
+                continue
+            except Exception as ex:
+                # We can't just raise an exception because that will kill this
+                # thread, and prevent 503 errors from being sent to future
+                # connections.
+                self.server.error_log(
+                    repr(ex),
+                    level=logging.ERROR,
+                    traceback=True,
+                )
+            conn.linger = True
+            conn.close()
+
     def serve(self):
         """Serve requests, after invoking :func:`prepare()`."""
+        # This thread will handle unservicable connections, as added to
+        # self._unservicable_conns queue. It will run forever, until
+        # self.stop() tells it to shut down.
+        threading.Thread(target=self._serve_unservicable).start()
+
         while self.ready and not self.interrupt:
             try:
                 self._connections.run(self.expiration_interval)
@@ -2162,8 +2225,7 @@ class HTTPServer:
         try:
             self.requests.put(conn)
         except queue.Full:
-            # Just drop the conn. TODO: write 503 back?
-            conn.close()
+            self._unservicable_conns.put(conn)
 
     @property
     def interrupt(self):
@@ -2201,6 +2263,11 @@ class HTTPServer:
             return  # already stopped
 
         self.ready = False
+
+        # This tells the thread that handles unservicable connections to shut
+        # down:
+        self._unservicable_conns.shutdown(immediate=True)
+
         if self._start_time is not None:
             self._run_time += time.time() - self._start_time
         self._start_time = None
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -8,6 +8,7 @@ import threading
 import types
 import urllib.parse  # noqa: WPS301
 import uuid
+from http import HTTPStatus
 
 import pytest
 
@@ -570,3 +571,36 @@ def test_threadpool_multistart_validation(monkeypatch):
         match='Threadpools can only be started once.',
     ):
         tp.start()
+
+
+def test_overload_results_in_suitable_http_error(request):
+    """A server that can't keep up with requests returns a 503 HTTP error."""
+    localhost = '127.0.0.1'
+    httpserver = HTTPServer(
+        bind_addr=(localhost, EPHEMERAL_PORT),
+        gateway=Gateway,
+    )
+    # Can only handle on request in parallel:
+    httpserver.requests = ThreadPool(
+        min=1,
+        max=1,
+        accepted_queue_size=1,
+        accepted_queue_timeout=0,
+        server=httpserver,
+    )
+
+    httpserver.prepare()
+    serve_thread = threading.Thread(target=httpserver.serve)
+    serve_thread.start()
+    request.addfinalizer(httpserver.stop)
+    # Stop the thread pool to ensure the queue fills up:
+    httpserver.requests.stop()
+
+    _host, port = httpserver.bind_addr
+
+    # Use up the very limited thread pool queue we've set up, so future
+    # requests fail:
+    httpserver.requests._queue.put(None)
+
+    response = requests.get(f'http://{localhost}:{port}', timeout=20)
+    assert response.status_code == HTTPStatus.SERVICE_UNAVAILABLE
--- a/docs/changelog-fragments.d/745.feature.rst
+++ b/docs/changelog-fragments.d/745.feature.rst
@@ -1,0 +1,4 @@
+When load is too high, Cheroot now responds with a 503 Service Unavailable HTTP error.
+Previously it silently closed the connection.
+
+-- by :user:`itamarst`


cheroot/test/test_server.py
```diff
@@ -570,3 +570,35 @@ def test_threadpool_multistart_validation(monkeypatch):
...
+    httpserver.requests.stop()
+
+    host, port = httpserver.bind_addr
+
+    # Use up the very limited thread pool queue we've set up, so future
+    # requests fail:
+    httpserver.requests._queue.put(None)
+
+    response = requests.get(f'http://{host}:{port}', timeout=5)
+    assert response.status_code == 503
```
webknjaz, line 604: Cosmetic nitpick: can we `from http import HTTPStatus` and use those constants instead?

cheroot/test/test_server.py
```diff
@@ -570,3 +570,35 @@ def test_threadpool_multistart_validation(monkeypatch):
...
+        max=1,
+        accepted_queue_size=1,
+        accepted_queue_timeout=0,
+        server=httpserver,
+    )
+
+    httpserver.prepare()
+    serve_thread = threading.Thread(target=httpserver.serve)
+    serve_thread.start()
+    request.addfinalizer(httpserver.stop)
```
webknjaz, line 593: Wow, TIL `pytest.FixtureRequest` has [this method](https://docs.pytest.org/en/8.3.x/reference/reference.html#pytest.FixtureRequest.addfinalizer)!

cheroot/test/test_server.py
```diff
@@ -570,3 +570,35 @@ def test_threadpool_multistart_validation(monkeypatch):
...
+    # Stop the thread pool to ensure the queue fills up:
+    httpserver.requests.stop()
+
+    host, port = httpserver.bind_addr
```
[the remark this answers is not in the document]
itamarst: Seems easier to just bind to 127.0.0.1?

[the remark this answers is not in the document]
webknjaz: Perhaps. The test infra here could use some refactoring. I just didn't want it to diverge too much.

itamarst:
So now it ... fails, but not as consistently? I don't really know.

cheroot/test/test_server.py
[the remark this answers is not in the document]
itamarst: There was no output.

itamarst:
Any other thoughts? It's pretty annoying that it passes some of the time but not all the time.

itamarst:
Looks like we're down to only 3.14 failing, which were already failing on `main` so hopefully unrelated to my changes.

cheroot/test/test_server.py
```diff
@@ -570,3 +571,36 @@ def test_threadpool_multistart_validation(monkeypatch):
...
+    httpserver.requests.stop()
+
+    _host, port = httpserver.bind_addr
+
+    # Use up the very limited thread pool queue we've set up, so future
+    # requests fail:
+    httpserver.requests._queue.put(None)
+
+    response = requests.get(f'http://127.0.0.1:{port}', timeout=20)
+    assert response.status_code == HTTPStatus.SERVICE_UNAVAILABLE.value
```
webknjaz: Doesn't this work?
```suggestion
    assert response.status_code == HTTPStatus.SERVICE_UNAVAILABLE
```

I'm pretty sure those constants have `__eq__()` set up to work with `int`.

cheroot/test/test_server.py
```diff
@@ -570,3 +571,36 @@ def test_threadpool_multistart_validation(monkeypatch):
...
+    # Stop the thread pool to ensure the queue fills up:
+    httpserver.requests.stop()
+
+    _host, port = httpserver.bind_addr
+
+    # Use up the very limited thread pool queue we've set up, so future
+    # requests fail:
+    httpserver.requests._queue.put(None)
+
+    response = requests.get(f'http://127.0.0.1:{port}', timeout=20)
```
webknjaz: 127.0.0.1 is already in a variable. Why not reuse it?

cheroot/test/test_server.py
```diff
@@ -570,3 +571,36 @@ def test_threadpool_multistart_validation(monkeypatch):
         match='Threadpools can only be started once.',
     ):
         tp.start()
+
+
+def test_overload_results_in_suitable_http_error(request):
+    """A server that can't keep up with requests returns a 503 HTTP error."""
+    localhost = '127.0.0.1'
```
webknjaz, line 578: I think it might be possible to import a constant with this from somewhere.

cheroot/server.py
```diff
@@ -2201,6 +2221,7 @@ def stop(self):  # noqa: C901  # FIXME
             return  # already stopped
 
         self.ready = False
+        self._unservicable_conns.put(None)
```
webknjaz: Could you add a code comment on the semantics of adding this magic value?

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
...
+                    repr(ex),
+                    level=logging.ERROR,
+                    traceback=True,
+                )
+            conn.linger = True
+            conn.close()
+
     def serve(self):
         """Serve requests, after invoking :func:`prepare()`."""
+        threading.Thread(target=self._serve_unservicable).start()
```
webknjaz, line 1932: Do we need to wait for the thread to complete somewhere?

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
         self.ready = True
         self._start_time = time.time()
 
+    def _serve_unservicable(self):
+        """Serve connections we can't handle a 503."""
+        while self.ready:
+            conn = self._unservicable_conns.get()
+            if conn is None:
```
webknjaz: Can we use a named variable for the sentinel?

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
         self.ready = True
         self._start_time = time.time()
 
+    def _serve_unservicable(self):
+        """Serve connections we can't handle a 503."""
+        while self.ready:
+            conn = self._unservicable_conns.get()
```
webknjaz: Does this need a timeout?

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
...
+            if conn is None:
+                return
```
webknjaz, line 1907: I think we should call `shutdown()` here (under Python 3.13+).

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
...
+            request = HTTPRequest(self, conn)
+            try:
+                request.simple_response('503 Service Unavailable')
+            except Exception as ex:
+                self.server.error_log(
+                    repr(ex),
+                    level=logging.ERROR,
+                    traceback=True,
+                )
+            conn.linger = True
```
webknjaz, line 1924: Why is this necessary?

cheroot/server.py
```diff
@@ -1866,8 +1868,27 @@ def prepare(self):  # noqa: C901  # FIXME
...
+    def _serve_unservicable(self):
+        """Serve connections we can't handle a 503."""
+        while self.ready:
+            conn = self._unservicable_conns.get()
+            if conn is None:
+                return
+            request = HTTPRequest(self, conn)
+            try:
+                request.simple_response('503 Service Unavailable')
+            except Exception as ex:
```
webknjaz, line 1915: Is there a way to make this only intercept narrow exceptions?

cheroot/test/test_server.py
itamarst: Ah, good point.

cheroot/test/test_server.py
itamarst: Couldn't find it.

cheroot/server.py
itamarst: ... Probably not? other way around, it's possible it should be a daemon thread if you don't want sending 503s to delay shutdown of a process.

cheroot/server.py
itamarst: Assuming stop() is always used, then no.

cheroot/server.py
itamarst: Can you clarify what you mean?

cheroot/server.py
itamarst: I was just copying existing code that did that in error handling, but empirically looks like not on Linux at least. Will see what CI does on Windows and macOS.

cheroot/server.py
itamarst: We need a catchall so the thread doesn't die. But I can make more likely exceptions not get logged, at least.

itamarst:
Failures are do to 3.14 and #747, unrelated to this PR as far as I can tell.

cheroot/server.py
webknjaz: I just learned about https://docs.python.org/3/library/queue.html#queue.Queue.shutdown and thought that it could perhaps be incorporated. I haven't used it myself, so I'm asking if there's a place for calling this new method.

webknjaz:
@itamarst i've hit rebase on GH UI post a few CI fixes to `main` and it looks like the new test is flaky in there: https://github.com/cherrypy/cheroot/actions/runs/15811177788/job/44562688410?pr=745#step:15:1574

itamarst:
Seems better maybe with `linger` added back?

cheroot/server.py
webknjaz: Sounds good. I think it'd be a good idea to also add a code comment explaining the need for catch-all. Though, maybe it's something to implement via a `@contextlib.contextmanager` or `contextlib.suppress`, so that the thread protection is separate from expected exception handling.

cheroot/test/test_server.py
webknjaz: Ah, okay. Maybe I imagined it existing..

webknjaz: Marking this thread as resolved.

cheroot/server.py
itamarst: Ahh, yes that might be useful.

cheroot/server.py
itamarst: I added a comment, seemed easiest.

cheroot/server.py
```diff
@@ -101,6 +101,38 @@
...
+                        self.get_nowait()
+                    except queue.Empty:
+                        break
+            self.put(self._SHUT_DOWN_TOKEN)
+
+        def get(self, *args, **kwargs):
+            result = super(QueueWithShutdown, self).get(*args, **kwargs)
+            if result is self._SHUT_DOWN_TOKEN:
+                raise QueueShutDown
+            return result
```
webknjaz: I like what you did here. Can we have better var names + drop the super args that aren't necessary with Python 3?
```suggestion
            next_item_in_line = super().get(*args, **kwargs)
            if next_item_in_line is self._SHUT_DOWN_TOKEN:
                raise QueueShutDown
            return next_item_in_line
```

cheroot/server.py
```diff
@@ -101,6 +101,38 @@
...
+        _SHUT_DOWN_TOKEN = object()
+
+        def shutdown(self, immediate=False):
+            if immediate:
+                while True:
+                    try:
+                        self.get_nowait()
+                    except queue.Empty:
+                        break
+            self.put(self._SHUT_DOWN_TOKEN)
```
webknjaz: Is it enough for us to only be able to retrieve this once?

cheroot/server.py
```diff
@@ -1866,8 +1900,39 @@ def prepare(self):  # noqa: C901  # FIXME
...
+        """Serve connections we can't handle a 503."""
+        while self.ready:
+            try:
+                conn = self._unservicable_conns.get()
+            except QueueShutDown:
+                return
+            request = HTTPRequest(self, conn)
+            try:
+                request.simple_response('503 Service Unavailable')
+            except (socket.error, errors.FatalSSLAlert):
```
webknjaz: `socket.error` is an ancient alias for `OSError`
```suggestion
            except (OSError, errors.FatalSSLAlert):
```

webknjaz:
@itamarst one thing I forgot to ask for is adding a change note for Towncrier. Could you do that? 
https://cheroot.cherrypy.dev/en/latest/contributing/guidelines/#adding-change-notes-with-your-prs

webknjaz:
@itamarst I think this is mostly ready, short of the requests above. When writing a change note, https://cheroot--745.org.readthedocs.build/en/745/history/#to-be-included-in-the-next-release can be used to preview rendering and compare the writing style with the surrounding text.

cheroot/server.py
itamarst: Been writing Python for way too long :)

cheroot/server.py
itamarst: For this use case, yes, for future use cases, not necessarily. So I switched to a different mechanism.

itamarst:
OK back to you @webknjaz

cheroot/server.py
webknjaz: I don't blame you :) There's also a lot of ancient compat hacks all over. I think I was cleaning out some Python 2.4 compat things in CherryPy the other day…

webknjaz (approved):
Wonderful! Thanks for the contrib.

cheroot/server.py
```diff
@@ -1866,8 +1898,39 @@ def prepare(self):  # noqa: C901  # FIXME
...
+                request.simple_response('503 Service Unavailable')
+            except (OSError, errors.FatalSSLAlert):
+                # We're sending the 503 error to be polite, it it fails that's
+                # fine.
+                continue
+            except Exception as ex:
+                # We can't just raise an exception because that will kill this
+                # thread, and prevent 503 errors from being sent to future
+                # connections.
+                self.server.error_log(
```
martin-vi, line 1919: There is no `self.server` within `HTTPServer`.

webknjaz: Good catch! I filed https://github.com/cherrypy/cheroot/issues/797 to track this as I missed it originally. @martin-vi would you like to send a quick PR?

@itamarst if hit, this would cause the problem the comment above warns against, right?

PR #752, opened 2025-07-16
Add private key password option to ssl adapters

jatalahd:
- It is now possible to use password protected private keys in both builtin and openssl ssl-adapters
- Added also positive and negative unit test cases
- With reference to https://github.com/cherrypy/cherrypy/issues/1583

* [ ] 🐞 bug fix
* [X] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

https://github.com/cherrypy/cherrypy/issues/1583

With the current functionality it is only possible to use ssl adapters with private keys without password protection

With this change, there is a new option to give the ssl adapter a "private_key_password" argument, which can be in either string or bytestring format.

📋 **Other information**:
Added also unit tests to test the new functionality

📋 **Contribution checklist:**

* [X] I wrote descriptive pull request text above
* [X] I think the code is well written
* [X] I wrote [good commit messages]
* [X] Unit tests for the changes exist
* [X] I used the same coding conventions as the rest of the project
* [X] The new code doesn't generate linter offenses
* [X] The PR relates to *only* one subject with a clear title

[related squash]:

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/752)


jatalahd:
Fixed linter issues with second commit, but now it seems that I need to do the git squash... Anyway, it would be important for us to get this improvement into some upcoming cheroot release, which we can then reference in our pipfile so that we can take these changes into use.

webknjaz:
@jatalahd thanks for the contrib! You don't *have to* squash commits unconditionally. Only if the commits are non-atomic. If they aren't, it's usually a good idea to combine them to keep Git history clean.

One thing that's definitely missing is a change note. Read https://cheroot.cherrypy.dev/en/latest/contributing/guidelines/#adding-change-notes-with-your-prs and follow the guidelines. Do you best and if it needs editing, I'll tell you.

I can't give you a timeline right now. I'm a bit unhappy about the state of the TLS adapters in general and wanted to redesign them eventually. And so I'm a bit hesitant on what changes would be acceptable in the public API. I'll need to think about it first.

I'll leave a few notes in the diff but that'll be an incomplete review.

cheroot/ssl/pyopenssl.pyi
```diff
@@ -31,9 +31,16 @@ class pyOpenSSLAdapter(Adapter):
         private_key,
         certificate_chain: Any | None = ...,
         ciphers: Any | None = ...,
+        private_key_password: Any | None = ...,
```
webknjaz: Plz avoid `Any` at all cost. Don't we know what types are expected here? Looks like there's `bytes | str` in the other place.

cheroot/test/test_ssl.py
```diff
@@ -726,3 +786,99 @@ def test_http_over_https_error(
...
+    tls_adapter = tls_adapter_cls(
+        certificate=cert_file,
+        private_key=key_file,
+        private_key_password=str(uuid.uuid4()),
+    )
+
+    httpserver.ssl_adapter = tls_adapter
+
+    # expecting SSLError when preparing server
+    with pytest.raises(OpenSSL.SSL.Error):
```
webknjaz: Always use the `match=` argument with `pytest.raises()`. This will also make the code comment unnecessary.

cheroot/test/test_ssl.py
```diff
@@ -118,6 +123,61 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
     return httpserver
 
 
+def create_self_signed_certificate(key_object):
```
webknjaz: Is it possible to use `trustme` to get more TLS certificates? Can we use it for initial provisioning and modify to assign a password if needed?

cheroot/test/test_ssl.py
```diff
@@ -118,6 +123,61 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
...
+    """Create a dummy certificate that can be only used to test server start up."""
+    with tempfile.NamedTemporaryFile(delete=False, suffix='.pem') as tf:
+        cert_file_name = Path(tf.name)
```
webknjaz: Looks like this leaves garbage on disk that nothing cleans after. Ideally, the `tmp_path` fixture must be used as it takes care of that automatically. But in general, we should integrate with the existing fixtures that make use of `trustme`.

cheroot/test/test_ssl.py
```diff
@@ -726,3 +786,99 @@ def test_http_over_https_error(
...
+    ('adapter_type', 'key_password'),
+    (('pyopenssl', 'Pw-1234567'),),
+)
+def test_openssl_adapter_with_false_key_password(
+    create_private_key_with_password,
+    adapter_type,
+    key_password,
+):
+    """Check that server init fails when wrong private key password given."""
+    key_file, cert_file = create_private_key_with_password
```
webknjaz: It's uncommon to use verbs in fixture names that don't return callables.

cheroot/ssl/pyopenssl.pyi
jatalahd: Replaced Any | None with str | bytes | None

cheroot/test/test_ssl.py
jatalahd: Added `match=` argument, but the textual error messages are quite elaborous, so a partial regex match is now used.

cheroot/test/test_ssl.py
jatalahd: Reworked the test fixtures so that removed the old methods and added one new that gets the private key from trustme and rewrites that with password protection using cryptography module. The rewritten file is removed at teardown (after yield) using `Path.unlink()` . Much cleaner than the previous, but not perfect. I cannot do better than this, tho.

cheroot/test/test_ssl.py
jatalahd: These methods are now removed, however similar approach is still used in the new replacing method. Garbage is however always cleaned using `Path.unlink()` method after `yield` expression, which is always executed as teardown.

cheroot/test/test_ssl.py
jatalahd: The problematic fixture is now removed.

jatalahd:
@webknjaz ; Thanks for reviewing the code. I fixed all problems you mentioned (as well as I could). Added also change note and had to do the git squash to clean up the git history. Due to that, I had to force push and therefore your review comments disappeared from the "files changed" -tab. However I answered all those comments in this conversation-tab. Please let me know if there is something that still needs fixing.

cheroot/ssl/__init__.py
```diff
@@ -20,12 +20,14 @@ def __init__(
         private_key,
         certificate_chain=None,
         ciphers=None,
+        private_key_password=None,
```
webknjaz, line 24: I think any new args should be kw-only. A growing list of posargs makes it difficult to reason about the args being passed.
```suggestion
        *,
        private_key_password=None,
```

cheroot/ssl/builtin.py
```diff
@@ -76,10 +76,19 @@ def _loopback_for_cert_thread(context, server):
             ssl_sock.send(b'0000')
 
 
-def _loopback_for_cert(certificate, private_key, certificate_chain):
+def _loopback_for_cert(
+    certificate,
+    private_key,
+    certificate_chain,
+    private_key_password=None,
```
webknjaz, line 84:
```suggestion
    *,
    private_key_password=None,
```

cheroot/ssl/builtin.py
```diff
@@ -112,15 +121,25 @@ def _loopback_for_cert(certificate, private_key, certificate_chain):
         server.close()
 
 
-def _parse_cert(certificate, private_key, certificate_chain):
+def _parse_cert(
+    certificate,
+    private_key,
+    certificate_chain,
+    private_key_password=None,
```
webknjaz, line 130:
```suggestion
    *,
    private_key_password=None,
```

cheroot/ssl/builtin.py
```diff
@@ -208,6 +230,7 @@ def __init__(
         private_key,
         certificate_chain=None,
         ciphers=None,
+        private_key_password=None,
```
webknjaz, line 236:
```suggestion
        *,
        private_key_password=None,
```

cheroot/ssl/pyopenssl.py
```diff
@@ -293,12 +293,16 @@ class pyOpenSSLAdapter(Adapter):
...
+    private_key_password = None
+    """Optional passphrase for password protected private key."""
+
     def __init__(
         self,
         certificate,
         private_key,
         certificate_chain=None,
         ciphers=None,
+        private_key_password=None,
```
webknjaz, line 307:
```suggestion
        *,
        private_key_password=None,
```

cheroot/ssl/pyopenssl.pyi
```diff
@@ -31,9 +31,16 @@ class pyOpenSSLAdapter(Adapter):
...
         ciphers: Any | None = ...,
+        private_key_password: str | bytes | None = ...,
     ) -> None: ...
     def bind(self, sock): ...
     def wrap(self, sock): ...
+    def _password_callback(
+        self,
+        _password_max_length: int,
+        _verify_twice: bool,
+        password: bytes | str,
```
webknjaz: According to https://www.pyopenssl.org/en/stable/api/ssl.html#OpenSSL.SSL.Context.set_passwd_cb, this should be
```suggestion
        password: bytes | str | None,
```

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +333,28 @@ def wrap(self, sock):
         # closing so we can't reliably access protocol/client cert for the env
         return sock, self._environ.copy()
 
+    def _password_callback(
+        self,
+        _password_max_length,
+        _verify_twice,
+        password,
```
webknjaz, line 342: https://www.pyopenssl.org/en/stable/api/ssl.html#OpenSSL.SSL.Context.set_passwd_cb says the args are all positional:
```suggestion
        password,
        /,
```

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +333,28 @@ def wrap(self, sock):
         # closing so we can't reliably access protocol/client cert for the env
         return sock, self._environ.copy()
 
+    def _password_callback(
+        self,
+        _password_max_length,
```
webknjaz: Can the implementation consult this value and emit a warning if password exceeds it?
```suggestion
        password_max_length,
```

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +333,28 @@ def wrap(self, sock):
...
+
     def get_context(self):
         """Return an ``SSL.Context`` from self attributes.
 
         Ref: :py:class:`SSL.Context <pyopenssl:OpenSSL.SSL.Context>`
         """
         # See https://code.activestate.com/recipes/442473/
         c = SSL.Context(SSL.SSLv23_METHOD)
+        if self.private_key_password is not None:
+            c.set_passwd_cb(self._password_callback, self.private_key_password)
```
webknjaz: https://www.pyopenssl.org/en/stable/api/ssl.html#OpenSSL.SSL.Context.set_passwd_cb says
> Set the passphrase callback. This function will be called when a private key with a passphrase is loaded.

This probably means that it can be called unconditionally and pyOpenSSL will ignore it when the private key is non-encrypted, right?
```suggestion
        c.set_passwd_cb(self._password_callback, self.private_key_password)
```

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +333,28 @@ def wrap(self, sock):
...
+        _password_max_length,
+        _verify_twice,
+        password,
+    ):
+        """Pass a passphrase to password protected private key."""
+        if not password:
+            return b''
+        if not isinstance(password, bytes):
+            return password.encode('utf-8')
+        return password
```
webknjaz: This value is end-user provided. This means they may mess up and set it to `int` or something silly. I think we should not be assuming a `str` like this.

This should be a better fit:
```suggestion
        b_password = (
            password.encode('utf-8') if isinstance(password, str)
            else password if isinstance(password, bytes)
            else b''
        )

        if (password_length := len(b_password)) > password_max_length:
            _warn(
                UserWarning,
                f'User-provided password is {password_length} bytes long and will '
                f'be truncated since it exceeds the maximum of {password_max_length}.',
            )

        return b_password
```

(`from warnings import warn as _warn` at the top)

docs/changelog-fragments.d/752.feature.rst
```diff
@@ -0,0 +1,2 @@
+Added optional private key password argument to SSL adapters to support password-protected private keys.
```
webknjaz:
```suggestion
Added optional private key password argument to SSL adapters to support password-protected private keys
```

cheroot/ssl/__init__.py
jatalahd: Done in latest push. Added the asterisk also to the .pyi file

cheroot/ssl/builtin.py
jatalahd: Done in latest push, added the asterisk also to the builtin.pyi file.

cheroot/ssl/builtin.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Done in latest push.

cheroot/ssl/builtin.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.pyi
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Good idea, adapted your example code below to solve this.

cheroot/ssl/pyopenssl.py
jatalahd: Yes, it seems so. Modified the code like yo suggested and I added unit test for this use case.

cheroot/ssl/pyopenssl.py
jatalahd: Great suggestion, adopted this with minor style modifications (since the linter was giving errors). The latest push includes this.

docs/changelog-fragments.d/752.feature.rst
jatalahd: Done.

cheroot/ssl/pyopenssl.py
webknjaz: Yep, it's expected or in-browser coding. By the way, I always recommend accepting suggestions via GH UI since this lets you preserve proper attribution.

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +335,38 @@ def wrap(self, sock):
...
+        """Pass a passphrase to password protected private key."""
+        b_password = b''
+        if isinstance(password, str):
+            b_password = password.encode('utf-8')
+        elif isinstance(password, bytes):
+            b_password = password
+
+        if len(b_password) > password_max_length:
+            _warn(
+                f'User-provided password is {len(b_password)} bytes long and will '
```
webknjaz: I'd rather not have length computed twice. Also, complex expressions reduce readability. Could you move this to a var?

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +335,38 @@ def wrap(self, sock):
...
+                f'be truncated since it exceeds the maximum of {password_max_length}.',
+                UserWarning,
+                stacklevel=2,
```
webknjaz: Do you have an example of where this is pointing to on the stack? Any output? Trace?

cheroot/test/test_ssl.py
```diff
@@ -751,3 +800,142 @@ def test_http_over_https_error(
...
     assert expected_error_text in err_text
+
+
+@pytest.mark.parametrize(
+    'adapter_type',
+    (
+        'builtin',
+        'pyopenssl',
+    )
+    * 2,
```
webknjaz: What are you trying to achieve with this? Double test runs? This seems wrong.

cheroot/test/test_ssl.py
```diff
@@ -177,6 +196,36 @@ def tls_certificate_private_key_pem_path(tls_certificate):
...
+        encryption_algorithm=BestAvailableEncryption(
+            password=private_key_password.encode('utf-8'),
+        ),
+    )
+
+    with tls_certificate.private_key_pem.tempfile() as tf:
+        keyfile_temp_path = Path(tf)
+    keyfile_temp_path.write_bytes(encrypted_key_as_bytes)
+
+    yield keyfile_temp_path
```
webknjaz: This should probably be under `with` so that the `.tempile()` CM would delete it by itlsef. OTOH, do we actually need to call `.tempfile()`? what are you after? I'd rather make a key next to the original one, in the same dir.

cheroot/test/test_ssl.py
```diff
@@ -177,6 +196,36 @@ def tls_certificate_private_key_pem_path(tls_certificate):
...
+    key_as_bytes = tls_certificate.private_key_pem.bytes()
+    private_key_object = load_pem_private_key(
+        key_as_bytes,
+        password=None,
+        backend=default_backend(),
+    )
+    encrypted_key_as_bytes = private_key_object.private_bytes(
+        encoding=Encoding.PEM,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=BestAvailableEncryption(
```
webknjaz, line 212: Instead of best encryption, it's better to use the worst. Less compute would contributed to tests running faster.

cheroot/test/test_ssl.py
```diff
@@ -751,3 +800,142 @@ def test_http_over_https_error(
...
+
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    tls_adapter = tls_adapter_cls(
+        certificate=tls_certificate_chain_pem_path,
+        private_key=tls_certificate_passwd_private_key_pem_path,
+        private_key_password=secrets.choice(
+            [
+                private_key_password,
+                private_key_password.encode('utf-8'),
+            ],
```
webknjaz: Going for a random password will make the test inconsistent. We should check the predictable thing. Instead, you can parametrize the test to run against one variant at a time.

cheroot/test/test_ssl.py
```diff
@@ -137,6 +148,14 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
     return httpserver
 
 
+@pytest.fixture
```
webknjaz: There's no reason to test against multiple passwords. So this should run just once:
```suggestion
@pytest.fixture(scope='session')
```

OTOH, we could probably just hardcode a silly easter egg value and go with it.

cheroot/test/test_ssl.py
```diff
@@ -137,6 +148,14 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
...
+def private_key_password():
+    """Provide random 10 character password for private key."""
+    return ''.join(
+        secrets.choice(string.ascii_letters + string.digits) for _ in range(10)
+    )
```
webknjaz: Let's just keep it simple
```suggestion
    """Return a password for private key."""
    return 'криївка'
```

cheroot/test/test_ssl.py
```diff
@@ -751,3 +800,142 @@ def test_http_over_https_error(
...
+
+
+@pytest.mark.parametrize(
+    'adapter_type',
+    (
+        'builtin',
+        'pyopenssl',
+    ),
+)
+def test_ssl_adapters_key_without_password(
```
webknjaz: Why not combine this with the previous function through `parametrize`?

cheroot/test/test_ssl.py
```diff
@@ -751,3 +800,142 @@ def test_http_over_https_error(
...
+    tls_adapter = tls_adapter_cls(
+        certificate=tls_certificate_chain_pem_path,
+        private_key=tls_certificate_passwd_private_key_pem_path,
+        private_key_password=false_password,
+    )
+
+    httpserver.ssl_adapter = tls_adapter
+
+    with pytest.raises(OpenSSL.SSL.Error, match=r'.+bad decrypt.+'):
+        httpserver.prepare()
```
webknjaz, line 929: Why does pyOpenSSL fail in a different place?

cheroot/test/test_ssl.py
webknjaz, line 1: Could you also test that password length warning with `pytest.warns()`?

cheroot/test/test_ssl.py
```diff
@@ -751,3 +800,142 @@ def test_http_over_https_error(
...
+    tls_certificate_passwd_private_key_pem_path,
+    adapter_type,
+):
+    """Check that builtin ssl-adapter initialization fails when wrong private key password given."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    with pytest.raises(ssl.SSLError, match=r'\[SSL\] PEM.+'):
+        tls_adapter_cls(
+            certificate=tls_certificate_chain_pem_path,
+            private_key=tls_certificate_passwd_private_key_pem_path,
+            private_key_password=str(uuid.uuid4()),
```
webknjaz: We could hardcode a long value here too. Like

```suggestion
            private_key_password='x' * 256,
```

cheroot/ssl/pyopenssl.py
jatalahd: Separate variable added in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Here is the trace when the password is too long. The warning is given first and then the exception, because the password did not match in this case:

```python-traceback
/Users/xxx/lib/python3.11/site-packages/OpenSSL/SSL.py:805: UserWarning: User-provided password is 2048 bytes long and will be truncated since it exceeds the maximum of 1024.
  return callback(size, verify, self._passphrase_userdata)
Exception in thread Thread-1 (run):
  File "/Users/xxx/temp/xxx/xxx.py", line 53, in run
    self.server.start()
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/server.py", line 1966, in start
    self.prepare()
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/server.py", line 1880, in prepare
    self.bind(af, socktype, proto)
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/server.py", line 2016, in bind
    sock = self.prepare_socket(
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/server.py", line 2176, in prepare_socket
    sock = ssl_adapter.bind(sock)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/ssl/pyopenssl.py", line 326, in bind
    self.context = self.get_context()
                   ^^^^^^^^^^^^^^^^^^
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/cheroot/ssl/pyopenssl.py", line 371, in get_context
    c.use_privatekey_file(self.private_key)
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/OpenSSL/SSL.py", line 1018, in use_privatekey_file
    self._raise_passphrase_exception()
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/OpenSSL/SSL.py", line 994, in _raise_passphrase_exception
    _raise_current_error()
  File "/Users/xxx/.local/share/virtualenvs/xxx/lib/python3.11/site-packages/OpenSSL/_util.py", line 57, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.SSL.Error: [('Provider routines', '', 'bad decrypt'), ('PKCS12 routines', '', 'pkcs12 cipherfinal error'), ('Provider routines', '', 'bad decrypt'), ('PKCS12 routines', '', 'pkcs12 cipherfinal error'), ('SSL routines', '', 'PEM lib')]
```

cheroot/test/test_ssl.py
jatalahd: Removed double run in the latest push.

cheroot/test/test_ssl.py
jatalahd: Moved under `with`. You commented on this already 4 months ago, and I managed to get this much cleaner since then. Now it uses the base certificate provided by trustme and only encrypts the output. After this change the yield part is similar to other certificate fixtures. I cannot do any better than this. All the certs are stored in the same temp directory and cleaned after test.

cheroot/test/test_ssl.py
jatalahd: There is no other viable option for `encryption_algorithm` in this part. Only other option is NoEncryption(), which does not serve the purpose.

cheroot/test/test_ssl.py
jatalahd: Parameterized the test with fixed `str` and `bytes` passwords

cheroot/test/test_ssl.py
jatalahd: Done as suggested in the latest push.

cheroot/test/test_ssl.py
jatalahd: Done as suggested in the latest push.

cheroot/test/test_ssl.py
jatalahd: The builtin adapter prepares the server env with a call to already in `__init__` while pyopenssl uses the cert only when doing `bind`. Also it seems that the builtin adapter makes two calls to load the certs, so when starting the server manually, user needs to enter the password twice. So there is a clear difference how builtin and openssl adapters behave in this sense. I am not proficient enough to start making changes to unify the behaviour.

cheroot/test/test_ssl.py
jatalahd: Done in latest push by parametering the last test case.

cheroot/test/test_ssl.py
jatalahd: Implemented in the latest push as suggested.

cheroot/test/test_ssl.py
jatalahd: Combined in the latest push.

cheroot/ssl/pyopenssl.py
webknjaz: Would `stacklevel=1` read better?

cheroot/test/test_ssl.py
webknjaz: Why? Is this some sort of a documented limitation?

cheroot/test/test_ssl.py
```diff
@@ -178,6 +193,32 @@ def tls_certificate_private_key_pem_path(tls_certificate):
...
+    )
+    encrypted_key_as_bytes = private_key_object.private_bytes(
+        encoding=Encoding.PEM,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=BestAvailableEncryption(
+            password=private_key_password.encode('utf-8'),
+        ),
+    )
+
+    with tls_certificate.private_key_pem.tempfile() as tf:
```
webknjaz: It seems pointless to make the library write something into file only to write on top of it. Let's just use the built-in `tmp_path` fixture which would also save us some indentation.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +798,124 @@ def test_http_over_https_error(
...
+        ('builtin', 'tls_certificate_passwd_private_key_pem_path', False),
+        ('builtin', 'tls_certificate_private_key_pem_path', True),
+        ('builtin', 'tls_certificate_private_key_pem_path', False),
+        ('pyopenssl', 'tls_certificate_passwd_private_key_pem_path', True),
+        ('pyopenssl', 'tls_certificate_passwd_private_key_pem_path', False),
+        ('pyopenssl', 'tls_certificate_private_key_pem_path', True),
+        ('pyopenssl', 'tls_certificate_private_key_pem_path', False),
+    ),
+)
+# pylint: disable-next=R0917
```
webknjaz: Always use human-readable identifiers in suppressions. These numbers and a letter don't explain context even a little bit. When there's not enough context or no way to avoid numbers, a code comment explaining when to remove it should be added.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +798,124 @@ def test_http_over_https_error(
...
+    ('adapter_type', 'key_file', 'password_as_bytes'),
+    (
+        ('builtin', 'tls_certificate_passwd_private_key_pem_path', True),
+        ('builtin', 'tls_certificate_passwd_private_key_pem_path', False),
+        ('builtin', 'tls_certificate_private_key_pem_path', True),
+        ('builtin', 'tls_certificate_private_key_pem_path', False),
+        ('pyopenssl', 'tls_certificate_passwd_private_key_pem_path', True),
+        ('pyopenssl', 'tls_certificate_passwd_private_key_pem_path', False),
+        ('pyopenssl', 'tls_certificate_private_key_pem_path', True),
+        ('pyopenssl', 'tls_certificate_private_key_pem_path', False),
```
webknjaz: The purpose of these is difficult to grasp. Use explicit test ids so its visible what the intention is.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +798,124 @@ def test_http_over_https_error(
...
+    key_file,
+    password_as_bytes,
+):
+    """Check that server starts using ssl adapter with password-protected private key."""
+    httpserver = HTTPServer(
+        bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT),
+        gateway=Gateway,
+    )
+
+    key_file = request.getfixturevalue(key_file)
```
webknjaz: I don't understand the motivation here.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +798,124 @@ def test_http_over_https_error(
...
+    tls_adapter = tls_adapter_cls(
+        certificate=tls_certificate_chain_pem_path,
+        private_key=tls_certificate_passwd_private_key_pem_path,
+        private_key_password=false_password,
+    )
+
+    httpserver.ssl_adapter = tls_adapter
+
+    with expected_warn:
+        with pytest.raises(OpenSSL.SSL.Error, match=r'.+bad decrypt.+'):
```
webknjaz: I believe these can be collapsed into a single with-block.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +798,124 @@ def test_http_over_https_error(
...
+        (
+            'pyopenssl',
+            '@' * 2048,
+            pytest.warns(
+                UserWarning,
+                match=r'User-provided password is 2048 bytes.+',
+            ),
+        ),
+    ),
+    ids=('false_password', 'integer_password', 'too_long_password'),
```
```suggestion
    ids=('false-password', 'integer-password', 'too-long-password'),
```

cheroot/ssl/pyopenssl.py
jatalahd: Differences between 1 and 2:
```console
stacklevel=1:
/pyopenssl.py:354: UserWarning: User-provided password is 2048 bytes long and will be truncated since it exceeds the maximum of 1024.
  _warn(
```
```console
stacklevel=2:
/SSL.py:805: UserWarning: User-provided password is 2048 bytes long and will be truncated since it exceeds the maximum of 1024.
  return callback(size, verify, self._passphrase_userdata)
```
Changed to use 1 for now.

cheroot/test/test_ssl.py
jatalahd: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/rsa/#key-serialization

I tried mocking the BestAvailableEncryption and creating custom encryption methods, but no luck. I took timings on my machine, it takes max 3 milliseconds to execute this part of code, typically 1 millisecond. These tests on average take total of 10 to 20 milliseconds to execute, they are far faster than other tests in this test suite.

cheroot/test/test_ssl.py
jatalahd: Changed to use tmp_path, but I am not happy with it, it leaves the files on my machine. The earlier versions cleaned the files nicely.

cheroot/test/test_ssl.py
jatalahd: Changed to human readable.

cheroot/test/test_ssl.py
jatalahd: Added test ids, also changed the parameterization values.

cheroot/test/test_ssl.py
jatalahd: The motivation was to get the private key file path from the parametrization. This was the only way I could get it working. However, this is now removed, since I changed the parameterization.

cheroot/test/test_ssl.py
jatalahd: Done in latest push. However, the linter made it look quite nasty.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
```diff
@@ -328,13 +335,39 @@ def wrap(self, sock):
...
 
+    def _password_callback(
+        self,
+        password_max_length,
+        _verify_twice,
+        password,
+        /,
+    ):
+        """Pass a passphrase to password protected private key."""
+        b_password = b''
```
webknjaz: Enough time passed since the last review that I forgot what this means and almost started suggesting an explicit `else:` branch. Let's at least have a comment here.
```suggestion
        b_password = b''  # returning a falsy value communicates an error
```

cheroot/test/test_ssl.py
webknjaz: Okay, thanks for researching this! Marking this thread as resolved.

cheroot/test/test_ssl.py
```diff
@@ -178,6 +192,34 @@ def tls_certificate_private_key_pem_path(tls_certificate):
...
+
+    encrypted_key_as_bytes = private_key_object.private_bytes(
+        encoding=Encoding.PEM,
+        format=PrivateFormat.PKCS8,
+        encryption_algorithm=BestAvailableEncryption(
+            password=private_key_password.encode('utf-8'),
+        ),
+    )
+
+    key_file = tmp_path / 'key.pem'
```
webknjaz: Since these files remain on disk post-testing (for troubleshooting), it's a good idea to use more specific names:
```suggestion
    key_file = tmp_path / 'encrypted-private-key.pem'
```

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,142 @@ def test_http_over_https_error(
...
+        'builtin-encrypted-key-with-str-password',
+        'builtin-encrypted-key-with-bytes-password',
+        'builtin-non-encrypted-key-with-str-password',
+        'builtin-non-encrypted-key-with-bytes-password',
+        'pyopenssl-encrypted-key-with-str-password',
+        'pyopenssl-encrypted-key-with-bytes-password',
+        'pyopenssl-non-encrypted-key-with-str-password',
+        'pyopenssl-non-encrypted-key-with-bytes-password',
+    ),
+)
```
webknjaz, line 814: This can be simplified to let pytest multiply the factors internally. Plus bytes/str IDs were swapped.
```suggestion
@pytest.mark.parametrize('adapter_type', ('builtin', 'pyopenssl'))
@pytest.mark.parametrize(
    'encrypted_key',
    (True, False),
    ids=('encrypted-key', 'unencrypted-key'),
)
@pytest.mark.parametrize(
    'password_as_bytes',
    (True, False),
    ids=('with-bytes-password', 'with-str-password'),
)
```

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,142 @@ def test_http_over_https_error(
...
+
+    httpserver.ssl_adapter = tls_adapter
+    httpserver.prepare()
+
+    assert httpserver.ready
+    assert httpserver.requests._threads
+    for thr in httpserver.requests._threads:
+        assert thr.ready
+
+    httpserver.stop()
```
webknjaz: We could make an HTTPS request somewhere here to proove that TLS works.

cheroot/test/test_ssl.py
webknjaz: Okay, that's fair. Marking this as resolved.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,142 @@ def test_http_over_https_error(
...
+        (
+            'pyopenssl',
+            '@' * 2048,
+            pytest.warns(
+                UserWarning,
+                match=r'User-provided password is 2048 bytes.+',
+            ),
+        ),
+    ),
+    ids=('false-password', 'integer-password', 'too-long-password'),
```
webknjaz: I think "incorrect" is easier to put in the context than "false". At least, it sounds more obvious to me.
```suggestion
    ids=('incorrect-password', 'integer-password', 'too-long-password'),
```

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,142 @@ def test_http_over_https_error(
...
+    ('builtin',),
+)
+def test_builtin_adapter_with_false_key_password(
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    adapter_type,
+):
+    """Check that builtin ssl-adapter initialization fails when wrong private key password given."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    with pytest.raises(ssl.SSLError, match=r'\[SSL\] PEM.+'):
```
webknjaz: Is this the beginning of the string?
```suggestion
    with pytest.raises(ssl.SSLError, match=r'^\[SSL\] PEM.+'):
```

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,142 @@ def test_http_over_https_error(
...
+            '837550fd-bcb9-4320-87e6-09de6456b09',
+            contextlib.nullcontext(),
+        ),
+        ('pyopenssl', 555555, contextlib.nullcontext()),
+        (
+            'pyopenssl',
+            '@' * 2048,
+            pytest.warns(
+                UserWarning,
+                match=r'User-provided password is 2048 bytes.+',
```
webknjaz: I like being explicit about the start marker.
```suggestion
                match=r'^User-provided password is 2048 bytes.+',
```

cheroot/test/test_ssl.py
webknjaz: Keeping the files in tmp dirs is by design: https://docs.pytest.org/en/8.3.x/how-to/tmp_path.html#temporary-directory-location-and-retention. `tmp_path_retention_count` and `tmp_path_retention_policy` can be tweaked if you really need, in a centralized manner. But in general, it's not such a good idea to turn this off. They are being cleaned but not during the current session run — the following sessions will clean files from previous sessions.
Having temporary files on disk allows troubleshooting tests outside the pytest environments as it's easier to inspect things when they aren't wrapped in many additional layers/runtimes.

cheroot/ssl/pyopenssl.py
webknjaz: Yeah, I think this is the right thing to do. Marking this as resolved.

cheroot/test/test_ssl.py
jatalahd: OK, thanks for clarifying,

cheroot/ssl/pyopenssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Thank you, this was very helpful. Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push. Simplified the httpserver assert to only check server ready and after that added https request verification using the tls_http_server -fixture.

jatalahd:
@webknjaz Provided improvements as you suggested. I hope this is OK now.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,141 @@ def test_http_over_https_error(
...
+        certificate=tls_certificate_chain_pem_path,
+        private_key=key_file,
+        private_key_password=key_pass,
+    )
+
+    httpserver.ssl_adapter = tls_adapter
+
+    httpserver.prepare()
+    assert httpserver.ready
+    httpserver.stop()
```
webknjaz: @jatalahd why are you stopping the server before sending requests its way?

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,141 @@ def test_http_over_https_error(
...
+    tls_certificate_private_key_pem_path,
+    adapter_type,
+    encrypted_key,
+    password_as_bytes,
+):
+    """Check server start and response when using ssl adapter with password-protected private key."""
+    httpserver = HTTPServer(
+        bind_addr=(ANY_INTERFACE_IPV4, EPHEMERAL_PORT),
+        gateway=Gateway,
+    )
```
webknjaz: Looks like this could be moved after the adapter setup and replaced with `tls_http_server()`, right?

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,141 @@ def test_http_over_https_error(
...
+
+    httpserver.ssl_adapter = tls_adapter
+
+    httpserver.prepare()
+    assert httpserver.ready
+    httpserver.stop()
+
+    interface, _host, port = _get_conn_data(
+        tls_http_server(httpserver.bind_addr, tls_adapter).bind_addr,
```
webknjaz: Why aren't you using the same server that was used to check TLS adapter loading the certs? This feels like having two different tests under the same roof. My idea was that the same server that loaded the certificates would be used for HTTPS requests and this working would be the final assertion that all's good.

webknjaz: It seems this should be enough:
```suggestion
    key_file = (
        tls_certificate_passwd_private_key_pem_path
        if encrypted_key
        else tls_certificate_private_key_pem_path
    )
    key_pass = (
        private_key_password.encode('utf-8')
        if password_as_bytes
        else private_key_password
    )

    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
    tls_adapter = tls_adapter_cls(
        certificate=tls_certificate_chain_pem_path,
        private_key=key_file,
        private_key_password=key_pass,
    )

    httpserver = tls_http_server(httpserver.bind_addr, tls_adapter)

    interface, _host, port = _get_conn_data(
        httpserver.bind_addr,
```

cheroot/test/test_ssl.py
jatalahd: Removed this part in the lastest push.

cheroot/test/test_ssl.py
jatalahd: Refactored in the latest push, now using just one call to tls_http_server fixture.

cheroot/test/test_ssl.py
jatalahd: True, I did not understand what you wanted. In my opinion there should be two separate tests, one that checks that the server starts and one that actually makes the request to the running server. This is now fixed in the latest push.

jatalahd: Done as you suggested, but with a slight change that calling only tls_http_server inside the _get_conn_data().

jatalahd:
@webknjaz Yet another improved push provided.

cheroot/test/test_ssl.py
webknjaz: I guess that's fine. Although, nested constructor calls contribute to inline complexity so it's best not to overuse them.

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,134 @@ def test_http_over_https_error(
...
+    tls_http_server,
+    tls_ca_certificate_pem_path,
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    tls_certificate_private_key_pem_path,
+    adapter_type,
+    encrypted_key,
+    password_as_bytes,
+):
+    """Check server start and response when using ssl adapter with password-protected private key."""
```
webknjaz: Response is an implementation detail of how the assertion is done. It doesn't need to be in the docstring summary line. Let's communicate this.
```suggestion
    """Check server decrypts private TLS keys with password as bytes or str."""
```

cheroot/test/test_ssl.py
```diff
@@ -178,6 +192,34 @@ def tls_certificate_private_key_pem_path(tls_certificate):
         yield cert_key_pem
 
 
+@pytest.fixture
+def tls_certificate_passwd_private_key_pem_path(
+    tls_certificate,
+    private_key_password,
+    tmp_path,
+):
+    """Provide a certificate private key PEM file path via fixture."""
```
webknjaz: No need to duplicate that this is a fixture — the decorator isn't that far from this line. Additionally, I bet it'd look weird in the `pytest --fixtures` output.
```suggestion
    """Return a certificate private key PEM file path."""
```

cheroot/test/test_ssl.py
jatalahd: Used the suggested docstring in latest push.

cheroot/test/test_ssl.py
jatalahd: Used the suggested docstring in latest push.

webknjaz (approved):
Perfect, thanks!

cheroot/test/test_ssl.py
```diff
@@ -757,3 +799,134 @@ def test_http_over_https_error(
...
+        certificate=tls_certificate_chain_pem_path,
+        private_key=tls_certificate_passwd_private_key_pem_path,
+        private_key_password=false_password,
+    )
+
+    httpserver.ssl_adapter = tls_adapter
+
+    with expected_warn, pytest.raises(
+        OpenSSL.SSL.Error,
+        match=r'.+bad decrypt.+',
```
webknjaz, line 927: @jatalahd this test turned out to be flaky. See https://github.com/cherrypy/cheroot/actions/runs/19020536310/job/54314981390#step:15:272 where the error string was `"[('DECODER routines', '', 'unsupported'), ('asn1 encoding routines', '', 'too long'), ('asn1 encoding routines', '', 'bad object header'), ('asn1 encoding routines', '', 'nested asn1 error'), ('PKCS12 routines', '', 'decode error'), ('SSL routines', '', 'PEM lib')]"` which your regex doesn't match.

Could you send a follow-up PR fixing `cheroot/test/test_ssl.py::test_openssl_adapter_with_false_key_password[too-long-password]` — looks like the error is sometimes happening with passwords that are too long. Perpahs, encode the regex in `parametrize()` too, making the `too-long-password` param expect both error variations?

webknjaz:
@jatalahd I've released this under v11.1.0:

* https://github.com/cherrypy/cheroot/actions/runs/19020536310
* https://github.com/cherrypy/cheroot/discussions/789
* https://github.com/cherrypy/cheroot/releases/tag/v11.1.0
* https://pypi.org/project/cheroot/11.1.0/

---

Though, one test got flaky and needs a follow-up: https://github.com/cherrypy/cheroot/pull/752/files#r2485127496.
Could you handle that, please?

---

Now, this allows sending a PR in the CherryPy repo per https://github.com/cherrypy/cherrypy/issues/1583#issuecomment-287476777.

PR #755, opened 2025-09-17
Fixed circular reference in HTTPConnection

### LINKED ISSUE #746: Memory leak on connection close due to circular reference
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request

In an environment where the automatic garbage collection is disabled by default and `gc.collect(2)` is called manually (in this case, once every 10 minutes), there has been a considerable increase in the size of the garbage collector.
After analysis, this is the cause:

https://github.com/cherrypy/cheroot/blob/1391e717daf7a5c233c2713d82e6d7380cb5345a/cheroot/server.py#L1264-L1271

Reassigning a method to an object creates a circular reference.

This feature seems to be here for caching UNIX sockets credentials. But we use cheroot only for internet connection so this cache system is useless. And even for a UNIX server, this piece of code creates some problems.

💡 **To Reproduce**

1. Run a server with automatic garbage collection disabled and log number of objects in garbage ( using `len(gc.get_objets())` ) at regular interval.
2. Have a script which makes a lot of temporary connections ( without `Keep-Alive` ).
3. We can see the increase in the number of items tracked by the garbage collector.

💡 **Expected behavior**

No memory leak.

📋 **Details**

📋 **Environment**

- Cheroot version: 10.0.1
- CherryPy version: None
- Python version: 3.11.11
- OS: Linux
- Browser: all

📋 **Additional context**

francis-clairicia:

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves #746

The UNIX socket server's caching system for credentials involves a circular reference to `HTTPConnection`, even for an INET server. 

The cache system is still in place, but now uses private attributes.

📋 **Other information**:

📋 **Contribution checklist:**

[related squash]:

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/755)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -79,7 +79,6 @@ import threading
 import time
 import traceback as traceback_
 import urllib.parse
-from functools import lru_cache
 
 from . import __version__, connections, errors
 from ._compat import IS_PPC, bton
@@ -1321,10 +1320,8 @@ class HTTPConnection:
         self.peercreds_enabled = self.server.peercreds_enabled
         self.peercreds_resolve_enabled = self.server.peercreds_resolve_enabled
 
-        # LRU cached methods:
-        # Ref: https://stackoverflow.com/a/14946506/595220
-        self.resolve_peer_creds = lru_cache(maxsize=1)(self.resolve_peer_creds)
-        self.get_peer_creds = lru_cache(maxsize=1)(self.get_peer_creds)
+        self._peer_creds_resolved = None
+        self._socket_peer_creds_cache = None
 
     def communicate(self):  # noqa: C901  # FIXME
         """Read each request and respond appropriately.
@@ -1434,7 +1431,7 @@ class HTTPConnection:
             # Apache does, but not today.
             pass
 
-    def get_peer_creds(self):  # LRU cached on per-instance basis, see __init__
+    def get_peer_creds(self):
         """Return the PID/UID/GID tuple of the peer socket for UNIX sockets.
 
         This function uses SO_PEERCRED to query the UNIX PID, UID, GID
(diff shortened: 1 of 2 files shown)


webknjaz (commented):
Looks like this implementation adds complexity by putting cache management into multiple places. Can you find a decorator-based solution? It might be okay to add a dependency if there's a good implementation in the wild.

webknjaz:
@francis-clairicia would using a `@cached_property` work? This would mean having a new API and we could keep the old methods uncached for compat.

webknjaz:
@francis-clairicia I didn't realize it was writable. But here's how we can make it read-only. Put this helper into a `_functools.py`, so that it lives in a dedicated space. Add a `test/_functools_test.py` to test the setter raising an error so that it all's got good test coverage.

```pycon
>>> import functools as _ft
>>> class readonly_cached_property(_ft.cached_property):
...     def __set__(self, instance, value):
...         attr_name = self.attrname
...         raise AttributeError(f'Attribute {attr_name} is read-only and cannot be set.')
>>> class AttrsCls:
...     @cached_property
...     def cp(self):
...         return 42
...     @readonly_cached_property
...     def rocp(self):
...         return 84
>>> attro = AttrsCls()
>>> attro.rocp
84
>>> attro.cp
42
>>> attro.rocp = 3
  File "<python-input-42>", line 1, in <module>
    attro.rocp = 3
    ^^^^^^^^^^
  File "<python-input-37>", line 4, in __set__
    raise AttributeError(f'Attribute {attr_name} is read-only and cannot be set.')
AttributeError: Attribute rocp is read-only and cannot be set.
```

Additionally, make the change note a bit more specific. Referencing an abstract unnamed circular ref does not help a reader understand the context and the effect this has on them.

francis-clairicia:
@webknjaz I forgot this one. You can take over to use another approach.

PR #764, opened 2025-09-21
Fixes for WantWriteError and WantReadError handling

julianz-:
This adds improved handling for WantWriteError and WantReadError so that it retries writing the full buffer in BufferedWriter and retries reading in BufferedReader after a short delay allowing for network buffers to settle. The retry logic also makes use of new changes in PyOpenSSL v25.2.0 and Cryptography v45.0.7 that allows for a moving buffer as discussed in #245.


cheroot/test/test_ssl.py
```diff
@@ -625,6 +628,41 @@ def test_ssl_env(  # noqa: C901  # FIXME
     )
 
 
+class TestBufferedWriterSSLWantErrors:
```
webknjaz: Please avoid classes to group tests — modules already namespace them. Use fixtures for setup/teardown instead.

cheroot/test/test_ssl.py
```diff
@@ -10,13 +10,16 @@
 import threading
 import time
 import traceback
+from unittest import mock
```
webknjaz: Instead of importing the stdlib module directly, use the `mocker` fixture from `pytest-mock` or the builtin `monkeypatch` fixture where applicable.

cheroot/test/test_ssl.py
```diff
@@ -625,6 +628,41 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
+    def test_want_write_error_retry(self):
+        """Test that WantWriteError causes retry with same data."""
+        test_data = b'hello world'
+
+        self.mock_raw.write.side_effect = [
+            OpenSSL.SSL.WantWriteError(),
+            len(test_data),
+        ]
+
+        result = self.writer.write(test_data)
```
webknjaz: Could you name the vars meaningfully? Stuff like $output, $result, $thing is too generic to make sense of.

cheroot/test/test_ssl.py
webknjaz: This also causes linting to fail.

webknjaz:
@julianz- do you think your thing will let us make use of https://github.com/pyca/pyopenssl/pull/954 ?

cheroot/test/test_ssl.py
julianz-: Good point. Have revised.

cheroot/test/test_ssl.py
julianz-: Thanks for the suggestion - have replaced accordingly.

cheroot/test/test_ssl.py
julianz-: Agreed! Replaced.

julianz-:
> @julianz- do you think your thing will let us make use of [pyca/pyopenssl#954](https://github.com/pyca/pyopenssl/pull/954) ?

I made a comment about the code there but did you want me to add in your change to this PR?

cheroot/workers/threadpool.pyi
webknjaz, line 1: Could you move unrelated typing updates into a separate PR? Ideally, we should just migrate the annotations from stubs into Python modules. But that'd be another effort. I think `pyrefly autotype` might help — I've tried it on a portion of a different code base and it was able to inject some types (that required post-processing, though).

cheroot/test/test_ssl.py
```diff
@@ -625,6 +627,47 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
 
 
+@pytest.fixture
+def ssl_writer(mocker):
+    """Return a BufferedWriter instance with a mocked raw socket."""
+    mock_raw = mocker.Mock()
+    mock_raw.closed = False
+    writer = BufferedWriter(mock_raw)
+
+    writer.mock_raw = mock_raw
```
webknjaz: No need to assign it to this object — just add another fixture returning the prepared mock and inject it into this fixture and into the test functions.

cheroot/test/test_ssl.py
webknjaz, line 1: Any chance to have an integration-style test?

cheroot/makefile.py
```diff
@@ -45,9 +56,15 @@ def __init__(self, sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
...
     def read(self, *args, **kwargs):
         """Capture bytes read."""
-        val = super().read(*args, **kwargs)
-        self.bytes_read += len(val)
-        return val
+        while True:
+            try:
+                val = super().read(*args, **kwargs)
+                self.bytes_read += len(val)
+                return val
```
webknjaz, line 78: Avoid having more than one instruction in the try-except block. Move the rest into `else`.

cheroot/makefile.py
```diff
@@ -45,9 +56,15 @@ def __init__(self, sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
 
     def read(self, *args, **kwargs):
         """Capture bytes read."""
-        val = super().read(*args, **kwargs)
-        self.bytes_read += len(val)
-        return val
+        while True:
```
webknjaz: Can this cause infinite loops?

cheroot/makefile.py
```diff
@@ -31,7 +34,15 @@ def _flush_unlocked(self):
                 # so perhaps we should conditionally wrap this for perf?
                 n = self.raw.write(bytes(self._write_buf))
             except io.BlockingIOError as e:
-                n = e.characters_written
+                n = e.characters_writteni
```
webknjaz: typo?
```suggestion
                n = e.characters_written
```

webknjaz:
> > @julianz- do you think your thing will let us make use of [pyca/pyopenssl#954](https://github.com/pyca/pyopenssl/pull/954) ?
> 
> I made a comment about the code there but did you want me to add in your change to this PR?

That's a long-standing idea. And I've lost some context over the years. But I think I wanted us to use `sendall()` instead of reimplementing it on our side. Since you've spend substantial effort in that project, I imagine you're the best person to judge whether this is doable in general and whether this sounds like it should be in the scope of this PR or not.

cheroot/test/test_ssl.py
julianz-: Have revised.

cheroot/test/test_ssl.py
julianz-: Have added one. See what you think.

cheroot/makefile.py
julianz-: Revised!

cheroot/makefile.py
julianz-: Revised - limiting to 10 iterations.  Wasn't sure about that but see what you think.

cheroot/makefile.py
julianz-: Yes! Fixed.

julianz-:
> > > @julianz- do you think your thing will let us make use of [pyca/pyopenssl#954](https://github.com/pyca/pyopenssl/pull/954) ?
> > 
> > 
> > I made a comment about the code there but did you want me to add in your change to this PR?
> 
> That's a long-standing idea. And I've lost some context over the years. But I think I wanted us to use `sendall()` instead of reimplementing it on our side. Since you've spend substantial effort in that project, I imagine you're the best person to judge whether this is doable in general and whether this sounds like it should be in the scope of this PR or not.

I think the change there is pretty similar so it might be possible though maybe we should just get this PR done first?

webknjaz:
> maybe we should just get this PR done first?

Yes, let's do so.

cheroot/test/test_ssl.py
```diff
@@ -625,6 +627,145 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
+
+
+def test_want_write_error_integration(ssl_writer_integration):
+    """Integration test for SSL writer handling of WantWriteError."""
+    writer, mock_write = ssl_writer_integration
+    test_data = b'integration test data'
+    successful_write_length = len(test_data)
+
+    # Determine the failure mechanism
+    if adapter_type == 'pyopenssl':
```
webknjaz: Did you forget to actually inject the fixture value into the test function?

cheroot/test/test_ssl.py
```diff
@@ -625,6 +627,145 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
+        OpenSSL.SSL.WantReadError(),
+        len(test_data),
+    ]
+
+    bytes_written = ssl_writer.write(test_data)
+    assert bytes_written == len(test_data)
+
+
+@pytest.fixture(
+    params=['builtin', 'pyopenssl'],
```
webknjaz: We normally have tuples for these here
```suggestion
    params=('builtin', 'pyopenssl'),
```

cheroot/test/test_ssl.py
```diff
@@ -625,6 +627,145 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
+            autospec=True,
+        )
+        raw_io_object.closed = False
+        raw_io_object.writable = mocker.Mock(return_value=True)
+
+    # Return both the writer and the specific mock assertion target
+    return BufferedWriter(raw_io_object), raw_io_object.write
+
+
+def test_want_write_error_integration(ssl_writer_integration):
```
```suggestion
def test_want_write_error_integration(adapter_type, ssl_writer_integration):
```

cheroot/test/test_ssl.py
webknjaz: *(pylint caught this error: https://github.com/cherrypy/cheroot/actions/runs/18078538540/job/51438566760?pr=764#step:16:63)*

cheroot/test/test_ssl.py
```diff
@@ -625,6 +627,145 @@ def test_ssl_env(  # noqa: C901  # FIXME
...
+        raw_io_object = tls_adapter
+        raw_io_object.write = mocker.Mock(
+            name='builtin_adapter_write',
+            autospec=True,
+        )
+        raw_io_object.closed = False
+        raw_io_object.writable = mocker.Mock(return_value=True)
+
+    # Return both the writer and the specific mock assertion target
+    return BufferedWriter(raw_io_object), raw_io_object.write
```
webknjaz: Just have a fixture returning `raw_io_object` and instantiation + accessing the callable can happen outside w/o operating tuples of things. You could add fixtures returning these separately, if need be.

Actually, you could have a fixture setting up side effects instead of having that in a test. Call it `first_write_side_effect` and it'd be a good abstraction.

pyproject.toml
```diff
@@ -72,6 +72,9 @@ file = "README.rst"
 content-type = "text/x-rst"
 
 [project.optional-dependencies]
+ssl = [
```
webknjaz: Adding new deps metadata isn't a bugfix and shouldn't be included in this PR.

cheroot/makefile.py
```diff
@@ -45,9 +56,22 @@ def __init__(self, sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
...
+        while True:
+            try:
+                val = super().read(*args, **kwargs)
+            except (SSL.WantReadError, SSL.WantWriteError):
+                attempts += 1
+                if attempts >= MAX_ATTEMPTS:
+                    # Raise an error if max retries reached
+                    raise TimeoutError(
+                        'Max retries exceeded while waiting for data.',
+                    )
```
webknjaz:
```suggestion
                    ) from ssl_want_error
```

cheroot/makefile.py
```diff
@@ -45,9 +56,22 @@ def __init__(self, sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
...
         """Capture bytes read."""
-        val = super().read(*args, **kwargs)
-        self.bytes_read += len(val)
-        return val
+        MAX_ATTEMPTS = 10
+        attempts = 0
+        while True:
+            try:
+                val = super().read(*args, **kwargs)
+            except (SSL.WantReadError, SSL.WantWriteError):
```
webknjaz:
```suggestion
            except (SSL.WantReadError, SSL.WantWriteError) as ssl_want_error:
```

cheroot/makefile.py
webknjaz: Not sure how I feel about it in a low-level method. But if we were to do this, I'd probably go for a `for`-loop with `range()`.

cheroot/makefile.py
```diff
@@ -32,6 +35,14 @@ def _flush_unlocked(self):
...
             except io.BlockingIOError as e:
                 n = e.characters_written
+            except (
+                SSL.WantReadError,
+                SSL.WantWriteError,
+                SSL.WantX509LookupError,
+            ):
+                # these errors require retries with the same data
+                # if some data has already been written
+                n = 0
```
webknjaz: perhaps, just skip that `del` altogether?
```suggestion
                continue
```

julianz-: Agreed

cheroot/makefile.py
julianz-: Agreed

cheroot/makefile.py
julianz-: Agreed

cheroot/makefile.py
julianz-: Ok will do

cheroot/test/test_ssl.py
julianz-: Yes my bad. Strangely, the test passed even without the parameter I think in part because the fixture ssl_writer_integration() has it as a parameter and this picked it up from the fixture adapter_type(). But still not sure how if adapter_type == 'pyopenssl' succeeded in this function.

julianz-:
Some tests fail because the timeouts on SSL tests are too short. 
In conftest.py, `computed_timeout` needs to be higher. 0.5 seems to work fairly well.
I didn't add to this PR. Will create another one for this.

def http_request_timeout():
    """Return a common HTTP request timeout for tests with queries."""
    computed_timeout = 0.1

cheroot/makefile.py
```diff
@@ -3,6 +3,9 @@
 # prefer slower Python-based io module
 import _pyio as io
 import socket
+import time
+
+from OpenSSL import SSL
```
webknjaz, line 8: Looks like this will have to become conditional/guarded since pyOpenSSL is an optional dependency. It might be a good idea to add infra for running the tests w/o optional deps but that certainly out of the scope of this PR.

cheroot/makefile.py
```diff
@@ -31,7 +34,24 @@ def _flush_unlocked(self):
...
+                SSL.WantReadError,
+                SSL.WantWriteError,
+                SSL.WantX509LookupError,
+            ):
+                # these errors require retries with the same data
+                # regardless of whether data has already been written
+                continue
+            except OSError:
+                # This catches errors like EBADF (Bad File Descriptor)
+                # or EPIPE (Broken pipe), which indicate the underlying
```
webknjaz, line 50: These can be made more specific instead of a broad `OSError`.

cheroot/makefile.py
```diff
@@ -31,7 +34,24 @@ def _flush_unlocked(self):
...
+                # we need to remove that from the buffer before retryings
                 n = e.characters_written
+            except (
+                SSL.WantReadError,
+                SSL.WantWriteError,
+                SSL.WantX509LookupError,
+            ):
+                # these errors require retries with the same data
+                # regardless of whether data has already been written
+                continue
```
webknjaz, line 47: How do these leak into the generic writer? Is there an underlying layer where we could catch these and re-raise and common/generic exceptions? Can we scope accessing PyOpenSSL to the `pyopenssl.py` module?

If not, we'll probably have to have a common var in `errors.py` but I don't really like leaking this into outer layers. Need to think of a better structure while we're on it.

cheroot/makefile.py
```diff
@@ -45,9 +65,22 @@ def __init__(self, sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
...
         """Capture bytes read."""
-        val = super().read(*args, **kwargs)
-        self.bytes_read += len(val)
-        return val
+        MAX_ATTEMPTS = 10
+        last_error = None
+        for _ in range(MAX_ATTEMPTS):
+            try:
+                val = super().read(*args, **kwargs)
+            except (SSL.WantReadError, SSL.WantWriteError) as ssl_want_error:
```
webknjaz, line 73: I've basically the same concern with leaking PyOpenSSL into the outer scope here as in https://github.com/cherrypy/cheroot/pull/764/files#r2432889204.

docs/spelling_wordlist.txt
webknjaz, line 1: This shouldn't be needed.

docs/changelog-fragments.d/764.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Added handling for WantWriteError and WantReadError in BufferedWriter
+and StreamReader to enable retries. This addresses long standing issues
+discussed in #245. The reliability of the fix relies on using pyOpenSSL
```
webknjaz, line 3: Use `` :issue:`245` `` for refs. It's not Markdown.

Additionally, you could symlink this change note to that number as well so both will be linked in the change log.

docs/changelog-fragments.d/764.bugfix.rst
webknjaz, line 1: Symlink `245.bugfix.rst` to this file.

docs/changelog-fragments.d/764.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Added handling for WantWriteError and WantReadError in BufferedWriter
+and StreamReader to enable retries. This addresses long standing issues
```
webknjaz, line 2: All these are linkable in Sphinx. pyOpenSSL is plugged via intersphinx. And internal objects are exposed too.

See these to discover the refs:
* https://webknjaz.github.io/intersphinx-untangled/www.pyopenssl.org/
* https://webknjaz.github.io/intersphinx-untangled/cheroot.cherrypy.dev/

hardikmodha:
I recently upgraded from v8.0.0. Previously, I was using v8.0.0 with the patch from https://github.com/cherrypy/cheroot/pull/332 applied.

After upgrading to v11.0.0 and applying the patch from this pull request, I'm seeing `WantReadError` again.

My initial suspicion was the reduced wait time. The earlier patch effectively waited up to ~3 seconds (100 ms intervals), whereas the current implementation retries 10 times with 100 ms intervals, resulting in a total wait time of ~1 second.

However, this does not appear to be the root cause. I tested the same reproducible example against both versions, and the tests that pass on the older version consistently fail on the newer one.

I have attached a minimal reproducible example after extracting the required parts from my application. 

[cheroot_server.py](https://github.com/user-attachments/files/24307187/cheroot_server.py)
[reproduce_issue_245.py](https://github.com/user-attachments/files/24307212/reproduce_issue_245.py)

PR #770, opened 2025-09-22
Fix CI errors and tear down socket errors

julianz-:
This PR deal with numerous errors that arise on submitting even the smallest changes to CI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/770)

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -13,7 +13,7 @@ on:
     - patchback/backports/**  # Patchback always creates PRs
     - pre-commit-ci-update-config  # pre-commit.ci always creates a PR
   pull_request:
-    ignore-paths:  # changes to the cron workflow are triggered through it
+    paths-ignore:  # changes to the cron workflow are triggered through it
     - .github/workflows/scheduled-runs.yml
   workflow_call:  # a way to embed the main tests
   workflow_dispatch:
@@ -83,6 +83,7 @@ env:
     PYTHONIOENCODING
     PYTHONLEGACYWINDOWSSTDIO
     PYTHONUTF8
+    PYTHONWARNINGS: "ignore::ResourceWarning:sqlite3"
   TOX_VERSION: tox < 4.12
   UPSTREAM_REPOSITORY_ID: >-
     16620627
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -286,11 +286,19 @@ class BuiltinSSLAdapter(Adapter):
             raise errors.FatalSSLAlert(
                 *tls_connection_drop_error.args,
             ) from tls_connection_drop_error
-        except ssl.SSLError as generic_tls_error:
-            peer_speaks_plain_http_over_https = (
-                generic_tls_error.errno == ssl.SSL_ERROR_SSL
-                and _assert_ssl_exc_contains(generic_tls_error, 'http request')
-            )
+        except (
+            ssl.SSLError,
+            OSError,
+        ) as generic_tls_error:
+            # When the client speaks plain HTTP into a TLS-only connection,
+            # Python's builtin ssl raises an SSLError with `http request`
+            # in its message. Sometimes, due to a race condition, the socket
+            # is closed by the time we try to handle the error, resulting in an
+            # OSError: [Errno 9] Bad file descriptor.
+            peer_speaks_plain_http_over_https = isinstance(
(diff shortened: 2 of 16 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -13,7 +13,7 @@ on:
     - patchback/backports/**  # Patchback always creates PRs
     - pre-commit-ci-update-config  # pre-commit.ci always creates a PR
   pull_request:
-    ignore-paths:  # changes to the cron workflow are triggered through it
+    paths-ignore:  # changes to the cron workflow are triggered through it
```
webknjaz, line 16: I extracted this into `main`. I have this typo across a bunch of repos, actually. Good catch!

@sirosen any insight into why jsonschema didn't catch this?

sirosen: I need to work up a fix I can submit to schemastore for this, but I did... sort of figure it out.

The [definition of a `pull_request` event](https://github.com/SchemaStore/schemastore/blob/71c836f7a50aa0f796c990c433307dc7b87e300e/src/schemas/json/github-workflow.json#L1348-L1390) mixes explicit properties with a `$ref` usage (the target of the JSON Schema `$ref` is also _named_ `"ref"`, which is a little quirky but fine). When `$ref` loading is _mixed_ with an explicit schema, as in this case, I don't have any clear expectation about what should happen. It might be covered by the spec, but it's definitely going to trip up implementations.

The definition of `"ref"` looks fine at a glance:
https://github.com/SchemaStore/schemastore/blame/71c836f7a50aa0f796c990c433307dc7b87e300e/src/schemas/json/github-workflow.json#L340-L386

But I notice that it doesn't set `"additionalProperties": false`, and I think that's the issue. If I remove that `$ref` usage in a copy of the schema, I get the appropriate error.

It looks like this is the only usage site for `#/definitions/ref`, so I think I'll just inline it and add a test which demonstrates the issue. JSON Schema is always a headtrip! 😵‍💫

sirosen: I did end up getting [a PR](https://github.com/SchemaStore/schemastore/pull/5027) posted!
It ate a lot of my free time for FOSS work this evening, but I think it's worth it. There's a pretty large class of mistakes which `check-jsonschema` can't catch until this is fixed.

webknjaz: @sirosen nice, thanks!

pytest.ini
```diff
@@ -72,6 +74,7 @@ junit_suite_name = cheroot_test_suite
 
 # A mapping of markers to their descriptions allowed in strict mode:
 markers =
+  flaky: mark a test as flaky
```
webknjaz, line 77: This shouldn't be needed as it's coming from a plugin.

tox.ini
```diff
@@ -15,6 +15,7 @@ warnings-to-errors = -Werror
 [testenv]
 deps =
   -rrequirements{/}tests.in
+  pytest-xdist>=3.0.0
```
webknjaz, line 18: This is already pinned to a higher version in a proper place.

tox.ini
```diff
@@ -131,7 +132,6 @@ setenv =
   PYTHONDONTWRITEBYTECODE=x
   WEBTEST_INTERACTIVE=false
 
-
```
webknjaz, line 134: unrelated

cheroot/ssl/builtin.py
```diff
@@ -299,10 +307,6 @@ def wrap(self, sock):
             raise reraised_connection_drop_exc_cls(
                 *generic_tls_error.args,
             ) from generic_tls_error
-        except OSError as tcp_connection_drop_error:
```
webknjaz, line 302: I don't understand why you felt the need to move this into another except-block. That makes it more complicated to follow. And AFAICS the lotig is about the same just more branchy. I was keeping it separate so that it's simple.

cheroot/ssl/builtin.py
```diff
@@ -286,11 +286,19 @@ def wrap(self, sock):
             raise errors.FatalSSLAlert(
                 *tls_connection_drop_error.args,
             ) from tls_connection_drop_error
-        except ssl.SSLError as generic_tls_error:
-            peer_speaks_plain_http_over_https = (
-                generic_tls_error.errno == ssl.SSL_ERROR_SSL
```
webknjaz, line 291: What other error codes can this case have?

PR #774, opened 2025-09-28
Fix stubtest failures in type stub files

julianz-:
Added missing __all__ declarations to server.pyi and threadpool.pyi to match their corresponding Python modules. Removed cheroot.connections.IS_WINDOWS from stubtest allowlist and added to connections.pyi to fix environment-dependent stubtest failures

These changes resolve stubtest validation errors that were preventing CI from passing due to mismatched public API declarations between stub files and runtime modules.


julianz-:
The CI failures are unrelated to the changes in this PR. I will be proposing fixes for the other failures in a separate PR.

docs/changelog-fragments.d/774.contrib.rst
```diff
@@ -0,0 +1 @@
+Resolved various issues causing mypy stubtest errors
```
webknjaz: Could you use full sentences with trailing periods and be more specific here? Vague generic statements aren't very useful in the change log.

docs/spelling_wordlist.txt
```diff
@@ -57,6 +58,7 @@ signalling
 Sigstore
 ssl
 stdout
+stubtest
```
webknjaz: This should probably be wrapped with double backticks as an identifier rather than added to known words

docs/changelog-fragments.d/774.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added missing ``__all__`` and ``IS_WINDOWS`` definitions to ``.pyi`` stub files for mypy.
+-- by :user:`julianz-`
```
webknjaz, line 3: If you want a standalone byline, make it separate form the paragraph so there's no weird `.—` when it's rendered. https://cheroot--774.org.readthedocs.build/en/774/history/#to-be-included-in-the-next-release
```suggestion

-- by :user:`julianz-`
```

PR #777, opened 2025-10-01
Fix: Increase keepalive and HTTP request timeouts to prevent CI failures

julianz-:
These fixes address:

1. The `test_keepalive_conn_management` test was intermittently failing in CI with `http.client.NotConnected`. This was due to the combined execution time of test logic and the explicit `time.sleep()` exceeding the server timeout (2.0s).

2. The http_request_timeout was also too short and causing failures in test_http_over_https_error().

This commit addresses the flakiness by increasing both timeout values.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/777)

### DIFF
--- a/docs/changelog-fragments.d/777.contrib.rst
+++ b/docs/changelog-fragments.d/777.contrib.rst
@@ -1,0 +1,2 @@
+Increased timeout values ``test_client.server_instance.timeout`` and ``http_request_timeout``
+to make related tests more stable.
--- a/cheroot/test/conftest.py
+++ b/cheroot/test/conftest.py
@@ -23,7 +23,7 @@ from ..testing import (  # noqa: F401  # pylint: disable=unused-import
 @pytest.fixture
 def http_request_timeout():
     """Return a common HTTP request timeout for tests with queries."""
-    computed_timeout = 0.1
+    computed_timeout = 0.5
 
     if IS_MACOS:
         computed_timeout *= 2
--- a/cheroot/test/test_conn.py
+++ b/cheroot/test/test_conn.py
@@ -504,7 +504,7 @@ def test_keepalive(test_client, http_server_protocol):
 
 def test_keepalive_conn_management(test_client):
     """Test management of Keep-Alive connections."""
-    test_client.server_instance.timeout = 2
+    test_client.server_instance.timeout = 3
 
     def connection():
         # Initialize a persistent HTTP connection


docs/changelog-fragments.d/777.contrib.rst
webknjaz, line 1: This is not a user-facing fix. The category should be `contrib`.

Also, the note does not render well. You can see the preview @ https://cheroot--777.org.readthedocs.build/en/777/history/#to-be-included-in-the-next-release.

Remember to use RST markup for things you mention/reference.

docs/spelling_wordlist.txt
webknjaz, line 1: I bet if you use double proper RST around identifiers, this hack wouldn't be necessary.

docs/changelog-fragments.d/777.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+The changes in timeout values address two issues:
+1. The test_keepalive_conn_management test was intermittently
+failing in CI with http.client.NotConnected.
+This was due to the combined execution time of test logic and
+the explicit time.sleep() exceeding the server timeout (2.0s).
+
+2. The http_request_timeout was also too short and causing
+failures in test_http_over_https_error().
```
webknjaz: Here's a few notes on writing change logs:
https://cheroot--777.org.readthedocs.build/en/777/contributing/guidelines/#adding-change-notes-with-your-prs / https://pip-tools.rtfd.io/en/stable/contributing/#adding-change-notes-with-prs.

TL;DR consider the audience of the change log. For regular features/bugfixes it's the end-users, for contrib/packaging things it's somebody a little more involved. Change notes don't need to duplicate what's in Git commits because that's for the devs and whoever does Git archaeology and stuff. Change logs also have a different structure/style. For Git commits, it's imperative mood in the title and some motivation in the long description. For change logs, it's something like the past tense or non-imperative present simple. Git commits are scoped to small patches, while change logs represent cumulative changes since the previous release and what the high-level effect it has on the readers. Change logs are more like news, they give more overview and hints regarding what the end-user will get/need to do upon upgrade, not get into implementation details (that's what Git is for).

This note probably needs to just mention that specific tests are made more stable with higher timeouts. Or something like it. Maybe add a byline.

webknjaz:
@julianz- could you also link the example failures you were seeing in actual CI logs?

docs/changelog-fragments.d/777.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Increased timeout values `test_keepalive_conn_management` and `http_request_timeout`
```
webknjaz: The first one *is* a test, though. Also, RST uses double backticks.

webknjaz: Or did you mean to reference private API pages?

julianz-: I revised.

julianz-:
> @julianz- could you also link the example failures you were seeing in actual CI logs?

Here's a link for the failure with test_keepalive_conn_management().
https://github.com/cherrypy/cheroot/actions/runs/18145322195/job/51645700584?pr=764 for test_keepalive_conn_management() failure

Unfortunately, I can't find a log for the other timeout failure now. I saw it many times but it's bit of a nightmare rummaging through all these builds. I can drop this change and see it if comes back again.

webknjaz (approved):
Alright, I've skimmed through the second test and even though there's no log to check, the timeout change doesn't seem harmful so I'll just go with it. Thanks!

PR #778, opened 2025-10-02
Daemonize the serve_unservicable thread

### LINKED ISSUE #769: Should the new _serve_unservicable thread be daemon?
❓ **I'm submitting a ...**
- [ ] 🐞 bug report
- [ ] 🐣 feature request
- [X] ❓ question about the decisions made in the repository

The most recent change in 11.0.0 that added a non-daemon thread makes the library harder to use through pytest.  Specifically, I have an existing project that uses pytest to create a server in a fixture, and this thread is not closing at the end of pytest.  I can hack around it by disabling the new feature (`cheroot.server.HTTPServer._serve_unservicable = lambda: None`), but this definitely feels wrong.

I think that if the `_serve_unservicable` thread was daemon it would make using this in tests easier.  The side effect is that when shutting down a server it wouldn't drain the 503 responses; but prior to 11.x those didn't exist.

💡 **To Reproduce**

I assume in my tests where I start and stop a server, some request is getting queued after I ask the server to stop and this test, rather than that being ignored is stopping the process from stopping.

💡 **Expected behavior**
I'd like to be able to start and stop servers in pytest fixtures without worrying about exact timing from client tests.

📋 **Environment**
- Cheroot version: 11.0
- CherryPy version: 18.10.0
- Python version: 3.9 through 3.13
- OS: Ubuntu

manthey:
In Python < 3.13, the code backfills a shutdown method for a queue, but this isn't complete; an existing get call is not interrupted by shutdown and this means that when the unservicable queue is empty, it never stops, preventing the system from stopping.  By daemonizing the unservicable thread, the server can stop and exit.

A better solution would probably involve copying a fair amount of code from Python 3.13's queue implementation.  In my opinion that isn't worth the maintenance burden.  If desired, perhaps we should only daemonize this thread on Python < 3.13.

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves #769

It is impossible to stop a server that has no pending unservicable requests on Python < 3.13.

You can stop a server that has no pending unservicable requests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/778)

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1929,7 +1929,7 @@ class HTTPServer:
         # This thread will handle unservicable connections, as added to
         # self._unservicable_conns queue. It will run forever, until
         # self.stop() tells it to shut down.
-        threading.Thread(target=self._serve_unservicable).start()
+        threading.Thread(target=self._serve_unservicable, daemon=True).start()
 
         while self.ready and not self.interrupt:
             try:


itamarst:
I will try to look at it on Monday, if I don't do it then, remind me again.

webknjaz:
@itamarst hey, have you had a chance to lurk in?

webknjaz:
> This seems fine, it's supposed to be a no side-effect add on.

@itamarst what about the comment above? Looks like the thread remains lingering in some contexts (including maybe https://github.com/cherrypy/cherrypy/issues/2070).

@martin-vi you mentioned Python 3.12 — is that the only version you tested?

itamarst:
OK I'll dig in to this Wednesday (tomorrow I'm doing local election volunteering).

webknjaz:
🎉 v11.1.2 just hit PyPI: https://pypi.org/project/cheroot/11.1.2/
* https://github.com/cherrypy/cheroot/releases/tag/v11.1.2
* https://github.com/cherrypy/cheroot/discussions/796

This should be fixed via https://github.com/cherrypy/cheroot/pull/794 now.

Issue #734, opened 2024-11-05
random failures when running test_conn with Python 3.13


kulikjak:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**
And one more failure with Python 3.13: when running `cheroot/test/test_conn.py`, I am seeing one of the tests fail with the following error:
```python-traceback
E                   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <function IOBase.__del__ at 0x1ffd0ae81cea20>
E                   
E                   Traceback (most recent call last):
E                     File "/usr/lib/python3.13/_pyio.py", line 418, in __del__
E                       self.close()
E                       ~~~~~~~~~~^^
E                     File "/usr/lib/python3.13/_pyio.py", line 1313, in close
E                       self.flush()
E                       ~~~~~~~~~~^^
E                     File "/usr/lib/python3.13/_pyio.py", line 1274, in flush
E                       self._flush_unlocked()
E                       ~~~~~~~~~~~~~~~~~~~~^^
E                     File "/scratch/userland-gate/components/python/cheroot/cheroot-10.0.0/cheroot/makefile.py", line 33, in _flush_unlocked
E                       n = self.raw.write(bytes(self._write_buf))
E                     File "/usr/lib/python3.13/socket.py", line 737, in write
E                       return self._sock.send(b)
E                              ~~~~~~~~~~~~~~~^^^
E                   OSError: [Errno 9] Bad file number
```
I saw the following three fail: `test_streaming_10[True]`, `test_keepalive[HTTP/1.0]`, `test_keepalive[HTTP/1.1]` with this error, although it's always just one for each run (and it seems random).

This again seems to be related to the following io change:
https://docs.python.org/3/whatsnew/3.13.html#io

I wasn't able to find a fix for this one.

💡 **To Reproduce**
Run the test suite with Python 3.13.0.

📋 **Details**
Full traceback:
```python-traceback
==================================== ERRORS ====================================
__________________ ERROR at setup of test_keepalive[HTTP/1.0] __________________

cls = <class '_pytest.runner.CallInfo'>
func = <function call_and_report.<locals>.<lambda> at 0x1fffe71a7c2020>
when = 'setup'
reraise = (<class '_pytest.outcomes.Exit'>, <class 'KeyboardInterrupt'>)

    @classmethod
    def from_call(
[...]

PR #779, opened 2025-10-05
Fix race conditions and improve robustness during socket I/O

julianz-:
Fixed race conditions to make socket I/O more resilient during connection teardown.

1. BufferedWriter's ``write()``: Added error handling to ignore common socket errors (e.g., ``ECONNRESET``, ``EPIPE``, ``ENOTCONN``, ``EBADF``) that occur when the underlying connection has been unexpectedly closed by the client or OS. This prevents a crash when attempting to write to a defunct socket.
2. BufferedWriters's ``close()``: Made idempotent, allowing safe repeated calls without raising exceptions.
3. Needed to add explicit handling of WINDOWS environments as these are seen to throw Windows specific ``WSAENOTSOCK`` errors.

Includes new unit tests to cover the idempotency and graceful handling of already closed underlying buffers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/779)


cheroot/makefile.py
```diff
@@ -2,8 +2,30 @@
...
 import socket
+import sys
 
+from OpenSSL.SSL import SysCallError
+
+
+# Define a variable to hold the platform-specific "not a socket" error.
+_not_a_socket_err = None
+
+if sys.platform == 'win32':
```
webknjaz: Import this from the `_compat` module.

cheroot/makefile.py
```diff
@@ -30,10 +52,59 @@ def _flush_unlocked(self):
...
+                # Check for a different error attribute depending
+                # on the exception type
+                if isinstance(e, io.BlockingIOError):
+                    n = e.characters_written
+                else:
+                    error_code = (
+                        e.errno if isinstance(e, OSError) else e.args[0]
+                    )
+                    if error_code in {
+                        errno.EBADF,
```
webknjaz: Can this happen on Windows?

cheroot/errors.py
```diff
@@ -66,6 +72,7 @@ def plat_specific_errors(*errnames):
 
 
 acceptable_sock_shutdown_error_codes = {
+    errno.EBADF,
```
webknjaz: Can we have something like

```suggestion
_bad_fd_errors = {errno.EBADF}
if IS_WINDOWS:
    _bad_fd_errors(socket.WSAENOTSOCK)

acceptable_sock_shutdown_error_codes = {
    *_bad_fd_errors,
```

? Perhaps, with that fallback you're trying to add to `makefile`. We could just have that here.

cheroot/errors.py
```diff
@@ -87,4 +94,8 @@ def plat_specific_errors(*errnames):
 * https://docs.microsoft.com/windows/win32/api/winsock/nf-winsock-shutdown
 """
 
-acceptable_sock_shutdown_exceptions = (BrokenPipeError, ConnectionResetError)
+acceptable_sock_shutdown_exceptions = (
+    BrokenPipeError,
+    ConnectionResetError,
+    _OpenSSL_SysCallError,
```
webknjaz: Perhaps avoid adding `None` here and make it conditional?
```suggestion
    *(() if _OpenSSL_SysCallError is None else (_OpenSSL_SysCallError,)),
```

cheroot/makefile.py
```diff
@@ -30,10 +52,59 @@ def _flush_unlocked(self):
                 # ssl sockets only except 'bytes', not bytearrays
                 # so perhaps we should conditionally wrap this for perf?
                 n = self.raw.write(bytes(self._write_buf))
-            except io.BlockingIOError as e:
-                n = e.characters_written
+            except (io.BlockingIOError, OSError, SysCallError) as e:
+                # Check for a different error attribute depending
+                # on the exception type
+                if isinstance(e, io.BlockingIOError):
+                    n = e.characters_written
```
webknjaz: Keep it in a separate `except` block to avoid unnecessary nested blocks. It'll be easier to reason about.

cheroot/makefile.py
```diff
@@ -30,10 +52,59 @@ def _flush_unlocked(self):
...
+        A SysCallError or OSError with errno.EBADF or errno.ENOTCONN is caught
+        and ignored, as these indicate a normal, expected connection teardown.
+        Other exceptions are re-raised.
+        """
+        if self.closed:  # pylint: disable=W0125
+            return
+
+        try:
+            super().close()
+        except (OSError, SysCallError) as e:
```
webknjaz: Should this be in a var?

cheroot/server.py
```diff
@@ -1541,9 +1577,10 @@ def _close_kernel_socket(self):
             self.socket.shutdown,
         )
 
+        acceptable_exceptions = errors.acceptable_sock_shutdown_exceptions
         try:
             shutdown(socket.SHUT_RDWR)  # actually send a TCP FIN
-        except errors.acceptable_sock_shutdown_exceptions:
+        except acceptable_exceptions:  # pylint: disable=E0712
```
webknjaz: why?

cheroot/test/test_conn.py
```diff
@@ -1637,3 +1638,33 @@ def test_invalid_selected_connection(test_client, monkeypatch):
...
     assert faux_select.os_error_triggered
     assert faux_get_map.conn_closed
+
+
+class TestBufferedWriter:
+    """Test cases for BufferedWriter close() method race condition handling."""
+
+    def test_close_is_idempotent(self):
+        """Test that close() can be called multiple times safely."""
+        import io
```
webknjaz: Put imports at the top.

cheroot/test/test_conn.py
```diff
@@ -1637,3 +1638,33 @@ def test_invalid_selected_connection(test_client, monkeypatch):
     time.sleep(test_client.server_instance.expiration_interval * 2)
     assert faux_select.os_error_triggered
     assert faux_get_map.conn_closed
+
+
+class TestBufferedWriter:
```
webknjaz: Drop the class. This should go into `test_makefile.py`.

docs/changelog-fragments.d/779.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed race conditions to make socket I/O more resilient during connection teardown.
+-- by :user:`julianz-`
```
webknjaz, line 4:
```suggestion

-- by :user:`julianz-`
```

docs/spelling_wordlist.txt
webknjaz, line 1: this looks unnecessary

cheroot/server.py
```diff
@@ -1186,12 +1189,25 @@ def ensure_headers_sent(self):
...
+                self.conn.wfile.write(EMPTY.join(buf))
+            else:
+                self.conn.wfile.write(chunk)
+        except (SysCallError, ConnectionError, OSError) as e:
+            error_code = e.errno if isinstance(e, OSError) else e.args[0]
+            if error_code in {
+                errno.ECONNRESET,
+                errno.EPIPE,
+                errno.ENOTCONN,
+                errno.EBADF,
```
webknjaz: We should probably get this from a var.

cheroot/server.py
```diff
@@ -1186,12 +1189,25 @@ def ensure_headers_sent(self):
...
-            self.conn.wfile.write(chunk)
+        try:
+            if self.chunked_write and chunk:
+                chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
+                buf = [chunk_size_hex, CRLF, chunk, CRLF]
+                self.conn.wfile.write(EMPTY.join(buf))
+            else:
+                self.conn.wfile.write(chunk)
+        except (SysCallError, ConnectionError, OSError) as e:
+            error_code = e.errno if isinstance(e, OSError) else e.args[0]
```
webknjaz: There's a number of places where this is repeated. It's worth having a helper function, I think.

cheroot/server.py
```diff
@@ -1186,12 +1189,25 @@ def ensure_headers_sent(self):
...
-        else:
-            self.conn.wfile.write(chunk)
+        try:
+            if self.chunked_write and chunk:
+                chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
+                buf = [chunk_size_hex, CRLF, chunk, CRLF]
+                self.conn.wfile.write(EMPTY.join(buf))
+            else:
+                self.conn.wfile.write(chunk)
+        except (SysCallError, ConnectionError, OSError) as e:
```
webknjaz: Plz don't use single-chars for vars. Add some context through naming.

cheroot/server.py
```diff
@@ -1186,12 +1189,25 @@ def ensure_headers_sent(self):
...
-            self.conn.wfile.write(EMPTY.join(buf))
-        else:
-            self.conn.wfile.write(chunk)
+        try:
+            if self.chunked_write and chunk:
+                chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
+                buf = [chunk_size_hex, CRLF, chunk, CRLF]
+                self.conn.wfile.write(EMPTY.join(buf))
+            else:
+                self.conn.wfile.write(chunk)
```
webknjaz: Avoid having more than a single instruction in try blocks. Having more will let exceptions unexpectedly originate from unexpected places in the future.

cheroot/makefile.py
```diff
@@ -30,10 +52,59 @@ def _flush_unlocked(self):
                 # ssl sockets only except 'bytes', not bytearrays
                 # so perhaps we should conditionally wrap this for perf?
                 n = self.raw.write(bytes(self._write_buf))
-            except io.BlockingIOError as e:
-                n = e.characters_written
+            except (io.BlockingIOError, OSError, SysCallError) as e:
```
webknjaz: Looks like you could make use of something like `acceptable_sock_shutdown_exceptions` here since corresponding specific exceptions are being raised: https://github.com/cherrypy/cheroot/issues/734#issuecomment-2994479816.

cheroot/makefile.py
julianz-: Apparently so - the revised version should handle this.

cheroot/errors.py
julianz-: have revised

cheroot/errors.py
julianz-: nice!

cheroot/makefile.py
julianz-: Agreed!

cheroot/makefile.py
julianz-: i can reuse acceptable_sock_shutdown_error_codes?

cheroot/server.py
julianz-: revised

cheroot/server.py
julianz-: Revised. Yes that was silly - I think this was the aftermath of refactoring.

cheroot/errors.py
```diff
@@ -64,13 +72,31 @@ def plat_specific_errors(*errnames):
...
 }
+
+if IS_WINDOWS:
+    # Define Windows socket error code constant
+    # WSAENOTSOCK: Socket operation on non-socket
+    # Should be available from errno, but provide
+    # a value as backup
+    WSAENOTSOCK = 10038
+    acceptable_sock_shutdown_error_codes.add(
+        getattr(errno, 'WSAENOTSOCK', WSAENOTSOCK),
```
webknjaz: What are the cases when this constant isn't available on windows?

cheroot/errors.py
```diff
@@ -86,5 +112,3 @@ def plat_specific_errors(*errnames):
 * https://github.com/python/cpython/blob/c39b52f/Lib/poplib.py#L297-L302
 * https://docs.microsoft.com/windows/win32/api/winsock/nf-winsock-shutdown
 """
-
-acceptable_sock_shutdown_exceptions = (BrokenPipeError, ConnectionResetError)
```
webknjaz, line 90: Moving this looses a docstring defined a few lines below.

cheroot/errors.pyi
```diff
@@ -1,5 +1,7 @@
 from typing import List, Set, Tuple, Type
 
+from cheroot._compat import IS_WINDOWS as IS_WINDOWS
```
webknjaz: Can we import `_compat` in there instead so that it remains private?

cheroot/errors.py
```diff
@@ -3,6 +3,14 @@
 import errno
 import sys
 
+from cheroot._compat import IS_WINDOWS
```
webknjaz: Let's just have a private var. Plus for normal runtime code, there's no need to use absolute imports.
```suggestion
from . import _compat
```

cheroot/makefile.py
```diff
@@ -32,8 +36,57 @@ def _flush_unlocked(self):
...
+                error_code = e.args[0]
+                if error_code in errors.acceptable_sock_shutdown_error_codes:
+                    # The socket is gone, so just ignore this error.
+                    return
+                raise
+            else:
+                # The 'try' block completed without an exception
+                if n is None:
+                    # This could happen with non-blocking write
+                    # when nothing was written
```
webknjaz: Wouldn't it be `0` in that case?

cheroot/errors.py
julianz-: I thought I saw an error one time when it wasn't defined but I just tried removing and everything passed so will leave this out.

cheroot/errors.py
julianz-: moved things back again

cheroot/errors.py
julianz-: i get linter error: 

cheroot/errors.py:7:1: WPS300 Found local folder import
from ._compat import IS_WINDOWS

if I use relative import?

webknjaz:
Oh, it seems there's some more errors showing up on `main` after `pre-commit.ci`'s cache got invalidated. I'll see what needs to be done.

cheroot/errors.py
webknjaz: I'll add a global import for that rule. But plz use `import` w/o `from` for this one.

.github/workflows/ci-cd.yml
```diff
@@ -632,10 +629,7 @@ jobs:
       post-toxenv-preparation-command: >-
         ${{
           matrix.toxenv == 'pre-commit'
-          && format(
-            '$(pwd)/.tox/{0}/bin/python -Im pre_commit install-hooks',
-            matrix.toxenv
-          )
+          && 'python -m pip install --upgrade pip setuptools pre-commit'
```
webknjaz: revert this

cheroot/test/test_makefile.py
```diff
@@ -51,3 +53,27 @@ def test_bytes_written():
...
+    assert buffered_writer.closed
+
+
+def test_close_handles_already_closed_buffer():
+    """Test that close() handles already closed underlying buffer."""
+    raw_buffer = io.BytesIO()
+    buffered_writer = makefile.BufferedWriter(raw_buffer)
+
+    # Close the underlying buffer first
+    raw_buffer.close()
```
webknjaz, line 108:
```suggestion
    raw_buffer.close()

    assert raw_writer.closed and not buffered_writer.closed
```

cheroot/test/test_makefile.py
```diff
@@ -51,3 +53,27 @@ def test_bytes_written():
...
+
+    # This should not raise an exception
+    buffered_writer.close()
```
webknjaz, line 111:
```suggestion
    buffered_writer.close()

    assert buffered_writer.closed
```

cheroot/test/test_makefile.py
```diff
@@ -51,3 +53,27 @@ def test_bytes_written():
...
+def test_close_is_idempotent():
+    """Test that close() can be called multiple times safely."""
+    raw_buffer = io.BytesIO()
+    buffered_writer = makefile.BufferedWriter(raw_buffer)
+
+    # Should not raise any exceptions
+    buffered_writer.close()
+    buffered_writer.close()  # Second call should be safe
+
+    assert buffered_writer.closed
```
webknjaz, line 96: This will make it explicit that the second call is performed on an already-closed buffer.
```suggestion
    buffered_writer.close()

    assert buffered_writer.closed

    buffered_writer.close()  # Second call should be safe
```

cheroot/test/test_makefile.py
```diff
@@ -51,3 +53,27 @@ def test_bytes_written():
...
+
+
+def test_close_handles_already_closed_buffer():
+    """Test that close() handles already closed underlying buffer."""
```
webknjaz:
```suggestion
    """Test that ``close()`` handles already closed underlying buffer."""
```

cheroot/test/test_makefile.py
```diff
@@ -51,3 +53,27 @@ def test_bytes_written():
     wfile = makefile.MakeFile(sock, 'w')
     wfile.write(b'bar')
     assert wfile.bytes_written == 3
+
+
+def test_close_is_idempotent():
+    """Test that close() can be called multiple times safely."""
```
webknjaz:
```suggestion
    """Test that double ``close()`` does not error out."""
```

cheroot/test/test_makefile.py
```diff
@@ -1,10 +1,12 @@
 """Tests for :py:mod:`cheroot.makefile`."""
 
+import io
+
 from cheroot import makefile
 
 
 class MockSocket:
-    """A mock socket."""
+    """A mock socket for emulating buffered I/O."""
```
webknjaz, line 13: Let's extract this into a separate PR.

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
     socket_errors_to_ignore.extend(plat_specific_errors('EPROTOTYPE'))
     socket_errors_nonblocking.extend(plat_specific_errors('EPROTOTYPE'))
 
+sock_shutdown_errs = (OSError, _OpenSSL_SysCallError)
+write_shutdown_errs = (OSError, _OpenSSL_SysCallError, ConnectionError)
```
webknjaz: I don't like that these tuples can contain `None`. Only add the exception if it's importable. Additionally, you can use the first var to populate the second one.

```suggestion
sock_shutdown_errs = (
    OSError,
    *(() if _OpenSSL_SysCallError is None else (_OpenSSL_SysCallError,)),
)
write_shutdown_errs = (*sock_shutdown_errs, ConnectionError)
```

Also, could you explain why `ConnectionError` can't happen on shutdown?

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
...
 }
-"""Errors that may happen during the connection close sequence.
 
+if IS_WINDOWS:
+    # Add Windows socket error code constant
+    # WSAENOTSOCK: Socket operation on non-socket
+    # Should be available from errno
+    acceptable_sock_shutdown_error_codes.add(
+        errno.WSAENOTSOCK,
+    )
```
webknjaz: The docstring should remain on the line following the variable declaration. It shouldn't be detached by a block of code inserted in-between.

This actually breaks the doc inclusion. Compare `main`
(https://cheroot.cherrypy.dev/en/latest/pkg/cheroot.errors/#cheroot.errors.acceptable_sock_shutdown_error_codes) and this PR preview (https://cheroot--779.org.readthedocs.build/en/779/pkg/cheroot.errors/#cheroot.errors.acceptable_sock_shutdown_error_codes) to see how it affects the issue.

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
...
+    errno.EBADF,
     errno.ENOTCONN,
     errno.EPIPE,
     errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
     errno.ECONNRESET,  # corresponds to ConnectionResetError in Python 3
 }
-"""Errors that may happen during the connection close sequence.
 
+if IS_WINDOWS:
+    # Add Windows socket error code constant
```
webknjaz: This phrase just repeats what's already in code. Code comments are supposed to explain motivation, not duplicate what's right below.

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
...
+    # WSAENOTSOCK: Socket operation on non-socket
```
webknjaz: This could go into a docstring.

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
...
+    # Should be available from errno
+    acceptable_sock_shutdown_error_codes.add(
+        errno.WSAENOTSOCK,
+    )
+
+"""Errors that may happen during the connection close sequence.
```
webknjaz, line 80: Make sure to keep an empty line separator between the docstring summary line and the paragraph.

cheroot/errors.py
```diff
@@ -64,15 +72,27 @@ def plat_specific_errors(*errnames):
     socket_errors_to_ignore.extend(plat_specific_errors('EPROTOTYPE'))
     socket_errors_nonblocking.extend(plat_specific_errors('EPROTOTYPE'))
 
+sock_shutdown_errs = (OSError, _OpenSSL_SysCallError)
+write_shutdown_errs = (OSError, _OpenSSL_SysCallError, ConnectionError)
 
 acceptable_sock_shutdown_error_codes = {
+    errno.EBADF,
```
webknjaz: We can just inline this so it's not detached in some separate place in the module.
```suggestion
    errno.EBADF,
    *((errno.WSAENOTSOCK,) if IS_WINDOWS else ()),
```

cheroot/errors.py
```diff
@@ -87,4 +107,26 @@ def plat_specific_errors(*errnames):
 * https://docs.microsoft.com/windows/win32/api/winsock/nf-winsock-shutdown
 """
 
-acceptable_sock_shutdown_exceptions = (BrokenPipeError, ConnectionResetError)
+acceptable_sock_shutdown_exceptions = (
+    BrokenPipeError,
+    ConnectionResetError,
+    # conditionally add _OpenSSL_SysCallError to the list
```
webknjaz: This just repeats what's on the next line exactly. Such comments are noise and shouldn't be included. But you could explain motivation/impact instead and it would be useful.

cheroot/errors.py
```diff
@@ -3,6 +3,14 @@
 import errno
 import sys
 
+from cheroot._compat import IS_WINDOWS
+
+
+try:
+    from OpenSSL.SSL import SysCallError as _OpenSSL_SysCallError
+except ImportError:
+    _OpenSSL_SysCallError = None
```
webknjaz: How about we declare the iterable here and just unpack it in those vars below?
```suggestion
    from OpenSSL.SSL import SysCallError as _OpenSSL_SysCallError
except ImportError:
    _pyopenssl_syscall_errors = ()
else:
    _pyopenssl_syscall_errors = (_OpenSSL_SysCallError,)
```

cheroot/errors.py
webknjaz:
```suggestion
sock_shutdown_errs = (OSError, *_pyopenssl_syscall_errors)
write_shutdown_errs = (*sock_shutdown_errs, ConnectionError)
```

cheroot/makefile.py
```diff
@@ -4,6 +4,8 @@
 import _pyio as io
 import socket
 
+from cheroot import errors
```
webknjaz: Let's do

```suggestion
from . import errors as _errors
```

to avoid exposing another public name in this module.

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
                 n = self.raw.write(bytes(self._write_buf))
             except io.BlockingIOError as e:
                 n = e.characters_written
+            except errors.sock_shutdown_errs as sock_err:  # pylint: disable=catching-non-exception
```
webknjaz: Perhaps, the linting suppression won't be necessary when you drop `None` from that var.
```suggestion
            except _errors.sock_shutdown_errs as sock_err:
```

cheroot/errors.py
webknjaz: And reference + mention explanation from https://learn.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2

cheroot/errors.py
webknjaz: Does this attempt to duplicate `acceptable_sock_shutdown_exceptions`? I think that `OSError` shouldn't be here since it's a broad exception and `ConnectionError` is its subclass anyway.
As these vars are defined now, they are basically the same.

cheroot/test/test_makefile.py
julianz-: asserting ``not buffered_writer.closed`` fails. As soon as I close the ``raw_buffer`` then ``buffered_writer`` becomes closed.  i think your following comment corrected this.

cheroot/test/test_makefile.py
julianz-: Not sure I understand. The only thing modified here was the docstring. This was to distinguish ``MockSocket`` which was already there from ``MockRawSocket`` which I am adding. Originally, I added the latter in the same file but had to move to test_errors.py because the linter said I had too many modules in test_makefile.py.

cheroot/errors.py
julianz-: I agree some of this was messed up. Have revised and reverted closer to the original.

cheroot/errors.py
julianz-: this has been absorbed now

cheroot/errors.py
julianz-: I changed to using just import but now get:

```
6:1      WPS301 Found dotted raw import: cheroot._compat
           -> https://pyflak.es/WPS301
  import cheroot._compat
  ^

WPS301: Found dotted raw import: cheroot._compat
  1     cheroot/errors.py
Total: 1
```

julianz-:
To get anywhere with the CI, I had to remove darglint ~= 1.8.1 from flake in .pre-commit-config.yaml because it is apparently very buggy and no longer supported. I could not get past errors it was throwing. I also had to fix docs/conf.py adding missing commas to the sphinx-doc list. I can remove these changes as they are unrelated to the PR.

.pre-commit-config.yaml
webknjaz, line 1: This shouldn't be necessary if you're rebased on the latest `main`.

cheroot/errors.py
```diff
@@ -3,6 +3,16 @@
 import errno
 import sys
 
+import cheroot._compat
```
webknjaz:
```suggestion
from . import _compat
```

cheroot/errors.py
```diff
@@ -66,13 +76,15 @@ def plat_specific_errors(*errnames):
 
 
 acceptable_sock_shutdown_error_codes = {
+    errno.EBADF,
     errno.ENOTCONN,
     errno.EPIPE,
     errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
     errno.ECONNRESET,  # corresponds to ConnectionResetError in Python 3
+    *((errno.WSAENOTSOCK,) if cheroot._compat.IS_WINDOWS else ()),
```
webknjaz:
```suggestion
    *((errno.WSAENOTSOCK,) if _compat.IS_WINDOWS else ()),
```

cheroot/errors.py
```diff
@@ -87,4 +99,26 @@ def plat_specific_errors(*errnames):
...
+    # Assuming SysCallError has the code at index 0
+    elif isinstance(sock_err, _pyopenssl_syscall_errors):
+        error_code = sock_err.args[0]
+
+    return error_code in acceptable_sock_shutdown_error_codes
+
+
+acceptable_sock_shutdown_exceptions = (
+    BrokenPipeError,  # Covers EPIPE and ESHUTDOWN
+    ConnectionResetError,  # Covers ECONNRESET)
```
webknjaz:
```suggestion
    ConnectionResetError,  # Covers ECONNRESET
```

cheroot/errors.py
```diff
@@ -66,13 +76,15 @@ def plat_specific_errors(*errnames):
...
 
 acceptable_sock_shutdown_error_codes = {
+    errno.EBADF,
     errno.ENOTCONN,
     errno.EPIPE,
     errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
     errno.ECONNRESET,  # corresponds to ConnectionResetError in Python 3
+    *((errno.WSAENOTSOCK,) if cheroot._compat.IS_WINDOWS else ()),
 }
 """Errors that may happen during the connection close sequence.
```
webknjaz, line 80: You forgot this:
```suggestion
"""Errors that may happen during the connection close sequence.

```
*(https://github.com/cherrypy/cheroot/pull/779/files#r2432793008)*

webknjaz: Missing it makes the render broken: https://cheroot--779.org.readthedocs.build/en/779/pkg/cheroot.errors/#cheroot.errors.acceptable_sock_shutdown_error_codes

webknjaz: While on it, I think `Refs:` below also needs an empty line before the list.

webknjaz:
@julianz- 

> To get anywhere with the CI, I had to remove darglint ~= 1.8.1 from flake in .pre-commit-config.yaml because it is apparently very buggy and no longer supported. I could not get past errors it was throwing. I also had to fix docs/conf.py adding missing commas to the sphinx-doc list. I can remove these changes as they are unrelated to the PR.

pre-commit is passing on `main`. You might've been rebased on a temporarily broken state at some point but it shouldn't be the case anymore. So changes outside of your patch shouldn't be needed.

We can't replace `darglint` with Ruff because there's no replacement in it (https://github.com/astral-sh/ruff/issues/458). However, I thought it'd be a good idea to try migrating to https://pypi.org/p/pydoclint. Feel free to look into it separately.

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
+        This method is designed to be idempotent (it can be called multiple
+        times without side effects). It gracefully handles a race condition
+        where the underlying socket may have already been closed by the remote
+        client or another thread.
+
+        A SysCallError or OSError with ``EBADF`` or ``ENOTCONN`` is caught
+        and ignored, as these indicate a normal, expected connection teardown.
+        Other exceptions are re-raised.
+        """
+        if self.closed:  # pylint: disable=W0125
```
webknjaz: Use human-readable codes for pylint. A letter and a bunch of numbers don't really give away any context regarding when this can be dropped or why it's here is the first place.

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
+    def close(self):
+        """
+        Close the stream and its underlying file object.
+
+        This method is designed to be idempotent (it can be called multiple
+        times without side effects). It gracefully handles a race condition
+        where the underlying socket may have already been closed by the remote
+        client or another thread.
+
+        A SysCallError or OSError with ``EBADF`` or ``ENOTCONN`` is caught
```
webknjaz: All these are referencible
```suggestion
        A :exc:`~OpenSSL.SSL.SysCallError` or :exc:`OSError` with
        :data:`~errno.EBADF` or :data:`~errno.ENOTCONN` is caught
```

cheroot/server.py
```diff
@@ -1192,9 +1192,21 @@ def write(self, chunk):
...
-            self.conn.wfile.write(chunk)
+            data = chunk
+
+        try:
+            self.conn.wfile.write(data)
+        except errors.acceptable_sock_shutdown_exceptions as write_error:
+            if isinstance(write_error, OSError):
+                error_code = write_error.errno
+            else:
+                error_code = write_error.args[0]
```
webknjaz: Pretty sure this branch is unreachable.

webknjaz: see https://app.codecov.io/gh/cherrypy/cheroot/commit/85bedd19aa56533a58272a4786fb64cbe0069c5b/blob/cheroot/server.py?dropdown=coverage#L1205

cheroot/server.py
```diff
@@ -1192,9 +1192,21 @@ def write(self, chunk):
...
+            if error_code in errors.acceptable_sock_shutdown_error_codes:
+                # The socket is gone, so just ignore this error.
+                return
+            raise
```
webknjaz: Looks like this is equivalent to
```suggestion
        except errors.acceptable_sock_shutdown_exceptions:
            # The socket is gone, so just ignore this error.
            return
```

cheroot/errors.py
```diff
@@ -87,4 +99,26 @@ def plat_specific_errors(*errnames):
 * https://docs.microsoft.com/windows/win32/api/winsock/nf-winsock-shutdown
 """
 
-acceptable_sock_shutdown_exceptions = (BrokenPipeError, ConnectionResetError)
+
+def _is_acceptable_socket_shutdown_error(sock_err):
```
webknjaz: I'd rather have this helper only handling non-specific exceptions and leave the narrowed down ones in explicit except blocks.

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
                 n = e.characters_written
+            except OSError as sock_err:
+                error_code = sock_err.errno
+                if error_code in _errors.acceptable_sock_shutdown_error_codes:
+                    # The socket is gone, so just ignore this error.
+                    return
+                raise
+            else:
+                # The 'try' block completed without an exception
+                if n == 0:
```
webknjaz: So https://docs.python.org/3/library/io.html#io.RawIOBase.write says

> `None` is returned if the raw stream is set not to block and no single byte could be readily written to it. The caller may release or mutate *b* after this method returns, so the implementation should only access *b* during the method call.

Now should this be handled? With a retry? Also, is the difference between `0` and `None` significant?

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
+        A SysCallError or OSError with ``EBADF`` or ``ENOTCONN`` is caught
+        and ignored, as these indicate a normal, expected connection teardown.
+        Other exceptions are re-raised.
+        """
+        if self.closed:  # pylint: disable=W0125
+            return
+
+        try:
+            super().close()
+        except OSError as sock_err:
```
webknjaz: I'd rather have explicit exceptions in explicit `except` blocks with the generic exception being separate.
```suggestion
        except _errors.acceptable_sock_shutdown_exceptions:
            return
        except OSError as sock_err:
```

cheroot/errors.py
```diff
@@ -3,6 +3,16 @@
 import errno
 import sys
 
+import cheroot._compat
+
+
+try:
+    from OpenSSL.SSL import SysCallError as _OpenSSL_SysCallError
```
webknjaz: Is it possible to handle this in `cheroot.ssl.pyopenssl` and re-raise as `ConnectionError` or similar so that the external code would only deal with the unified exception and wouldn't need to deal with an import scoped to an optional module?

I think this is an abstraction leak and if we can modify methods in TLS readers/writers to handle their specific cases and re-raise as some normalized `ConnectionError`, then the outer layers wouldn't need to know of such implementation details.

It looks like `NoSSLError` and `FatalSSLAlert` exceptions in this module are trying to achieve this. New refactoring should attempt the same.

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
+            if _errors._is_acceptable_socket_shutdown_error(sock_err):
+                return
+            raise
```
webknjaz: I think at some point I wanted to fully migrate to the Python 3 exception hierarchy and handle the entire class of errors with this elegance:
```suggestion
        except ConnectionError:
            return
```
If we need to exclude some subclasses, we'd just do
```suggestion
        except ConnectionRefusedError:
            raise
        except ConnectionError:
            return
```

That's another reason I wanted to push library-specific exception handling to respective modules that would transform their own exceptions into something common. Perhaps a Cheroot-specific subclass of `ConnectionError` (or of its subclasses), for example.

Could you see if you'd be able to sketch such a solution?

cheroot/test/test_makefile.py
webknjaz: I didn't realize. Okay, keep this change. But I think `test_flush_unlocked_handles_blocking_io_error` belongs in this test module. I was trying to keep a mapping of `module.py <-> test_module.py` in the test dir. It seems you're mostly testing behavior of makefile in there anyway.

docs/changelog-fragments.d/779.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed race conditions to make socket I/O more resilient during connection teardown.
```
webknjaz:
```suggestion
Socket I/O is now resilient to race conditions happening during connection teardown
due to sockets dying independently or being closed externally.
```

docs/conf.py
```diff
@@ -116,17 +116,17 @@
...
+    r'https://github\.com/cherrypy/cherrypy/commit',
     r'https://github\.com/cherrypy/cherrypy/issues',
     r'https://github\.com/cherrypy/cherrypy/pull',
     # Has an ephemeral anchor (line-range) but actual HTML has separate per-
     # line anchors.
-    r'https://github\.com'
+    r'https://github\.com',
     r'/python/cpython/blob/c39b52f/Lib/poplib\.py#L297-L302',
-    r'https://github\.com'
+    r'https://github\.com',
```
webknjaz: These are wrong. There's implicit string concatenation that's was happening. Your change breaks it into individual strings in the list.

docs/conf.py
```diff
@@ -116,17 +116,17 @@
     # Too many links to GitHub so they cause
     # "429 Client Error: too many requests for url"
     # Ref: https://github.com/sphinx-doc/sphinx/issues/7388
-    r'https://github\.com/cherrypy/cheroot/commit'
+    r'https://github\.com/cherrypy/cheroot/commit',
     r'https://github\.com/cherrypy/cheroot/issues',
     r'https://github\.com/cherrypy/cheroot/pull',
-    r'https://github\.com/cherrypy/cherrypy/commit'
+    r'https://github\.com/cherrypy/cherrypy/commit',
```
webknjaz, line 104: This change is correct, I think I'll just extract it directly into `main`. Thanks!

webknjaz: I think this might be the reason for the linkcheck job taking this long in CI.

docs/conf.py
webknjaz, line 1: Now that the relevant changes have been absorbed into `main`, you can drop them from this PR.

cheroot/test/test_errors.py
```diff
@@ -41,3 +45,218 @@ def test_plat_specific_errors(err_names, err_nums):
...
+    ``self.args[0]``, which is needed for handling
+    ``_pyopenssl_syscall_errors`` in ``_is_acceptable_socket_shutdown_error()``.
+    No additional methods are needed.
+    """
+
+
+UNACCEPTABLE_ERR_NUM = 99999
+
+# Combine known and unknown cases for both types of exceptions
+TEST_CASES = (
```
webknjaz: FTR, I'd expect to use explicit test IDs for these.

cheroot/test/test_errors.py
```diff
@@ -41,3 +45,218 @@ def test_plat_specific_errors(err_names, err_nums):
...
+
+    # 2. Instantiate the error
+    if exc_class is OSError:
+        err = exc_class(err_num, f'Mocked Error: {err_name}')
+    elif exc_class is MockSysCallError:
+        # MockSysCallError sets the error code at args[0] for the test
+        err = exc_class(err_num, f'Mocked SysCall Error: {err_name}')
+    else:
+        # Handle other types if necessary
+        raise ValueError('Invalid exception class for test case')
```
webknjaz: This much logic seems suspicious. Let's circle back once the abstraction leak I mentioned earlier is handled.

cheroot/errors.py
```diff
@@ -87,4 +99,26 @@ def plat_specific_errors(*errnames):
...
+    if isinstance(sock_err, acceptable_sock_shutdown_exceptions):
+        return True
+
+    # Check for generic OSError/SysCallError with expected errno codes
+    error_code = None
+    if isinstance(sock_err, OSError):
+        error_code = sock_err.errno
+    # Assuming SysCallError has the code at index 0
+    elif isinstance(sock_err, _pyopenssl_syscall_errors):
+        error_code = sock_err.args[0]
```
webknjaz: Perhaps, addressing the abstraction separation suggested in https://github.com/cherrypy/cheroot/pull/779/files#r2446506618 will make such hacks redundant.

cheroot/test/test_makefile.py
julianz-: i had to move some stuff out of test_makefile.py as one of the linters complained about too many modules?

webknjaz: We can override it when it's justified. Perhaps, at some point it'll make sense to have subdirs but that's out of the scope. You can append `WPS202` to https://github.com/cherrypy/cheroot/blob/4a8dc433d4805ed3449639eeee13cbf7856ae7d6/.flake8#L136C70-L136C84 and it won't bother you anymore.

cheroot/makefile.py
```diff
@@ -32,8 +34,45 @@ def _flush_unlocked(self):
...
+        times without side effects). It gracefully handles a race condition
+        where the underlying socket may have already been closed by the remote
+        client or another thread.
+
+        A :exc:`ConnectionError` or :exc:`OSError` with
+        :data:`~errno.EBADF` or :data:`~errno.ENOTCONN` is caught
+        and ignored, as these indicate a normal, expected connection teardown.
+        Other exceptions are re-raised.
+        """
+        if self.closed:  # pylint: disable=using-constant-test
```
webknjaz, line 63: Could you explain why it's acceptable to ignore this violation? What's the justification?

https://github.com/pylint-dev/pylint/issues/3484 / https://github.com/pylint-dev/pylint/issues/9556 imply that pylint gets confused when it sees unknown property decorators.

Would it be possible to fix it by extending https://github.com/cherrypy/cheroot/blob/4a8dc433d4805ed3449639eeee13cbf7856ae7d6/.pylintrc#L316-L317 ?

If not, this needs a code comment explaining that this is because of a pylint bug and linking respective references.

cheroot/errors.py
```diff
@@ -3,6 +3,12 @@
 import errno
 import sys
 
+from . import _compat
+
+
+class ConnectionError(Exception):
+    """Exception raised when a connection closure occurs."""
+
```
webknjaz: Let's have Cheroot-specific naming plus mirror the stdlib hierarchy while inheriting stdlib exceptions. Plus, we probably shouldn't shaddow builtins anyway.

This has a chance of becoming a good public-facing API, so we should work on the naming too.

Something like this and so on:
```suggestion
class CherootConnectionError(ConnectionError):
    """An umbrella exception for connection-related problems."""

class CherootConnectionResetError(ConnectionResetError, CherootConnectionError):
    """An exception corresponding to :data:`~errno.ECONNRESET` when a peer sends a TCP RST on the connection."""
```

Do this for the subclasses listed in https://docs.python.org/3/library/exceptions.html#ConnectionError. And perhaps think of possible other cases.
Then, use these specific exceptions when re-raising in `pyopenssl.py` and other places.

Mentioning connection closure is unnecessary because these exceptions should be usable in any context when working with network connections, not just in `.close()`.

cheroot/errors.py
```diff
@@ -66,19 +72,23 @@ def plat_specific_errors(*errnames):
...
+    errno.EBADF,
     errno.ENOTCONN,
     errno.EPIPE,
     errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
     errno.ECONNRESET,  # corresponds to ConnectionResetError in Python 3
+    *((errno.WSAENOTSOCK,) if _compat.IS_WINDOWS else ()),
 }
 """Errors that may happen during the connection close sequence.
 
+* EBADF - Bad file descriptor
```
webknjaz: The rest of the codes explain what they may mean in the context, they don't just repeat error strings from the docs. Could you do the same for this one?

cheroot/ssl/pyopenssl.py
```diff
@@ -269,6 +271,41 @@ def __init__(self, *args):
...
+            # These are error codes that makefile.py and server.py might expect
+            if error_code in {errno.EBADF, errno.ENOTCONN, errno.ESHUTDOWN}:
+                # Translate the OpenSSL error to a cheroot error
+                raise errors.ConnectionError(
+                    error_code,
+                    os.strerror(error_code),
+                ) from err
+
+            # If it's another SysCallError, let it propagate
+            raise
```
webknjaz: Let's fully map the specific exceptions where possible and always convert the exception without reraising so that the outer layer always knows how to handle it w/o the internal exception leaking up the stack.

```suggestion
        except SSL.SysCallError as ssl_syscall_err:
            connection_error_map = {
                errno.EBADF: errors.CherootConnectionError,  # socket is gone?
                errno.ECONNABORTED: errors.CherootConnectionAbortedError,
                errno.ECONNREFUSED: errors.CherootConnectionRefusedError,
                errno.ECONNRESET: errors.CherootConnectionResetError,
                errno.ENOTCONN: errors.CherootConnectionError,
                errno.EPIPE: errors.CherootBrokenPipeError,
                errno.ESHUTDOWN: errors.CherootBrokenPipeError,
            }

            error_code = ssl_syscall_err.args[0] if ssl_syscall_err.args else None
            error_msg = os.strerror(error_code) if error_code is not None else repr(ssl_syscall_err)

            conn_err_cls = connection_error_map.get(error_code, errors.CherootConnectionError)

            raise conn_err_cls(
                error_code,
                f'Faied to {method_name!s} the PyOpenSSL connection: {error_msg!s}',
            ) from ssl_syscall_err
```

cheroot/errors.py
webknjaz: Now that I'm thinking about it, perhaps it's too early to introduce out exception hierarchy. It might be an overkill for now. How about we try to just rely on stdlib?

cheroot/ssl/pyopenssl.py
webknjaz: Per https://github.com/cherrypy/cheroot/pull/779/files#r2448628303, let's try out just stdlib for now, we can always add own exception classes later.
```suggestion
        except SSL.SysCallError as ssl_syscall_err:
            connection_error_map = {
                errno.EBADF: ConnectionError,  # socket is gone?
                errno.ECONNABORTED: ConnectionAbortedError,
                errno.ECONNREFUSED: ConnectionRefusedError,
                errno.ECONNRESET: ConnectionResetError,
                errno.ENOTCONN: ConnectionError,
                errno.EPIPE: BrokenPipeError,
                errno.ESHUTDOWN: BrokenPipeError,
            }

            error_code = ssl_syscall_err.args[0] if ssl_syscall_err.args else None
            error_msg = os.strerror(error_code) if error_code is not None else repr(ssl_syscall_err)

            conn_err_cls = connection_error_map.get(error_code, errors.CherootConnectionError)

            raise conn_err_cls(
                error_code,
                f'Faied to {method_name!s} the PyOpenSSL connection: {error_msg!s}',
            ) from ssl_syscall_err
```

cheroot/server.py
```diff
@@ -1288,7 +1300,17 @@ def send_headers(self):  # noqa: C901  # FIXME
...
+            self.conn.wfile.write(EMPTY.join(buf))
+        except OSError as sock_err:
+            # Ignore errors indicating the client already closed the connection,
+            # which is expected during a race condition.
+            if errors._is_acceptable_socket_shutdown_error(sock_err):
+                # The socket is already closed
+                self.close_connection = True
+                self.conn.close()
+                return
+            raise
```
webknjaz: Looks like with the exception conversion, we might be able to get away with just
```suggestion
        except ConnectionError:
            # Ignore errors indicating the client already closed the connection,
            # which is expected during a race condition.
            self.close_connection = True  # The socket is already closed
            self.conn.close()
```
here and other places.

cheroot/server.py
webknjaz: Even better, if we get to properly reraise all exceptions.
```suggestion
        except ConnectionError:
            # The socket is gone, so just ignore this error.
            return
```

webknjaz: This could even be
```suggestion
        with contextlib.suppress(ConnectionError):
            self.conn.wfile.write(data)
```
eventually.

docs/conf.py
webknjaz: @julianz- this still needs reverting.

cheroot/test/test_errors.py
```diff
@@ -41,3 +45,186 @@ def test_plat_specific_errors(err_names, err_nums):
...
+
+    # 3. Call the function under test
+    is_shutdown_error = errors._is_acceptable_socket_shutdown_error(err)
+
+    assert is_shutdown_error is expected, (
+        f'Failed test for {exc_class.__name__} with error code {err_num} ({err_name}). '
+        f'Expected {expected}, got {is_shutdown_error}.'
+    )
```
webknjaz: Doesn't `--showlocals` already expose these values?
```suggestion
    assert is_shutdown_error is expected
```

cheroot/test/test_makefile.py
julianz-: good to know thanks!

cheroot/makefile.py
julianz-: I added comment

cheroot/makefile.py
```diff
@@ -32,8 +34,47 @@ def _flush_unlocked(self):
...
+        """
+        # pylint incorrectly flags inherited self.closed property as constant
+        # pylint: disable=using-constant-test
+        if self.closed:
+            return
+
+        try:
+            super().close()
+        except _errors.acceptable_sock_shutdown_exceptions:
+            return
```
webknjaz: These are all subclasses of `ConnectionError`. So having this block is an unnecesasry duplication of the following one.
```suggestion
```

cheroot/makefile.py
```diff
@@ -32,8 +34,47 @@ def _flush_unlocked(self):
                 n = self.raw.write(bytes(self._write_buf))
             except io.BlockingIOError as e:
                 n = e.characters_written
+            except OSError as sock_err:
+                error_code = sock_err.errno
+                if error_code in _errors.acceptable_sock_shutdown_error_codes:
+                    # The socket is gone, so just ignore this error.
```
webknjaz: Thinking of the refactoring being made with exception hierarchy, this would also be replaced.

I wonder, however, if it's right to ignore the problem here. I think that if the connection is broken, such an exception should bubble up to the layer where the reset of the data writing context exists. This would probably be the caller. And not just `write()` but whatever attempts writing. That layer would know how it needs to handle connection errors.

If we just suppress them, then `write()` would pretend to have written something into the socket successfully.

cheroot/errors.py
```diff
@@ -87,4 +93,23 @@ def plat_specific_errors(*errnames):
 * https://docs.microsoft.com/windows/win32/api/winsock/nf-winsock-shutdown
 """
 
-acceptable_sock_shutdown_exceptions = (BrokenPipeError, ConnectionResetError)
+
+def _is_acceptable_socket_shutdown_error(sock_err):
```
webknjaz: Looks like with the exception conversion this function is becoming redundant.

cheroot/makefile.py
```diff
@@ -32,8 +34,47 @@ def _flush_unlocked(self):
...
+        where the underlying socket may have already been closed by the remote
+        client or another thread.
+
+        A :exc:`ConnectionError` or :exc:`OSError` with
+        :data:`~errno.EBADF` or :data:`~errno.ENOTCONN` is caught
+        and ignored, as these indicate a normal, expected connection teardown.
+        Other exceptions are re-raised.
+        """
+        # pylint incorrectly flags inherited self.closed property as constant
+        # pylint: disable=using-constant-test
```
webknjaz: Either use
```suggestion
        # pylint: disable-next=using-constant-test
```
or better keep it on the same line. Using bare `disable` on a dedicated line makes pylint suppress this violation through the end of the file.

cheroot/server.py
```diff
@@ -1192,9 +1192,12 @@ def write(self, chunk):
         if self.chunked_write and chunk:
             chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
             buf = [chunk_size_hex, CRLF, chunk, CRLF]
-            self.conn.wfile.write(EMPTY.join(buf))
+            data = EMPTY.join(buf)
         else:
-            self.conn.wfile.write(chunk)
+            data = chunk
+
+        with contextlib.suppress(ConnectionError):
```
webknjaz: Looking at the layering, I think that something up the stack should process this exception, not a low-ish level write method. What calls it?

webknjaz: Looking deeper, I see that `_process_connections_until_interrupted()` in thread pool is already set up to process connection errors. So this can just be dropped.

If an network error occurrs anytime while processing a single request, there's nothing to do but just log it. And such things should be happening on the top level. Let the connection errors bubble up all the way to that layer and don't suppress them.

The only thing we need to make sure of is that any network errors that present themselves in other ways, are turned into one of the specific connection error exceptions.

cheroot/server.py
```diff
@@ -1192,9 +1192,12 @@ def write(self, chunk):
...
+            self.conn.wfile.write(data)
```
```suggestion
            self.conn.wfile.write(chunk)
```

cheroot/server.py
```diff
@@ -1192,9 +1192,12 @@ def write(self, chunk):
         if self.chunked_write and chunk:
             chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
             buf = [chunk_size_hex, CRLF, chunk, CRLF]
-            self.conn.wfile.write(EMPTY.join(buf))
+            data = EMPTY.join(buf)
```
```suggestion
            self.conn.wfile.write(EMPTY.join(buf))
```

cheroot/server.py
```diff
@@ -1288,7 +1291,13 @@ def send_headers(self):  # noqa: C901  # FIXME
...
             buf.append(k + COLON + SPACE + v + CRLF)
         buf.append(CRLF)
-        self.conn.wfile.write(EMPTY.join(buf))
+        try:
+            self.conn.wfile.write(EMPTY.join(buf))
+        except ConnectionError:
+            # Ignore errors indicating the client already closed the connection,
+            # which is expected during a race condition.
+            self.close_connection = True  # The socket is already closed
+            self.conn.close()
```
webknjaz: Per https://github.com/cherrypy/cheroot/pull/779/files#r2450086992, let it bubble up:
```suggestion
            self.conn.close()
            raise
```

webknjaz: Actually... `_process_connections_until_interrupted()` already has logic to close connections. So none of that duplication is necessary here. Just revert it.
```suggestion
        self.conn.wfile.write(EMPTY.join(buf))
```

cheroot/server.py
webknjaz, line 1: I think I saw a few places in this module where `socket.error` / `OSError` is being suppressed. I haven't checked them all but it seems they need to be unshielded too so that the connection error bubbles to the top properly.

cheroot/ssl/pyopenssl.py
```diff
@@ -269,6 +271,54 @@ def __init__(self, *args):
...
+        try:
+            return self._safe_close_call('shutdown')
+        finally:
+            self._lock.release()
+
+    def _safe_close_call(self, method_name):
+        """Handle SysCallError during close/shutdown."""
+        try:
+            # Call the proxied method (e.g., self._ssl_conn.close())
+            return getattr(self._ssl_conn, method_name)()
```
webknjaz: Let's just move calling `self._ssl_conn.close()` / `self._ssl_conn.shutdown()` into the methods above. And put the below exception handling/conversion into a decorator. It's quite easy to do with `@contextlib.contextmanager`.

cheroot/ssl/pyopenssl.py
```diff
@@ -269,6 +271,54 @@ def __init__(self, *args):
...
+                ssl_syscall_err.args[0] if ssl_syscall_err.args else None
+            )
+            error_msg = (
+                os.strerror(error_code)
+                if error_code is not None
+                else repr(ssl_syscall_err)
+            )
+            conn_err_cls = connection_error_map.get(
+                error_code,
+                errors.CherootConnectionError,
```
webknjaz: Missed this place.

cheroot/ssl/pyopenssl.py
```diff
@@ -269,6 +271,54 @@ def __init__(self, *args):
...
+            )
+            raise conn_err_cls(
+                error_code,
+                f'Faied to {method_name!s} the PyOpenSSL connection: {error_msg!s}',
+            ) from ssl_syscall_err
```
webknjaz: Something like this should do:
```suggestion
    @_morph_syscall_to_connection_error('close')
    def close(self):
        """Close the connection, translating OpenSSL errors for shutdown."""
        with self._lock:
            return self._ssl_conn.close()

    @_morph_syscall_to_connection_error('shutdown')
    def shutdown(self):
        """Shutdown the connection, translating OpenSSL errors."""
        with self._lock:
            return self._ssl_conn.shutdown()


@contextlib.contextmanager
def _morph_syscall_to_connection_error(method_name, /):
    """Handle :exc:`SSL.SysCallError` in a wrapped method."""
    try:
        yield
    except SSL.SysCallError as ssl_syscall_err:
        connection_error_map = {
            errno.EBADF: ConnectionError,  # socket is gone?
            errno.ECONNABORTED: ConnectionAbortedError,
            errno.ECONNREFUSED: ConnectionRefusedError,
            errno.ECONNRESET: ConnectionResetError,
            errno.ENOTCONN: ConnectionError,
            errno.EPIPE: BrokenPipeError,
            errno.ESHUTDOWN: BrokenPipeError,
        }
        error_code = (
            ssl_syscall_err.args[0] if ssl_syscall_err.args else None
        )
        error_msg = (
            os.strerror(error_code)
            if error_code is not None
            else repr(ssl_syscall_err)
        )
        conn_err_cls = connection_error_map.get(
            error_code,
            ConnectionError,
        )
        raise conn_err_cls(
            error_code,
            f'Faied to {method_name!s} the PyOpenSSL connection: {error_msg!s}',
        ) from ssl_syscall_err


```

cheroot/makefile.py
```diff
@@ -32,8 +34,47 @@ def _flush_unlocked(self):
...
+                if error_code in _errors.acceptable_sock_shutdown_error_codes:
+                    # The socket is gone, so just ignore this error.
+                    return
+                raise
+            else:
+                # The 'try' block completed without an exception
+                if n == 0:
+                    # This could happen with non-blocking write
+                    # when nothing was written
+                    break
```
webknjaz: I'm now questioning the addition of this whole thing. Or at least the new `except` block. Yet to hear your thoughts about the `else` portion.

cheroot/makefile.py
julianz-: Good point. Maybe just remove this branch then and let the errors rise.

cheroot/makefile.py
julianz-: I've removed the OSError exception handling as I agree with your earlier point we should just let those errors rise. Regarding the else block - I've changed to a simple if statement to check for n == 0 to prevent an infinite loop.

julianz-:
Thank you @webknjaz for all your great suggestions and points. I will take a look at everything. The layering is actually more complicated than I had realized.

webknjaz:
> The layering is actually more complicated than I had realized.

Yep, that's why I have incomplete pieces of code somewhere locally that I never finalized. It does require some effort..

cheroot/makefile.py
```diff
@@ -32,8 +34,44 @@ def _flush_unlocked(self):
...
+        Other exceptions are re-raised.
+        """
+        # pylint incorrectly flags inherited self.closed property as constant
+        if self.closed:  # pylint: disable=using-constant-test
+            return
+
+        try:
+            super().close()
+        except _errors.acceptable_sock_shutdown_exceptions:
+            return
```
webknjaz: This is redundant because `_errors.acceptable_sock_shutdown_exceptions` only contains subclasses of `ConnectionError`.
```suggestion
```

cheroot/server.py
```diff
@@ -1192,9 +1192,11 @@ def write(self, chunk):
         if self.chunked_write and chunk:
             chunk_size_hex = hex(len(chunk))[2:].encode('ascii')
             buf = [chunk_size_hex, CRLF, chunk, CRLF]
-            self.conn.wfile.write(EMPTY.join(buf))
+            data = EMPTY.join(buf)
         else:
-            self.conn.wfile.write(chunk)
+            data = chunk
+
+        self.conn.wfile.write(data)
```
webknjaz: Let's revert this refactoring since it's not related to race conditions. It's best to make sure that the PR only contains a single type of change. Any refactoring or formatting of unrelated parts can go into separate PRs.

That said, it might be reasonable to drop things like `except socket.error` in some places in this module. But each case should be investigated for consequences.

cheroot/server.py
julianz-: There's no changes to server.py now.

cheroot/ssl/pyopenssl.py
webknjaz, line 1: I think the current state of this module is good but Codecov shows that none of the newly added lines are covered with tests. We'll have to fix this before merging.

It looks like the updates to `cheroot.ssl.pyopenssl` could go into a standalone pull request. Could you extract this and add tests+changenote in a separate PR? I hope this will simplify reviewing this one.

cheroot/errors.py
```diff
@@ -66,19 +68,23 @@ def plat_specific_errors(*errnames):
 
 
 acceptable_sock_shutdown_error_codes = {
+    errno.EBADF,
     errno.ENOTCONN,
     errno.EPIPE,
     errno.ESHUTDOWN,  # corresponds to BrokenPipeError in Python 3
     errno.ECONNRESET,  # corresponds to ConnectionResetError in Python 3
+    *((errno.WSAENOTSOCK,) if _compat.IS_WINDOWS else ()),
```
webknjaz: Does this exception correspond to `EBADF` semantically? If so, I'd probably keep them close to each other as suggested in https://github.com/cherrypy/cheroot/pull/779/files#r2432797948.

Additionally, could you add a comment hinting what might be causing it in practice?

cheroot/ssl/pyopenssl.py
webknjaz, line 218: Will this interferre with the newly added decorator/methods? `proxy_wrapper()` below does effectively what we did with `@_morph_syscall_to_connection_error` and so `shutdown()` will end up double-decorated/locked if I'm reading this correctly.

webknjaz, line 198: Will this interferre with the newly added decorator/methods? `proxy_wrapper()` below does effectively what we did with `@_morph_syscall_to_connection_error` and so `shutdown()` will end up double-decorated/locked if I'm reading this correctly.

cheroot/test/test_makefile.py
```diff
@@ -51,3 +57,116 @@ def test_bytes_written():
...
+    mocker.patch.object(mock_buffer_writer.raw, 'write', new=mock_write_method)
+
+    # Check the initial state of the buffer
+    initial_len = len(mock_buffer_writer._write_buf)
+
+    # 4. Execute the code
+    try:
+        mock_buffer_writer._flush_unlocked()
+    except Exception as exc:
+        pytest.fail(f'Unexpected exception raised: {type(exc).__name__}')
```
webknjaz: It's unclear why this needs to be wrapped with try/except: tests are supposed to be as simple as possible — adding more logic makes it difficult to reason about them. Additionally, any exception would bubble up and render in the output with the entire stack and respective context attached unlike a short failure message with the stack being lost.

cheroot/errors.py
webknjaz: @julianz- missed this question?

cheroot/makefile.py
```diff
@@ -32,8 +34,42 @@ def _flush_unlocked(self):
...
+
+        try:
+            super().close()
+        except ConnectionError:
+            return
+        except OSError as err:
+            # Handle EBADF and other acceptable socket shutdown errors
+            if err.errno in _errors.acceptable_sock_shutdown_error_codes:
+                return
```
webknjaz, line 73: @julianz- do you know if error codes listed there contain things that aren't represented by `ConnectionError`? I'm wondering if newer Python versions would ever hit this code branch — is it possible that the above `ConnectionError` will handle this?

cheroot/makefile.py
```diff
@@ -32,8 +34,42 @@ def _flush_unlocked(self):
...
+            raise
```
webknjaz, line 74: I wish we have proper logging infra here to see what can actually happen. The coverage report shows that this line is never exercised in the tests.

webknjaz: Perhaps, we could have a warning emitted, though..

cheroot/makefile.py
```diff
@@ -32,8 +34,42 @@ def _flush_unlocked(self):
...
+        Other exceptions are re-raised.
+        """
+        # pylint incorrectly flags inherited self.closed property as constant
+        if self.closed:  # pylint: disable=using-constant-test
+            return
+
+        try:
+            super().close()
+        except ConnectionError:
+            return
```
webknjaz, line 69: Looks like tests never hit this.

cheroot/makefile.py
webknjaz: Also, do we know what's causing the tests to always hit this line?
https://app.codecov.io/gh/cherrypy/cheroot/pull/779?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=cherrypy#d99a7689c2159eab73db39f9e5b89a1d-R70.

It's best to make sure that this isn't due to some mocks but a real simulated scenario.

cheroot/makefile.py
```diff
@@ -32,8 +34,42 @@ def _flush_unlocked(self):
                 n = self.raw.write(bytes(self._write_buf))
             except io.BlockingIOError as e:
                 n = e.characters_written
+
+            if n == 0:
+                # If nothing was written we need to break
+                # to avoid infinte loops
+                break
```
webknjaz, line 44: The tests never hit this line. We should come up with a new regression test.

cheroot/errors.py
julianz-: Sorry missed these from October! Now addressed.

cheroot/ssl/pyopenssl.py
julianz-: Added tests and pyopenssl was updated in #800 and #801.

cheroot/test/test_makefile.py
julianz-: This is gone from the current version

cheroot/makefile.py
julianz-: EPIPE/ESHUTDOWN -> BrokenPipeError and ECONNRESET -> ConnectionResetError are ConnectionError subclasses. But EBADF and ENOTCONN are plain OSErrors, so the `except OSError` branch is needed for those. I added test_close_error_handling[ebadf] to cover this path.

Regarding the coverage, the Codecov report is for an older commit — the code has been significantly refactored since then. Not sure whether your question is still a concern?

cheroot/makefile.py
julianz-: Covered now. Logging would be nice but maybe for another PR?

cheroot/makefile.py
julianz-: Stale code i think. All lines in makefile.py covered now.

cheroot/makefile.py
julianz-: ditto

cheroot/makefile.py
```diff
@@ -32,8 +34,45 @@
...
             except io.BlockingIOError as e:
                 n = e.characters_written
+
+            if n == 0:
+                # Non-blocking socket can't write right now — stop flushing.
+                break
+            if not n:
+                # Defensive: write() returned None or other falsy value,
+                # which shouldn't happen but could cause an infinite loop.
+                break
```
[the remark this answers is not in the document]
julianz-: I think CodeQL is assuming n is an int but it's not clear that it has to be.

PR #786, opened 2025-10-31
Re-raise PyOpenSSL `SysCallError`s as respective errno-mapped standard Python `ConnectionError` subclasses in `close()`/`shutdown()`

julianz-:
Adds a more robust error-handling mechanism to the `PyOpenSSL` `SSLConnection` wrapper to better manage exceptions raised during connection shutdown and close operations.

The `pyopenssl.py` module now includes the context manager, `_morph_syscall_to_connection_error`, to wrap error-prone methods. This context manager intercepts low-level `OpenSSL.SSL.SysCallError` exceptions and translates them into standard Python exception types (e.g., `ConnectionAbortedError`, `BrokenPipeError`, `ConnectionRefusedError`, etc.) based on the underlying `errno` code.

Why this is useful: Currently, when a connection fails due to a system event (like a remote peer abruptly closing the socket), the resulting `SysCallError` can be difficult to catch and handle cleanly in application code.

This change ensures that applications using Cheroot can reliably catch a standard `ConnectionError` or one of its subclasses, improving the robustness and usability of the SSL connection handling.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/786)


docs/changelog-fragments.d/786.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+OpenSSL system call errors are now intercepted and reraised as standard Python ConnectionErrors.
```
webknjaz: Since we have the upstream CPython docs plugged via intersphinx, we can make them clickable.
```suggestion
OpenSSL system call errors are now intercepted and reraised as standard Python :exc:`ConnectionErrors`.
```

docs/changelog-fragments.d/786.bugfix.rst
webknjaz, line 1: Change log preview: https://cheroot--786.org.readthedocs.build/en/786/history/#to-be-included-in-the-next-release

cheroot/ssl/pyopenssl.py
webknjaz, line 1: As a pre-requisite for merging, we'll need to make this diff https://app.codecov.io/gh/cherrypy/cheroot/pull/786?flags%5B0%5D=pytest#b7eabf7f92af290323d5c8008601e930-R93 green.

cheroot/ssl/pyopenssl.py
```diff
@@ -229,7 +270,9 @@ def proxy_wrapper(self, *args):
                     new_args = (
                         args[:] if method not in proxy_methods_no_args else []
                     )
-                    return getattr(self._ssl_conn, method)(*new_args)
+                    # translate any SysCallError to ConnectionError
+                    with _morph_syscall_to_connection_error(method):
+                        return getattr(self._ssl_conn, method)(*new_args)
```
webknjaz: I guess this is one way of solving that.. we could do so for now. Although, I'd be good to get rid of `SSLConnectionProxyMeta` in some future iteration/PR. I've always found this metaclass patching confusing and wanted to simplify it if possible.

cheroot/ssl/pyopenssl.py
```diff
@@ -229,7 +270,9 @@ def proxy_wrapper(self, *args):
...
                 finally:
                     self._lock.release()
```
webknjaz: Maybe you could move locking into the same `with:`-block?
```python
with self._lock, _morph_syscall_to_connection_error(method):
    return getattr(self._ssl_conn, method)(*new_args)
```
Would be easier to track.

---

Or even
```python
with _morph_syscall_to_connection_error(method), self._lock:
    return getattr(self._ssl_conn, method)(*new_args)
```

cheroot/test/test_ssl_pyopenssl.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    # Dynamically retrieve the method on the underlying mock
+    underlying_method = getattr(conn._ssl_conn, tested_method_name)
+
+    # Patch the method to raise the simulated error
+    underlying_method.side_effect = simulated_error
+
+    # Assert the expected exception is raised based on the parameter
+    with pytest.raises(expected_exception) as excinfo:
```
webknjaz: `pytest.raises()` should always make use of `match=`.

cheroot/test/test_ssl_pyopenssl.py
```diff
@@ -0,0 +1,75 @@
...
+    ('shutdown', errno.EPIPE, BrokenPipeError),  # Test 'shutdown' here
+    (
+        'shutdown',
+        errno.ECONNRESET,
+        ConnectionResetError,
+    ),  # Test 'shutdown' here
+    ('close', errno.ENOTCONN, ConnectionError),
+    ('close', errno.EPIPE, BrokenPipeError),
+    ('close', errno.ESHUTDOWN, BrokenPipeError),
+)
```
webknjaz: Since patching now applies to more methods, we should probably use more names, not just `shutdown` / `close`.

Additionally, since the list is not reused, it's better to move it into the parametrize definition, so it's close to the param names list and isn't disconnected. These are coupled, after all — no need to keep them in separate places.

Finally, these should have better test IDs w/o unimportant bits. Like just `errno` names, for example. This can be achieved by using `ids=` with a callable or by wrapping each param with a `pytest.item(id=)`.

cheroot/test/ssl/test_ssl_pyopenssl.py
webknjaz, line 1: Let's move this to ‎`cheroot/test/ssl/test_pyopenssl.py` to mimic the directory layout of the main project.

cheroot/test/ssl/test_ssl_pyopenssl.py
```diff
@@ -0,0 +1,78 @@
...
+    simulated_errno,
+    expected_exception,
+):
+    """Check ``SSLConnection.close()`` morphs ``SysCallError`` to ``ConnectionError``."""
+    # Prevents the real OpenSSL.SSL.Connection.__init__ from running
+    mocker.patch('OpenSSL.SSL.Connection')
+
+    # Create SSLConnection object. The patched SSL.Connection() call returns
+    # a mock that is stored internally as conn._ssl_conn.
+    conn = SSLConnection(mock_ssl_context)
```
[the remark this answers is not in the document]
julianz-: i have now added @pytest.mark.filterwarnings('ignore:Non-callable called') to suppress this false positive. We are calling the mocked initializer with another mock.

cheroot/ssl/pyopenssl.py
julianz-: I tried refactoring to remove the metaclass and I think it works but will add as a new PR.

docs/changelog-fragments.d/786.bugfix.rst
julianz-: is ok now?

webknjaz: Yep, marking as resolved.

cheroot/test/ssl/test_ssl_pyopenssl.py
webknjaz: No need to have the second `ssl_`.

cheroot/ssl/pyopenssl.py
```diff
@@ -77,6 +79,45 @@
...
+@contextlib.contextmanager
+def _morph_syscall_to_connection_error(method_name, /):
+    """
+    Handle :exc:`OpenSSL.SSL.SysCallError` in a wrapped method.
+
+    This context manager catches and re-raises SSL system call errors
+    with appropriate exception types.
+
+    Yields:
+        None: Execution continues within the context block.
```
webknjaz, line 91: we should be using sphinx-native param lists, not napoleon.

cheroot/ssl/pyopenssl.py
```diff
@@ -77,6 +79,45 @@
...
+    """  # noqa: DAR301
```
webknjaz, line 92: This should be possible to drop

```suggestion
    """
```

.flake8
```diff
@@ -132,6 +132,7 @@ per-file-ignores =
   cheroot/ssl/pyopenssl.py: C815, DAR101, DAR201, DAR401, I001, I003, I005, N801, N804, RST304, WPS100, WPS110, WPS111, WPS117, WPS120, WPS121, WPS130, WPS210, WPS220, WPS221, WPS225, WPS229, WPS231, WPS238, WPS301, WPS335, WPS338, WPS420, WPS422, WPS430, WPS432, WPS501, WPS504, WPS505, WPS601, WPS608, WPS615
   cheroot/test/conftest.py: DAR101, DAR201, DAR301, I001, I003, I005, WPS100, WPS130, WPS325, WPS354, WPS420, WPS422, WPS430, WPS457
   cheroot/test/helper.py: DAR101, DAR201, DAR401, I001, I003, I004, N802, WPS110, WPS111, WPS121, WPS201, WPS220, WPS231, WPS301, WPS414, WPS421, WPS422, WPS505
+  cheroot/test/ssl/test_ssl_pyopenssl.py: DAR101, DAR201, WPS226
```
webknjaz, line 135:
```suggestion
  cheroot/test/ssl/test_pyopenssl.py: DAR101, DAR201, WPS226
```

PR #787, opened 2025-11-01
Drop locks from the `cheroot.ssl.pyopenssl.SSLConnection` method

julianz-:
Removes metaclass by using `__getattr__(self, name)` to intercept
method calls.


.DS_Store
webknjaz, line 1: Isn't gitignore supposed to block this?

https://git-scm.com/docs/gitignore#_synopsis / https://stackoverflow.com/a/22885996/595220 / https://gist.github.com/subfuzion/db7f57fff2fb6998a16c?permalink_comment_id=5429330#gistcomment-5429330

.vscode/settings.json
webknjaz, line 1: Another leak

docs/.DS_Store
webknjaz, line 1: This leaked too.

cheroot/ssl/pyopenssl.py
```diff
@@ -175,94 +216,89 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
+    proxy_props = ('family',)
+
+    @staticmethod
+    def _lock_decorator(method):
+        """Create the thread-safe, error-translating proxy method."""
+
+        def proxy_wrapper(self, *args):
+            new_args = (
+                args[:]
+                if method not in SSLConnection.proxy_methods_no_args
```
webknjaz: I wonder why this separation is even necessary just for shutdown. It might've been for some ancient pyopenssl compat.

cheroot/ssl/pyopenssl.py
```diff
@@ -175,94 +216,89 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-    r"""A thread-safe wrapper for an ``SSL.Connection``.
+        def proxy_prop_wrapper(self):
+            return getattr(self._ssl_conn, property_)
 
-    :param tuple args: the arguments to create the wrapped \
-                        :py:class:`SSL.Connection(*args) \
-                        <pyopenssl:OpenSSL.SSL.Connection>`
-    """
+        proxy_prop_wrapper.__name__ = property_
+        return property(proxy_prop_wrapper)
```
webknjaz: I would expect this to be somewhere in the helper, though. So we wouldn't be adding public attributes.
Additionally, I don't fully understand the property patching bit — why is it necessary? Also, do does the connection object even need to be made thread-safe? Why?

.DS_Store
julianz-: Sorry my bad. Hopefully fixed now.

cheroot/ssl/pyopenssl.py
julianz-: Shutdown on Python sockets expects a `how` parameter e.g. `socket.SHUT_WR`.  OpenSSL does not take a similar parameter so PyOpenSSL doesn't either.  It's not clear we need to support this parameter. I do see a couple of places where this question might be considered: 

In `threadpool.py`:
```python
def _force_close(conn):
        if conn.rfile.closed:
            return
        try:
            try:
                conn.socket.shutdown(socket.SHUT_RD)
            except TypeError:
                # pyOpenSSL sockets don't take an arg
                conn.socket.shutdown()
        except OSError:
            # shutdown sometimes fails (race with 'closed' check?)
            # ref #238
            pass
```

and `server.py`
```python
def _close_kernel_socket(self):
        """Terminate the connection at the transport level."""
        # Honor ``sock_shutdown`` for PyOpenSSL connections.
        shutdown = getattr(
            self.socket,
            'sock_shutdown',
            self.socket.shutdown,
        )

        try:
            shutdown(socket.SHUT_RDWR)  # actually send a TCP FIN
        except errors.acceptable_sock_shutdown_exceptions:
            pass
        except socket.error as e:
            if e.errno not in errors.acceptable_sock_shutdown_error_codes:
                raise
```

cheroot/ssl/pyopenssl.py
julianz-: I've revamped the whole thing - getting rid of the meta class, decorator, property patching, and avoiding adding any public attributes. See what you think. Great question about thread-safety. I think you are right - I don't think it's needed as the usual pattern involves one connection per thread. If so we can further simplify the SSLConnection class. Let me know if you think we should remove.

cheroot/ssl/pyopenssl.py
webknjaz: Okay.. So in `server.py` the arg is passed in but in `threadpool.py` it's not. Got it. I'm not sure how I feel that the locking wrappers at all. I wish the hackery was just dropped.

cheroot/ssl/pyopenssl.py
webknjaz: @julianz- so I did some research and here's my observations:
1) mass-searching source code, it seems like only CherryPy/Cheroot (often being vendored into something else) have any references to thread-safety: https://grep.app/search?f.lang=Python&regexp=true&q=%28threadsafe|thread-safe|thread+safe%29.*\n.*\n.*SSL\.Connection (CherryPy is what first had this module until the HTTP/WSGI part was split out of it).
2) mature projects like urllib3 that also make use of PyOpenSSL don't have any locking: https://github.com/urllib3/urllib3/blob/b20c836/src/urllib3/contrib/pyopenssl.py (although, it might be interesting to look at how they organize their socket wrapper)
3) somebody asked about thread-safety upstream and Paul said that it's fine: https://github.com/pyca/pyopenssl/issues/424. But he made a remark about subinterpreters / `mod_wsgi` and I know that CherryPy has some code that is probably not being tested in CI related to `mod_wsgi`. But this was nearly a decade ago.
4) https://stackoverflow.com/a/61826374/595220 hints that prior to OpenSSL 1.1.0 there would be some problems but not anymore. That version hit EOL 6 years ago: https://endoflife.date/openssl. So we shouldn't need to care about it.
5) https://github.com/openssl/openssl/discussions/25343 / https://crypto.stackexchange.com/a/41344/70030 / https://docs.openssl.org/1.0.2/man3/threads/#description claim that low-level `SSL_SESSION` isn't thread-safe. But given the PyOpenSSL issue, I'm inclined to believe that it's fine in the Python land.
6) I did some Git paleontology and the commit that added locking on Sep 12, 2006 https://github.com/cherrypy/cherrypy/commit/f2e3c1c7104dde913ec5b0f6ba7dae35c8a3f317 (https://github.com/cherrypy/cheroot/commit/f2e3c1c7104dde913ec5b0f6ba7dae35c8a3f317); being terrified of `exec()`, I rewrote it to a metaclass on Sep 2, 2018 in https://github.com/cherrypy/cheroot/commit/e9e070beb7186f74af6d767ddc032b91707a43d1
  That commit does not really explain anything but references `pyOpenSSL/tsafe.py`. This was pre-PyPI which explains why things were being copied across projects and/or vendored. The first PyOpenSSL release on PyPI was 0.6 on Jun 4, 2008. So it's already two years past the CherryPy commit. We can't say for sure which version this was copied from. The very first import of PyOpenSSL from some other VCS into Git was on Feb 2, 2008 (https://github.com/pyca/pyopenssl/commit/897bc2556fed43b76f6d1b14470c3e806df15af8) but it's tagged as v0.14a1 while is probably an error or some weird historic artifact. However, that commit does contain the `exec()` snippet with locking that ended up in CherryPy — https://github.com/pyca/pyopenssl/blob/897bc2556fed43b76f6d1b14470c3e806df15af8/tsafe.py. On Jul 28, 2010, that module moved within the repo to https://github.com/pyca/pyopenssl/blob/94cc894795226537daf2f4e79e2d142ff51eacec/OpenSSL/tsafe.py (https://github.com/pyca/pyopenssl/commit/94cc894795226537daf2f4e79e2d142ff51eacec#diff-eb4dc4606a8f26382d033c9859d6b5be8043e8bc511522f2e014e181bd09c09d). More history of this module: https://github.com/pyca/pyopenssl/commits/e41098f271dc96d6c411172bb11bbe5816f78710/src/OpenSSL/tsafe.py. It was deprecated on Sep 4, 2017 in PyOpenSSL v17.3.0 (https://github.com/pyca/pyopenssl/issues/655 / https://github.com/pyca/pyopenssl/pull/673) and removed fully on Nov 27, 2020 in v20.0.0 (https://github.com/pyca/pyopenssl/pull/913). The maintainers thought nobody uses it and weren't sure it was functional. At the time of deletion, it was still using the original `exec()` (with an additional method added to the list throughout its life).

---

So I'd get rid for the hacks and made it as simple as possible, instead of trying to preserve them w/o a good recorded/explainable justification.

webknjaz:
@julianz- linking the thread where I posted more thoughts in case it gets lost in the UI: https://github.com/cherrypy/cheroot/pull/787#discussion_r2484001174.

cheroot/ssl/pyopenssl.py
julianz-: i think we can get rid of the locking wrappers. see next comment.

webknjaz: @julianz- yep, and decorate the methods in #786 directly?

cheroot/ssl/pyopenssl.py
```diff
@@ -68,6 +70,7 @@
     SSL = None
 
 import contextlib
+from typing import NoReturn
```
webknjaz: Could you please change this to `import typing as _t`? There's 
a nice convention that originated in Pallets to always `import typing as t` so that the namespace isn't polluted with unintended public objects: https://github.com/sirosen/flake8-typing-as-t. We also use it in `ansible-core` but lately we're switching to an even more concervative approach and use private variable names. I'd like to start doing it here as well.

cheroot/ssl/pyopenssl.py
```diff
@@ -175,99 +178,77 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-        def make_property(property_):
-            """Create a proxy method for a new class."""
+    def __init__(self, *args):
+        """Initialize the SSLConnection."""
+        self._ssl_conn = SSL.Connection(*args)
+        self._lock = threading.RLock()
 
-            def proxy_prop_wrapper(self):
-                return getattr(self._ssl_conn, property_)
+    def _handle_syscall_error(self, ssl_syscall_err, method_name) -> NoReturn:
```
webknjaz: Can we plz keep the error conversion in a standalone PR? Also, I still prefer this to be a context manager rather than a callable.

cheroot/ssl/pyopenssl.py
julianz-: Yes i think so. In this version, SSLConnection is pretty bare bones but we can add the error handling in #786.

cheroot/ssl/pyopenssl.py
julianz-: Very thorough! But obviously removing the thread-safety is a potentially breaking change if anyone is relying on it.

cheroot/ssl/pyopenssl.py
julianz-: I don't need the NoReturn fix for this PR. It was for the error handling which will be in #786.

cheroot/ssl/pyopenssl.py
julianz-: Very thorough! But obviously removing the thread-safety is a potentially a breaking change if anyone is relying on it but seems like they are not.

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
+
+    def shutdown(self, how=None):
+        """Shutdown the SSL connection.
+
+        how: Ignored.
+        PyOpenSSL's shutdown() method does not accept any arguments.
+        Present here for interface compatibility with Python
+        socket.shutdown(how).
+        """
+        with self._lock:
```
webknjaz: no lock?

cheroot/ssl/pyopenssl.pyi
```diff
@@ -18,11 +18,14 @@ class SSLFileobjectMixin:
...
-    def __new__(mcl, name, bases, nmspc): ...
-
 class SSLConnection:
+    proxy_methods: tuple[str, ...]
+    proxy_methods_no_args: tuple[str, ...]
+    proxy_props: tuple[str, ...]
+
     def __init__(self, *args) -> None: ...
+    def __getattr__(self, name: str) -> Any: ...
+    def shutdown(self, how: Optional[int] = None) -> None: ...
```
webknjaz: `typing.Optional` is considered confusing because it's actually `Union[thing, None]`.
```suggestion
    def shutdown(self, how: int | None = None) -> None: ...
```

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
 
     def __init__(self, *args):
-        """Initialize SSLConnection instance."""
+        """Initialize the SSLConnection."""
         self._ssl_conn = SSL.Connection(*args)
-        self._lock = threading.RLock()
+
+    def __getattr__(self, name):
+        """Proxy attribute access to the underlying ``SSL.Connection``."""
+        return getattr(self._ssl_conn, name)
```
webknjaz: Instead of this, can we just have the class inherit `SSL.Connection`? I'd like to avoid dynamic things that aren't really lintable staticly (by things like MyPy and similar).

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-    r"""A thread-safe wrapper for an ``SSL.Connection``.
+class SSLConnection:
+    """
+    A compatibility wrapper around pyOpenSSL's SSL.Connection.
 
-    :param tuple args: the arguments to create the wrapped \
-                        :py:class:`SSL.Connection(*args) \
-                        <pyopenssl:OpenSSL.SSL.Connection>`
+    This class exists primarily to ensure the standard Python socket method
+    `   shutdown(how=None)` is available for interface compatibility..
```
webknjaz: Docstrings are expected to be RST, so it's either double backticks or proper roles to reference things. Try this:
```suggestion
    :py:meth:`.shutdown` is available for interface compatibility.
```

This is where it's rendered: https://cheroot--787.org.readthedocs.build/en/787/pkg/cheroot.ssl.pyopenssl/#cheroot.ssl.pyopenssl.SSLConnection

.flake8
```diff
@@ -129,9 +129,10 @@ per-file-ignores =
...
-  cheroot/ssl/pyopenssl.py: C815, DAR101, DAR201, DAR401, I001, I003, I005, N801, N804, RST304, WPS100, WPS110, WPS111, WPS117, WPS120, WPS121, WPS130, WPS210, WPS220, WPS221, WPS225, WPS229, WPS231, WPS238, WPS301, WPS335, WPS338, WPS420, WPS422, WPS430, WPS432, WPS501, WPS504, WPS505, WPS601, WPS608, WPS615
+  cheroot/ssl/pyopenssl.py: C815, DAR101, DAR201, DAR401, I001, I003, I005, N801, N804, RST304, WPS100, WPS110, WPS111, WPS117, WPS120, WPS121, WPS130, WPS201, WPS210, WPS220, WPS221, WPS225, WPS229, WPS231, WPS238, WPS301, WPS335, WPS338, WPS420, WPS422, WPS430, WPS432, WPS501, WPS504, WPS505, WPS601, WPS602, WPS608, WPS615
   cheroot/test/conftest.py: DAR101, DAR201, DAR301, I001, I003, I005, WPS100, WPS130, WPS325, WPS354, WPS420, WPS422, WPS430, WPS457
   cheroot/test/helper.py: DAR101, DAR201, DAR401, I001, I003, I004, N802, WPS110, WPS111, WPS121, WPS201, WPS220, WPS231, WPS301, WPS414, WPS421, WPS422, WPS505
+  cheroot/test/ssl/test_ssl_pyopenssl.py: DAR101, DAR201, WPS226
```
webknjaz:
```suggestion
```

.flake8
```diff
@@ -129,9 +129,10 @@ per-file-ignores =
...
   cheroot/ssl/builtin.py: DAR101, DAR201, DAR401, I001, I003, N806, RST304, WPS110, WPS111, WPS115, WPS117, WPS120, WPS121, WPS122, WPS130, WPS201, WPS210, WPS214, WPS229, WPS231, WPS338, WPS422, WPS501, WPS505, WPS529, WPS608, WPS612
-  cheroot/ssl/pyopenssl.py: C815, DAR101, DAR201, DAR401, I001, I003, I005, N801, N804, RST304, WPS100, WPS110, WPS111, WPS117, WPS120, WPS121, WPS130, WPS210, WPS220, WPS221, WPS225, WPS229, WPS231, WPS238, WPS301, WPS335, WPS338, WPS420, WPS422, WPS430, WPS432, WPS501, WPS504, WPS505, WPS601, WPS608, WPS615
+  cheroot/ssl/pyopenssl.py: C815, DAR101, DAR201, DAR401, I001, I003, I005, N801, N804, RST304, WPS100, WPS110, WPS111, WPS117, WPS120, WPS121, WPS130, WPS201, WPS210, WPS220, WPS221, WPS225, WPS229, WPS231, WPS238, WPS301, WPS335, WPS338, WPS420, WPS422, WPS430, WPS432, WPS501, WPS504, WPS505, WPS601, WPS602, WPS608, WPS615
```
webknjaz: What are these exactly?

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-        """Initialize SSLConnection instance."""
+        """Initialize the SSLConnection."""
         self._ssl_conn = SSL.Connection(*args)
-        self._lock = threading.RLock()
+
+    def __getattr__(self, name):
+        """Proxy attribute access to the underlying ``SSL.Connection``."""
+        return getattr(self._ssl_conn, name)
+
+    def shutdown(self, how=None):
```
webknjaz, line 186: I wonder if we should have something for `sock_shutdown()`.

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
+        """Shutdown the SSL connection.
+
+        how: Ignored.
+        PyOpenSSL's shutdown() method does not accept any arguments.
+        Present here for interface compatibility with Python
+        socket.shutdown(how).
```
webknjaz: Let's try to make this better: https://cheroot--787.org.readthedocs.build/en/787/pkg/cheroot.ssl.pyopenssl/#cheroot.ssl.pyopenssl.SSLConnection.shutdown

```suggestion
        :param how: Ignored. PyOpenSSL's :py:meth:`~OpenSSL.SSL.Connection.shutdown` \
                    method does not accept any arguments.\
                    Present here for interface compatibility with Python \
                    :py:meth:`socket.shutdown` that :py:class:`ssl.SSLSocket` wrapper \
                    exposes.
        :type how: object
```

docs/spelling_wordlist.txt
```diff
@@ -44,6 +44,7 @@ positionally
 pre
 preconfigure
 py
+pyOpenSSL
```
webknjaz, line 47: This is probably redundant once its wrapped in the docstring properly.
```suggestion
```

docs/changelog-fragments.d/787.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Removes thread-safety locking logic from
```
webknjaz: Changelog does't do things, it doesn't remove locking. It is supposed to describe what's changed. Use a different writing style to show what the effect is compared to the previous release. It can be the past tense or present simple but not third person.

Maybe this explanation will work for you https://pip-tools.rtfd.io/en/stable/contributing/#rationale

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,32 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-        # Doesn't work via super() for some reason.
-        # Falling back to type() instead:
-        return type(name, bases, nmspc)
-
-
-class SSLConnection(metaclass=SSLConnectionProxyMeta):
-    r"""A thread-safe wrapper for an ``SSL.Connection``.
+class SSLConnection:
+    """
+    A compatibility wrapper around pyOpenSSL's SSL.Connection.
```
```suggestion
    A compatibility wrapper around :py:class:`OpenSSL.SSL.Connection`.
```

docs/changelog-fragments.d/787.misc.rst
webknjaz, line 1: I'm actually not sure if this constitutes a breaking change. SemVer talks about breaking changes in the context of API compatibility and the API seems to have remained intact. WDYT?

cheroot/ssl/pyopenssl.py
julianz-: Oops sorry that slipped through the net.

cheroot/ssl/pyopenssl.py
julianz-: Sure. Using  `__getattr__` was attractive for adding the error translation in one fell swoop (currently destined for #786) but we can still do this explicitly for each method that needs it. That's probably better anyway.

.flake8
julianz-: i removed - these were added to deal with linting issues from earlier iterations.

cheroot/ssl/pyopenssl.py
julianz-: Unlike `makefile()` which I just had to add, the linter is not complaining about this method being missing. We could add for socket compatibility I guess, but it's just going to call the `shutdown` method. If we try to replicate the whole socket interface then wouldn't we need to add a bunch of other methods? Then we end up adding all kinds of cruft?

docs/spelling_wordlist.txt
julianz-: The linter complains about pyOpenSSL but not about PyOpenSSL but the former is actually more precise.

docs/changelog-fragments.d/787.misc.rst
julianz-: If the API offered thread-safety and now suddenly we are saying these methods are not thread-safe that sounds like a breaking change to me, no?  It's somewhat moot I guess if, in practice, no-one needs these methods to be thread-safe but just in case it's better to clear this is a potentially breaking change. If not a breaking change what category would be best?

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,41 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
-        # Doesn't work via super() for some reason.
-        # Falling back to type() instead:
-        return type(name, bases, nmspc)
+    This class exists primarily to ensure the standard Python socket method
+    :py:meth:`.shutdown()` is available for interface compatibility.
+    """
 
+    def makefile(self, *args, **kwargs):
+        """
+        Raise NotImplementedError for unsupported ``makefile()`` interface.
```
```suggestion
        Raise :exc:`NotImplementedError` for unsupported ``makefile()`` interface.
```

cheroot/ssl/pyopenssl.py
```diff
@@ -176,99 +175,41 @@ class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
...
+    A compatibility wrapper around :py:class:`OpenSSL.SSL.Connection`.
 
-        # Doesn't work via super() for some reason.
-        # Falling back to type() instead:
-        return type(name, bases, nmspc)
+    This class exists primarily to ensure the standard Python socket method
+    :py:meth:`.shutdown()` is available for interface compatibility.
+    """
 
+    def makefile(self, *args, **kwargs):
```
webknjaz: Why do we need this? Can we replicate the method signature exactly?

cheroot/ssl/pyopenssl.py
webknjaz: I'm mostly curious about the calling code using it rather than us reimplementing it. This class will get the implementation through inheritance.

@julianz- I just remembered that the calling code actually invokes `sock_shutdown()` and not `shutdown()`. See https://github.com/cherrypy/cheroot/blob/22bb4306737d9fd0b3a988ac6072d31d0de57665/cheroot/server.py#L1543.

I wonder if we could have interface that would be uniform across adapters and wouldn't have that ugly `getattr()` hack, which is effectively an abstraction leak.

docs/changelog-fragments.d/787.misc.rst
webknjaz: Honestly, I'm torn. We could keep it `breaking`. Otherwise, it'd probably be `misc` or something like that, which is not very prominent in the change log.

docs/spelling_wordlist.txt
webknjaz: Oh yeah, this is because you kept it in the docstrings. Alright. Marking this as resolved, then.

cheroot/ssl/pyopenssl.py
julianz-: makefile() I added because the linter complained about makefile being abstract and needing an implementation.

docs/changelog-fragments.d/787.misc.rst
julianz-: Sure.  That makes sense. I'm fine to move to misc.

webknjaz: @julianz- I'm not opposed to keeping it `breaking`. Just thinking out loud. If it's breaking, I'd bump the major version component on release. Otherwise, it might be a minor change.

cheroot/ssl/pyopenssl.py
webknjaz: Which linter? Do you have logs?

cheroot/ssl/pyopenssl.py
julianz-: maybe then we assume all connection adapters (including SSLConnection) implement sock_shutdown?

docs/changelog-fragments.d/787.misc.rst
julianz-: is Breaking Bad?  haha. i'm torn also. i changed to `misc` but if you want to go back to `breaking` let me know.

webknjaz: Let's keep it `misc`, I suppose.

cheroot/ssl/pyopenssl.py
webknjaz: @julianz- I'm pretty sure the abstract method is in the adapter and not in `SSL.Connection`. Are you sure this is needed?

cheroot/ssl/pyopenssl.py
webknjaz: @julianz- yes, that's sort of what I was considering. But looking at https://github.com/cherrypy/cheroot/blob/22bb4306737d9fd0b3a988ac6072d31d0de57665/cheroot/ssl/builtin.py#L313-L317, it seems like we'd have to subclass `socket.socket()`.

This means a separate PR as it feels like a significant effort to figure out. If you're up for it, please try composing a patch scoped to implementing this compatibility interface. It would probably need to come before this PR, though.

cheroot/server.py
```diff
@@ -1537,12 +1537,7 @@ def peer_group(self):
 
     def _close_kernel_socket(self):
         """Terminate the connection at the transport level."""
-        # Honor ``sock_shutdown`` for PyOpenSSL connections.
-        shutdown = getattr(
-            self.socket,
-            'sock_shutdown',
-            self.socket.shutdown,
-        )
+        shutdown = self.socket.sock_shutdown
```
webknjaz, line 1540: In https://github.com/cherrypy/cheroot/pull/787/files#r2498797283 I explained that this would require more work. Specifically because stdlib's `socket.socket` does not have this method. Also, this wouldn't need to go into a variable but called directly.

cheroot/ssl/pyopenssl.py
julianz-: https://github.com/cherrypy/cheroot/actions/runs/19145536408/job/54722201402?pr=787

julianz-: I guess it's not really abstract in SSL.Connection - it's just marked as not implemented. The linter interprets that as abstract?

cheroot/ssl/pyopenssl.py
julianz-: ok will give it a try

cheroot/ssl/pyopenssl.py
webknjaz: Oh, so it's pylint: https://pylint.rtfd.io/en/v4.0.2/user_guide/messages/warning/abstract-method.html.

In that case it makes sense to have this method, of course. @julianz- just include a code comment mentioning what linter+rule is violated so that the justification is recorded and doesn't confuse future readers.

----

Here's extracted console output for history:
```console
************* Module cheroot.ssl.pyopenssl
cheroot/ssl/pyopenssl.py:L 178:(abstract-method) Method 'makefile' is abstract in class 'Connection' but is not overridden in child class 'SSLConnection' (C  0), ::W0223

------------------------------------
Your code has been rated at 10.00/10
```

cheroot/server.py
webknjaz: @julianz- did you have any luck abstracting away the low-level socket class? I've fixed up the linkcheck job in CI so it shouldn't be failing anymore, FTR.

julianz-: I have something I should be able to upload any day now as a new PR. It was a more complicated process to get there than I was expecting so it's taken longer than I was hoping 😱

webknjaz: Yeah, I expected it to get rather involved. Just checking in.

julianz-: @webknjaz I have uploaded a new [branch](https://github.com/julianz-/cheroot/tree/tls_socket_refactor) to my fork. I initially tried creating a unified adapter but was not that satisfied with it so I chose I slightly different way - unifying the SSL sockets used in the different adapters with a new class TLSSocket. It acts as a delegating wrapper around the raw PyOpenSSL or Builtin SSL connection objects. The code change is quite a behemoth, but I think it makes unifying the error handling and various other things much nicer. For PyOpenSSL, it gets rid of SSLFileobjectMixin and SSLConnectionProxyMeta. The get_environ() processing is more unified too with delegation to the adapters where necessary. I can set up a PR but thought I should run it past you first!

webknjaz: > I have uploaded a new [branch](https://github.com/julianz-/cheroot/tree/tls_socket_refactor) to my fork.

You could open a draft PR right away — it's easier to review/follow changes/process this way...

> I initially tried creating a unified adapter but was not that satisfied with it so I chose I slightly different way - unifying the SSL sockets used in the different adapters with a new class TLSSocket. It acts as a delegating wrapper around the raw PyOpenSSL or Builtin SSL connection objects. The code change is quite a behemoth, but I think it makes unifying the error handling and various other things much nicer. For PyOpenSSL, it gets rid of SSLFileobjectMixin and SSLConnectionProxyMeta. The get_environ() processing is more unified too with delegation to the adapters where necessary.

Sounds along the lines of what I envisioned. I'll look deeper once there's a PR to leave comments in.

> I can set up a PR but thought I should run it past you first!

Filing an unfinished PR is cheap enough that I normally recommend making it right away upon topic branch creation, even if it's almost empty. It's actually more convenient to communicate changes that way since you can add inline comments in the diff and stuff.

julianz-: Sounds good. Opened as #799.

PR #791, opened 2025-11-03
Stabilize flaky `test_openssl_adapter_with_false_key_password`

jatalahd:
Sometimes the pyopenssl adapter tests failed with a different error message than expected. This patch adapts the regex to cover the new discovered error message.

Fix to an occasionally failing test: `test_openssl_adapter_with_false_key_password`

* [ ] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [X] 📋 tests/coverage improvement
* [ ] 📋 refactoring
* [ ] 💥 other

This is a follow-up for https://github.com/cherrypy/cheroot/pull/752#discussion_r2485127496

The test fails occasionally with different error message than expected

Extended the expected error message regex to cover the other use case too

📋 **Other information**:

📋 **Contribution checklist:**

* [X] I wrote descriptive pull request text above
* [X] I think the code is well written
* [X] I wrote [good commit messages]
* [X] I used the same coding conventions as the rest of the project
* [X] The new code doesn't generate linter offenses
* [X] The PR relates to *only* one subject with a clear title

[related squash]:

- - -
This change is [Reviewable](https://reviewable.io/reviews/cherrypy/cheroot/791)

### DIFF
--- a/cheroot/test/test_ssl.py
+++ b/cheroot/test/test_ssl.py
@@ -924,7 +924,10 @@ def test_openssl_adapter_with_false_key_password(
 
     with expected_warn, pytest.raises(
         OpenSSL.SSL.Error,
-        match=r'.+bad decrypt.+',
+        # Decode error has happened very rarely with Python 3.9 in MacOS.
+        # Might be caused by a random issue in file handling leading
+        # to interpretation of garbage characters in certificates.
+        match=r'.+\'(bad decrypt|decode error)\'.+',
     ):
         httpserver.prepare()
 


cheroot/test/test_ssl.py
```diff
@@ -924,7 +924,7 @@ def test_openssl_adapter_with_false_key_password(
 
     with expected_warn, pytest.raises(
         OpenSSL.SSL.Error,
-        match=r'.+bad decrypt.+',
+        match=r'.+\'bad decrypt\'.+|.+\'decode error\'.+',
```
webknjaz: @jatalahd thanks, can we collapse these?
```suggestion
        match=r'.+\'(decode error|bad decrypt)\'.+',
```

webknjaz: Also, could you add a code comment right above the `match=` line explaining why both of these are expected? Maybe some observations that would be useful in the future?

webknjaz:
> for me it reproduced in the `[integer-password]` test, so I decided to just generally extend the expected regex error message

Makes sense.

cheroot/test/test_ssl.py
jatalahd: Collapsed and added a comment. Could not reproduce anymore with about 30 test executions, so assuming it is a very rare case.

PR #798, opened 2025-11-16
Fix manual prompt in pyopenssl adapter for private key password

jatalahd:
If `pyopenssl` adapter is used with password protected private key and the manual entry option was not given, only a fail due to invalid password. The password callback used to also be triggered in the case where the `private_key_password` was `None`.

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves #

Manual prompt to enter private key password is not given in server startup
in the case where private_key_password = None. Functionality conflicts with
documentation and with builtin adapter

With this fix the functionality is matching the documentation
and both ssl adapters work in similar fashion.

📋 **Other information**:

📋 **Contribution checklist:**

[related squash]:

### DIFF
--- a/cheroot/ssl/__init__.py
+++ b/cheroot/ssl/__init__.py
@@ -2,6 +2,7 @@
 
 import socket as _socket
 from abc import ABC, abstractmethod
+from getpass import getpass as _ask_for_password_interactively
 from warnings import warn as _warn
 
 from .. import errors as _errors
@@ -113,3 +114,8 @@ class Adapter(ABC):
     def makefile(self, sock, mode='r', bufsize=-1):
         """Return socket file object."""
         raise NotImplementedError  # pragma: no cover
+
+    def _prompt_for_tls_password(self) -> str:
+        """Prompt for encrypted private key password interactively."""
+        prompt = 'Enter PEM pass phrase: '
+        return _ask_for_password_interactively(prompt)
--- a/cheroot/ssl/__init__.pyi
+++ b/cheroot/ssl/__init__.pyi
@@ -1,3 +1,4 @@
+import collections.abc as _c
 import typing as _t
 from abc import ABC, abstractmethod
 
@@ -6,8 +7,9 @@ class Adapter(ABC):
     private_key: _t.Any
     certificate_chain: _t.Any
     ciphers: _t.Any
-    private_key_password: str | bytes | None
+    private_key_password: _c.Callable[[], bytes | str] | bytes | str | None
     context: _t.Any
+
     @abstractmethod
     def __init__(
         self,
@@ -16,7 +18,10 @@ class Adapter(ABC):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: _c.Callable[[], bytes | str]
+        | bytes
+        | str
+        | None = ...,
     ): ...
     def bind(self, sock): ...
     @abstractmethod
@@ -25,3 +30,4 @@ class Adapter(ABC):
     def get_environ(self): ...
     @abstractmethod
     def makefile(self, sock, mode: str = ..., bufsize: int = ...): ...
+    def _prompt_for_tls_password(self) -> str: ...
--- a/cheroot/ssl/builtin.py
+++ b/cheroot/ssl/builtin.py
@@ -247,6 +247,9 @@ class BuiltinSSLAdapter(Adapter):
             private_key_password=private_key_password,
         )
 
+        if private_key_password is None:
+            private_key_password = self._prompt_for_tls_password
+
         self.context = ssl.create_default_context(
             purpose=ssl.Purpose.CLIENT_AUTH,
             cafile=certificate_chain,
--- a/cheroot/ssl/builtin.pyi
+++ b/cheroot/ssl/builtin.pyi
@@ -1,3 +1,4 @@
+import collections.abc as _c
 import typing as _t
 
 from . import Adapter
@@ -14,7 +15,10 @@ class BuiltinSSLAdapter(Adapter):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: _c.Callable[[], bytes | str]
+        | bytes
+        | str
+        | None = ...,
     ) -> None: ...
     @property
     def context(self): ...
--- a/cheroot/ssl/pyopenssl.py
+++ b/cheroot/ssl/pyopenssl.py
@@ -352,11 +352,20 @@ class pyOpenSSLAdapter(Adapter):
     def _password_callback(
         self,
         password_max_length,
-        _verify_twice,
-        password,
+        verify_twice,
+        password_or_callback,
         /,
     ):
         """Pass a passphrase to password protected private key."""
+        if callable(password_or_callback):
+            password = password_or_callback()
+            if verify_twice and password != password_or_callback():
+                raise ValueError(
+                    'Verification failed: entered passwords do not match',
+                ) from None
+        else:
+            password = password_or_callback
+
         b_password = b''  # returning a falsy value communicates an error
         if isinstance(password, str):
             b_password = password.encode('utf-8')
@@ -381,6 +390,8 @@ class pyOpenSSLAdapter(Adapter):
         """
         # See https://code.activestate.com/recipes/442473/
         c = SSL.Context(SSL.SSLv23_METHOD)
+        if self.private_key_password is None:
+            self.private_key_password = self._prompt_for_tls_password
         c.set_passwd_cb(self._password_callback, self.private_key_password)
         c.use_privatekey_file(self.private_key)
         if self.certificate_chain:
--- a/cheroot/ssl/pyopenssl.pyi
+++ b/cheroot/ssl/pyopenssl.pyi
@@ -1,3 +1,4 @@
+import collections.abc as _c
 import typing as _t
 
 from OpenSSL import SSL
@@ -32,14 +33,20 @@ class pyOpenSSLAdapter(Adapter):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: _c.Callable[[], bytes | str]
+        | bytes
+        | str
+        | None = ...,
     ) -> None: ...
     def wrap(self, sock): ...
     def _password_callback(
         self,
         password_max_length: int,
-        _verify_twice: bool,
-        password: bytes | str | None,
+        verify_twice: bool,
+        password_or_callback: _c.Callable[[], bytes | str]
+        | bytes
+        | str
+        | None,
         /,
     ) -> bytes: ...
     def get_environ(self): ...
--- a/cheroot/test/test_ssl.py
+++ b/cheroot/test/test_ssl.py
@@ -149,10 +149,18 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
     return httpserver
 
 
+def get_key_password():
+    """Return a predefined password string.
+
+    It is to be used for decrypting private keys.
+    """
+    return 'криївка'
+
+
 @pytest.fixture(scope='session')
 def private_key_password():
     """Provide hardcoded password for private key."""
-    return 'криївка'
+    return get_key_password()
 
 
 @pytest.fixture
@@ -900,9 +908,17 @@ def test_http_over_https_ssl_handshake(
     ids=('encrypted-key', 'unencrypted-key'),
 )
 @pytest.mark.parametrize(
-    'password_as_bytes',
-    (True, False),
-    ids=('with-bytes-password', 'with-str-password'),
+    'transform_password_arg',
+    (
+        lambda pass_factory: pass_factory().encode('utf-8'),
+        lambda pass_factory: pass_factory(),
+        lambda pass_factory: pass_factory,
+    ),
+    ids=(
+        'with-bytes-password',
+        'with-str-password',
+        'with-callable-password-provider',
+    ),
 )
 # pylint: disable-next=too-many-positional-arguments
 def test_ssl_adapters_with_private_key_password(
@@ -915,7 +931,7 @@ def test_ssl_adapters_with_private_key_password(
     tls_certificate_private_key_pem_path,
     adapter_type,
     encrypted_key,
-    password_as_bytes,
+    transform_password_arg,
 ):
     """Check server decrypts private TLS keys with password as bytes or str."""
     key_file = (
@@ -923,17 +939,13 @@ def test_ssl_adapters_with_private_key_password(
         if encrypted_key
         else tls_certificate_private_key_pem_path
     )
-    key_pass = (
-        private_key_password.encode('utf-8')
-        if password_as_bytes
-        else private_key_password
-    )
+    private_key_password = transform_password_arg(get_key_password)
 
     tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
     tls_adapter = tls_adapter_cls(
         certificate=tls_certificate_chain_pem_path,
         private_key=key_file,
-        private_key_password=key_pass,
+        private_key_password=private_key_password,
     )
 
     interface, _host, port = _get_conn_data(
@@ -1012,6 +1024,76 @@ def test_openssl_adapter_with_false_key_password(
             certificate=tls_certificate_chain_pem_path,
             private_key=tls_certificate_passwd_private_key_pem_path,
             private_key_password=false_password,
+        )
+
+
+@pytest.mark.parametrize(
+    'adapter_type',
+    ('pyopenssl', 'builtin'),
+)
+def test_ssl_adapter_with_none_key_password(
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    private_key_password,
+    adapter_type,
+    mocker,
+):
+    """Check that TLS-adapters prompt for password when set as ``None``."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    mocker.patch(
+        'cheroot.ssl._ask_for_password_interactively',
+        return_value=private_key_password,
+    )
+    tls_adapter = tls_adapter_cls(
+        certificate=tls_certificate_chain_pem_path,
+        private_key=tls_certificate_passwd_private_key_pem_path,
+    )
+
+    assert tls_adapter.context is not None
+
+
+class PasswordCallbackHelper:
+    """Collects helper methods for mocking password callback."""
+
+    def __init__(self, adapter: Adapter):
+        """Initialize helper variables."""
+        self.counter = 0
+        self.callback = adapter._password_callback
+
+    def get_password(self):
+        """Provide correct password on first call, wrong on other calls."""
+        self.counter += 1
+        return get_key_password() * self.counter
+
+    def verify_twice_callback(self, max_length, _verify_twice, userdata):
+        """Establish a mock callback for testing two-factor password prompt."""
+        return self.callback(self, max_length, True, userdata)
+
+
+@pytest.mark.parametrize('adapter_type', ('pyopenssl',))
+def test_openssl_adapter_verify_twice_callback(
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    adapter_type,
+    mocker,
+):
+    """Check that two-time password verification fails with correct error."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    helper = PasswordCallbackHelper(tls_adapter_cls)
+
+    mocker.patch(
+        'cheroot.ssl.pyopenssl.pyOpenSSLAdapter._password_callback',
+        side_effect=helper.verify_twice_callback,
+    )
+
+    with pytest.raises(
+        ValueError,
+        match='Verification failed: entered passwords do not match',
+    ):
+        tls_adapter_cls(
+            certificate=tls_certificate_chain_pem_path,
+            private_key=tls_certificate_passwd_private_key_pem_path,
+            private_key_password=helper.get_password,
         )
 
 
--- a/docs/changelog-fragments.d/798.bugfix.rst
+++ b/docs/changelog-fragments.d/798.bugfix.rst
@@ -1,0 +1,7 @@
+Fixed prompting for the encrypted private key password interactively,
+when the password in not set in the :py:attr:`private_key_password attribute
+<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key_password>` in the
+:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`.
+Also improved the private key password to accept the :py:class:`~collections.abc.Callable` type.
+
+-- by :user:`jatalahd`


cheroot/ssl/pyopenssl.py
```diff
@@ -367,7 +367,8 @@ def get_context(self):
         """
         # See https://code.activestate.com/recipes/442473/
         c = SSL.Context(SSL.SSLv23_METHOD)
-        c.set_passwd_cb(self._password_callback, self.private_key_password)
+        if self.private_key_password is not None:
+            c.set_passwd_cb(self._password_callback, self.private_key_password)
```
webknjaz: Was https://github.com/cherrypy/cheroot/pull/752#discussion_r2433024754 incorrect? Why? How?

Can this be tested? Could you add a change log fragment?

webknjaz: Why don't we have `_password_callback()` `raise RuntimeError` `if password is None`? If we don't set the callback, passing an encrypted private key will probably traceback in some weird place. It seems more reasonable to have our own error reporting for this case.

jatalahd: The comment you are referring to was correct in that context and the code works in that scenario. The current fix is for a scenario where the private key is encrypted, but the password is not provided via `private_key_password` argument. In that case when starting the server, it should ask the user to input the password manually, but in this case, as defined, the `set_passwd_callback()` is called every time it encounters an encrypted private key. If the password is not given as argument it returns the empty string and fails due to password mismatch. That is why the callback should not be called if there is no password given in arguments.

This cannot be tested here in cheroot side, but I am assuming that it could be tested in cherrypy side when starting the server in a subprocess and communicating the user input to the process.

webknjaz: Before we proceed, could you fact-check that it actually works as the doc describes? https://github.com/cherrypy/cherrypy/issues/1583#issuecomment-3573455638.

jatalahd: The password prompt arises from the underlying ssl implementation and is unaffected by cherrypy/cheroot code. The official documentation for the ssl library (used in the builtin adapter) for the method `SSLContext.load_cert_chain()` says (reference: https://docs.python.org/3/library/ssl.html#ssl-contexts)

> If the password argument is not specified and a password is required, OpenSSL’s built-in password prompting mechanism will be used to interactively prompt the user for a password.

As the pyOpenSSL is only a wrapper on top of OpenSSL, it behaves the same way. This can be verified by making an OpenSSL commandline operation on an encrypted private key: the result is `Enter PEM pass phrase:` command line prompt.

I had tested both adapters working this way from cherrypy (and actually using bottle app, along with cheroot to provide the https layer with the adapters) before starting my initial commit. That is why I originally had the `if` in the initial commit, but forgot about its meaning during the long process of finetuning the code through the review process.

webknjaz: Alright. Thanks for clearing that up!
Seeing how this confused me (since I never used it like this myself), I think it's a reasonable approach and it deserves a code comment. Could you add one? We need something that would stop people from trying to refactor this seemingly unnecessary conditional at some point in the future — a sufficient justification explanation. Such breakcrumbs are important — I've been looking into the past two decades of history of this module just the other day and trying to understand people's motivation is challenging sometimes.

jatalahd: Added code comment in latest push.

webknjaz: @jatalahd what if instead we put a `getpass.getpass()` into the callback and have more control over retrieving said password? I thought we'd do this when it was first implemented but forgot to ask then. And now I'm remembering this idea. WDYT?

webknjaz: We could tie this into the idea I suggested @ https://github.com/cherrypy/cherrypy/pull/2074#discussion_r2562151083. Something like this would do:

```python
from getpass import getpass as _ask_for_password_interactively

...

def _prompt_for_tls_password(prompt_prefix: str = '', /) -> str:
    password_prompt = (
        f'{prompt_prefix}Enter the password for deciphering '
        'the TLS private key PEM file: '
    )

    return _ask_for_password_interactively(password_prompt)

...

class pyOpenSSLAdapter(Adapter):

    ...

    def _password_callback(
        self,
        password_max_length,
        verify_twice,
        password_or_callback,
        /,
    ):
        ...
        if callable(password_or_callback):
            password = password_or_callback()
            if verify_twice and password != password_or_callback('Once again. '):
                raise ValueError(
                    'Verification failed: entered passwords do not match',
                ) from None
        else:
            password = password_or_callback
        ...

    ...

    def get_context(self) -> SSL.Context:
        ...
        if self.private_key_password is None:
            self.private_key_password = _prompt_for_tls_password
        c.set_passwd_cb(self._password_callback, self.private_key_password)
        ...
```

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed ssl adapter password prompt when encrypted private key used and password not given via private_key_password argument
```
webknjaz:
```suggestion
Fixed prompting for the encrypted private key password interactively, when the password
in not set in the :py:attr:`private_key_password attribute
<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key>` in the
:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`
```

cheroot/ssl/pyopenssl.py
jatalahd: Yes, good idea, but now you forced me to push the builtin adapter to accept Callable type. So this simple one liner will most likely turn into many weeks of pruning :)
Check out the latest push.

cheroot/ssl/builtin.py
```diff
@@ -247,14 +247,18 @@ def __init__(
...
+
         self.context = ssl.create_default_context(
             purpose=ssl.Purpose.CLIENT_AUTH,
             cafile=certificate_chain,
         )
         self.context.load_cert_chain(
             certificate,
             private_key,
-            password=private_key_password,
+            password=key_password,
```
webknjaz: According to https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain, this may be a callable too. Perhaps, we could reuse the one from the other adapter? We could put it into a `._pem_processing` module, for example use as the default here, providing argument defaults via `functools.partial()`, for example.

cheroot/ssl/__init__.pyi
```diff
@@ -1,12 +1,12 @@
...
-from typing import Any
+from typing import Any, Callable
 
 class Adapter(ABC):
     certificate: Any
     private_key: Any
     certificate_chain: Any
     ciphers: Any
-    private_key_password: str | bytes | None
+    private_key_password: str | bytes | Callable | None
```
webknjaz: Let's be more specific about the callable shape + incorporate the new typing import style I've just added on `main`.
```suggestion
    private_key_password: str | bytes | Callable[[], bytes | str] | None
```

cheroot/ssl/builtin.py
webknjaz: Or maybe not that, the `ssl` stdlib module doesn't need that wrapper, I imagine. We could just pass it as is, I think.

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Fixed prompting for the encrypted private key password interactively,
+when the password in not set in the :py:attr:`private_key_password attribute
+<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key>` in the
+:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`.
+Also improved :py:attr:`private_key_password` to accept Callable type.
+-- by :user:`jatalahd`.
```
webknjaz:
```suggestion

-- by :user:`jatalahd`
```

cheroot/test/test_ssl.py
```diff
@@ -924,6 +933,50 @@ def test_openssl_adapter_with_false_key_password(
...
+    tls_http_server,
+    private_key_password,
+    adapter_type,
+    monkeypatch,
+):
+    """Check that openssl ssl-adapter prompts password when set as None."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    monkeypatch.setattr(
+        pyopenssl,
+        '_prompt_for_tls_password',
```
webknjaz: It's probably best to patch `_ask_for_password_interactively` so that its internal lines get coverage measured.

cheroot/ssl/builtin.py
```diff
@@ -247,14 +247,18 @@ def __init__(
             private_key_password=private_key_password,
         )
 
+        key_password = private_key_password
+        if callable(private_key_password):
+            key_password = private_key_password()
```
webknjaz: Looking into this more, I think that this could be the same as in the pyopenssl adapter. Smth like
```suggestion
        if private_key_password is None:
            private_key_password = _prompt_for_tls_password
```
(provided the helper function is moved into a shared module)

jatalahd: This seems to work and it unifies the adapters perfectly.

cheroot/test/test_ssl.py
jatalahd: Done in latest push. But I think we need to make some cleanup still.

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: I am not very good at getting things right in these fragments. Either the wording is wrong or some checker has something to warn about.

cheroot/ssl/__init__.pyi
jatalahd: Done in latest push.

cheroot/ssl/builtin.py
jatalahd: Seems that we can pass it as is.

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Fixed prompting for the encrypted private key password interactively,
+when the password in not set in the :py:attr:`private_key_password attribute
+<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key>` in the
+:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`.
+Also improved the private key password to accept Callable type.
+-- by :user:`jatalahd`
```
webknjaz, line 7: When a byline is not a part of the sentece, we keep it on a separate line.
```suggestion

-- by :user:`jatalahd`
```

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+Fixed prompting for the encrypted private key password interactively,
+when the password in not set in the :py:attr:`private_key_password attribute
+<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key>` in the
```
webknjaz: Should this be
```suggestion
<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key_password>` in the
```
?

cheroot/ssl/pyopenssl.pyi
```diff
@@ -32,14 +32,20 @@ class pyOpenSSLAdapter(Adapter):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: str
+        | bytes
+        | _t.Callable[[], bytes | str]
+        | None = ...,
```
webknjaz: It may be a good idea to define a new type or a type alias for this (except for the `None` bit I suppose).

cheroot/test/test_ssl.py
```diff
@@ -26,6 +26,7 @@
     load_pem_private_key,
 )
 
+import cheroot.ssl as _cheroot_ssl
```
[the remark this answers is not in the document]
webknjaz: @jatalahd no need to import modules as private in tests. Also, CodeQL is right — we should have a single import. Let's adapt that second one to reuse this.

cheroot/test/test_ssl.py
```diff
@@ -810,8 +816,8 @@ def test_http_over_https_error(
 )
 @pytest.mark.parametrize(
     'password_as_bytes',
-    (True, False),
-    ids=('with-bytes-password', 'with-str-password'),
+    (True, False, None),
```
webknjaz: What if instead of having nested logic, we turn these into transformers?
`operator.call()` is a thing, `str.encode()` and `functools.partial()` are too...

The params could be
```suggestion
    'transform_password_arg',
    (
        lambda pass_factory: pass_factory().encode('utf-8'),
        lambda pass_factory: pass_factory(),
        lambda pass_factory: pass_factory,
    ),
```

And the test method would make use of a fixture that returns a factory instead of calling the function before test:

```python
@pytest.fixture
def retrieve_private_key_password():
    return get_key_password
```

So the test itself would not even need a `key_pass` var but pass things into the adapter initializer directly as `private_key_password=transform_password_arg(retrieve_private_key_password)`.

We could also bypass the fixture and use `private_key_password=transform_password_arg(get_key_password)` directly too.

cheroot/test/test_ssl.py
```diff
@@ -833,9 +839,13 @@ def test_ssl_adapters_with_private_key_password(
...
     key_pass = (
-        private_key_password.encode('utf-8')
+        private_key_password.encode('utf-8')  # use bytes
         if password_as_bytes
-        else private_key_password
+        else (
+            get_key_password  # use callable
+            if password_as_bytes is None
+            else private_key_password  # use str
+        )
```
webknjaz: This can be simpler: https://github.com/cherrypy/cheroot/pull/798/files#r2579194339

cheroot/ssl/__init__.py
```diff
@@ -48,6 +49,11 @@ def bind(self, sock):
         )
         return sock
 
+    def prompt_for_tls_password(self) -> str:
+        """Define interactive prompt for encrypted private key password."""
+        prompt = 'Enter PEM pass phrase:'
```
webknjaz: A whitespace is missing here:
```suggestion
        prompt = 'Enter PEM pass phrase: '
```
Usually, prompts have it there so that the cursor wouldn't look weirdly misplaced.

---

I originally suggested a different signature: https://github.com/cherrypy/cheroot/pull/798#discussion_r2566163502 — did you have a comment on why you decided to implement it differently?

cheroot/ssl/__init__.py
```diff
@@ -48,6 +49,11 @@ def bind(self, sock):
         )
         return sock
 
+    def prompt_for_tls_password(self) -> str:
```
webknjaz: Let's avoid making this a public method for now:
```suggestion
    def _prompt_for_tls_password(self) -> str:
```
Also, another in-progress PR (#800) is putting another helper at the end so this should probably be the last definition too.

cheroot/ssl/pyopenssl.py
```diff
@@ -347,11 +347,20 @@ def wrap(self, sock):
...
-        _verify_twice,
-        password,
+        verify_twice,
+        password_or_callback,
         /,
     ):
         """Pass a passphrase to password protected private key."""
+        if callable(password_or_callback):
+            password = password_or_callback()
+            if verify_twice and password != password_or_callback():
```
webknjaz, line 362: Why did you get rid of the prefix for the second prompt? Seeing two prompts that look the same without any visual feedback will make one think that the first attempt to enter the password was perhaps unsuccessful. This would be confusing UX.

cheroot/ssl/pyopenssl.py
```diff
@@ -376,6 +385,8 @@ def get_context(self):
         """
         # See https://code.activestate.com/recipes/442473/
         c = SSL.Context(SSL.SSLv23_METHOD)
+        if self.private_key_password is None:
+            self.private_key_password = self.prompt_for_tls_password
```
```suggestion
            self.private_key_password = self._prompt_for_tls_password
```

cheroot/ssl/builtin.py
```diff
@@ -247,6 +247,9 @@ def __init__(
             private_key_password=private_key_password,
         )
 
+        if private_key_password is None:
+            private_key_password = self.prompt_for_tls_password
```
webknjaz:
```suggestion
            private_key_password = self._prompt_for_tls_password
```

cheroot/ssl/__init__.pyi
```diff
@@ -19,6 +20,7 @@ class Adapter(ABC):
         private_key_password: str | bytes | None = ...,
     ): ...
     def bind(self, sock): ...
+    def prompt_for_tls_password(self): ...
```
webknjaz:
```suggestion
    def _prompt_for_tls_password(self) -> str: ...
```

cheroot/test/test_ssl.py
```diff
@@ -149,10 +150,15 @@ def make_tls_http_server(bind_addr, ssl_adapter, request):
     return httpserver
 
 
+def get_key_password():
+    """Provide hardcoded password for private key as callable."""
```
webknjaz: This is a bit misleading because this function does not actually return a callable. It returns a string.
```suggestion
    """Return a predefined password string.

    It is to be used for decrypting private keys.
    """
```

cheroot/test/test_ssl.py
```diff
@@ -810,8 +816,8 @@ def test_http_over_https_error(
 )
 @pytest.mark.parametrize(
     'password_as_bytes',
-    (True, False),
-    ids=('with-bytes-password', 'with-str-password'),
+    (True, False, None),
+    ids=('with-bytes-password', 'with-str-password', 'with-callable-password'),
```
```suggestion
    ids=('with-bytes-password', 'with-str-password', 'with-callable-password-provider'),
```

cheroot/test/test_ssl.py
```diff
@@ -924,6 +934,50 @@ def test_openssl_adapter_with_false_key_password(
...
+    http_request_timeout,
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    tls_ca_certificate_pem_path,
+    tls_http_server,
+    private_key_password,
+    adapter_type,
+    monkeypatch,
+):
+    """Check that openssl ssl-adapter prompts password when set as None."""
```
```suggestion
    """Check that TLS-adapters prompt for password when set as ``None``."""
```

cheroot/test/test_ssl.py
```diff
@@ -924,6 +934,50 @@ def test_openssl_adapter_with_false_key_password(
...
+    )
+
+    resp = requests.get(
+        f'https://{interface!s}:{port!s}/',
+        timeout=http_request_timeout,
+        verify=tls_ca_certificate_pem_path,
+    )
+
+    assert resp.status_code == 200
+    assert resp.text == 'Hello world!'
```
webknjaz: Can we check `tls_adapter.context` to verify the key is loaded instead of going through HTTP e2e? This would make the test faster.

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,6 @@
...
+:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`.
+Also improved the private key password to accept Callable type.
```
webknjaz:
```suggestion
Also improved the private key password to accept the :py:class:`~typing.Callable` type.
```

docs/changelog-fragments.d/798.bugfix.rst
webknjaz: This can only be linted to a certain extent. It'd need be something more involved than a regex-based check. The idea is that when a byline is a part of a sentence, it comes at the end and there's no period before the em-dash — there's one full stop at the end of the sentence.
But if it's a standalone byline, it should be visually separated (hence an empty line so that RST wouldn't attach it to the paragraph) and the trailing period is not needed because it's not a sentence.

You can actually preview your own note by either building the docs locally (`tox r -e build-docs -qq`) or checking out the RTD build for this PR. It's here: https://cheroot--798.org.readthedocs.build/en/798/history/#change-log.
The last build failed because I messed something up on `main` but that's fixed now. Once you rebase, it'll build the latest preview again.

Lurking into that page should give you some sense of the writing style of the change log in general and you'd see if it fits well among the surrounding change log entries written in the past.

webknjaz:
@jatalahd you'll need fix a conflict when rebasing.

cheroot/test/test_ssl.py
jatalahd: Yes, well the test should fail already when creating the adapter, so asserting `tls_adapter.context is not None` should be enough. Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/test/test_ssl.py
jatalahd: Done in latest push

cheroot/ssl/__init__.pyi
jatalahd: Done in latest push.

cheroot/ssl/builtin.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: My reasoning was that if someone provides their own Callable to `private_key_password` and (for some reason) the callback decides to verify twice, they need to understand that the Callable they are using needs to accept an optional string argument, which is useless in the case where the `private_key_password` Callable directly returns the password.

cheroot/ssl/__init__.py
jatalahd: Done, but I guess this still needs some work.

cheroot/ssl/__init__.py
jatalahd: I did not comment on that previously. I just don't see the reason to use anything else than the default prompt the internal ssl libraries are using. If you really want it to be different, we can still change the prompt.

cheroot/test/test_ssl.py
jatalahd: Great! Done in the latest push.

cheroot/test/test_ssl.py
jatalahd: This is good. Done in the latest push.

cheroot/ssl/pyopenssl.pyi
jatalahd: I tried a few ways to define the alias, but the linter and/or the MyPy stubtest never accepted my trials. So this is still an open issue.

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: Done in the latest push. I assume this is OK now...

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: Done in the latest push.

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: Done in the latest push.

cheroot/ssl/pyopenssl.pyi
webknjaz: Okay, disregard this, then. We can revisit when the annotations move from stubs to native syntax.

webknjaz: Okay, disregard this, then. We can revisit when the annotations move from stubs to native syntax.

cheroot/ssl/__init__.py
webknjaz: No, it's okay. I just wanted to make sure it's not being missed.

webknjaz: No, it's okay. I just wanted to make sure it's not being missed.

cheroot/ssl/pyopenssl.py
webknjaz: Alright, makes sense to me. Resolved.

cheroot/test/test_ssl.py
```diff
@@ -1015,6 +1027,32 @@ def test_openssl_adapter_with_false_key_password(
...
+def test_openssl_adapter_with_none_key_password(
+    tls_certificate_chain_pem_path,
+    tls_certificate_passwd_private_key_pem_path,
+    private_key_password,
+    adapter_type,
+    monkeypatch,
+):
+    """Check that TLS-adapters prompt for password when set as ``None``."""
+    tls_adapter_cls = get_ssl_adapter_class(name=adapter_type)
+    monkeypatch.setattr(
```
webknjaz: In this instance, `pytest-mock` is probably more elegant:
```suggestion
    mocker.patch(
        'cheroot.ssl._ask_for_password_interactively',
        side_effect=private_key_password,
    )
```

cheroot/ssl/pyopenssl.py
```diff
@@ -352,11 +352,20 @@ def wrap(self, sock):
...
-        password,
+        verify_twice,
+        password_or_callback,
         /,
     ):
         """Pass a passphrase to password protected private key."""
+        if callable(password_or_callback):
+            password = password_or_callback()
+            if verify_twice and password != password_or_callback():
+                raise ValueError(
```
webknjaz, line 363: @jatalahd could you add a test to cover this line?

https://app.codecov.io/gh/cherrypy/cheroot/commit/529dd97b88d902fe15931e79fd8fd93012f1fd77/blob/cheroot/ssl/pyopenssl.py?flags%5B0%5D=pytest&dropdown=coverage#L363

webknjaz (commented):
Just two small test improvements left.

cheroot/test/test_ssl.py
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.py
jatalahd: Added a test, but you need to check what you think about it. Had some hard time mocking the callback(s). Finally got a relatively clean solution, but possibly it can be improved using some advanced Python magic.

webknjaz: I think that's enough. Thanks!

docs/changelog-fragments.d/798.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Fixed prompting for the encrypted private key password interactively,
+when the password in not set in the :py:attr:`private_key_password attribute
+<cheroot.ssl.pyopenssl.pyOpenSSLAdapter.private_key_password>` in the
+:py:class:`pyOpenSSL TLS adapter <cheroot.ssl.pyopenssl.pyOpenSSLAdapter>`.
+Also improved the private key password to accept the :py:class:`~typing.Callable` type.
```
webknjaz: Just realized that `typing.Callable` is deprecated and moved to a new location:
```suggestion
Also improved the private key password to accept the
:py:class:`~collections.abc.Callable` type.
```

docs/changelog-fragments.d/798.bugfix.rst
webknjaz, line 1: Looks good (short of the callable link): https://cheroot--798.org.readthedocs.build/en/798/history/#to-be-included-in-the-next-release

cheroot/ssl/__init__.py
```diff
@@ -113,3 +114,8 @@ def get_environ(self):
     def makefile(self, sock, mode='r', bufsize=-1):
         """Return socket file object."""
         raise NotImplementedError  # pragma: no cover
+
+    def _prompt_for_tls_password(self) -> str:
+        """Define interactive prompt for encrypted private key password."""
```
webknjaz: (because this does actual prompting)
```suggestion
        """Prompt for encrypted private key password interactively."""
```

cheroot/ssl/builtin.pyi
```diff
@@ -14,7 +14,10 @@ class BuiltinSSLAdapter(Adapter):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: _t.Callable[[], bytes | str]
```
webknjaz: Add `import collections.abc as _c` at the top and use it instead of the deprecated type:
```suggestion
        private_key_password: _c.Callable[[], bytes | str]
```

cheroot/ssl/pyopenssl.pyi
```diff
@@ -32,14 +32,20 @@ class pyOpenSSLAdapter(Adapter):
         certificate_chain: _t.Any | None = ...,
         ciphers: _t.Any | None = ...,
         *,
-        private_key_password: str | bytes | None = ...,
+        private_key_password: _t.Callable[[], bytes | str]
```
webknjaz: Add `import collections.abc as _c` at the top and use it instead of the deprecated type:
```suggestion
        private_key_password: _c.Callable[[], bytes | str]
```

cheroot/ssl/pyopenssl.pyi
```diff
@@ -32,14 +32,20 @@ class pyOpenSSLAdapter(Adapter):
...
+        | None = ...,
     ) -> None: ...
     def wrap(self, sock): ...
     def _password_callback(
         self,
         password_max_length: int,
-        _verify_twice: bool,
-        password: bytes | str | None,
+        verify_twice: bool,
+        password_or_callback: _t.Callable[[], bytes | str]
```
webknjaz: Add `import collections.abc as _c` at the top and use it instead of the deprecated type:
```suggestion
        password_or_callback: _c.Callable[[], bytes | str]
```

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: Done in latest push.

cheroot/ssl/__init__.py
jatalahd: Done in latest push.

cheroot/ssl/builtin.pyi
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.pyi
jatalahd: Done in latest push.

cheroot/ssl/pyopenssl.pyi
jatalahd: Done in latest push.

docs/changelog-fragments.d/798.bugfix.rst
jatalahd: Hopefully ok in the latest push.

webknjaz: Yep, it's now clickable!

webknjaz (approved):
Perfect, thanks!

cheroot/test/test_ssl.py
```diff
@@ -900,9 +908,17 @@ def test_http_over_https_ssl_handshake(
     ids=('encrypted-key', 'unencrypted-key'),
 )
 @pytest.mark.parametrize(
-    'password_as_bytes',
-    (True, False),
-    ids=('with-bytes-password', 'with-str-password'),
+    'transform_password_arg',
+    (
+        lambda pass_factory: pass_factory().encode('utf-8'),
+        lambda pass_factory: pass_factory(),
```
webknjaz, line 914: @jatalahd so I replaced the deprecated `macos-13` with `macos-15-intel` on `main` (https://github.com/cherrypy/cheroot/commit/8b8d6f84291bcf8f2527bbd390d9f7a561a491bd) and then merged your PR right after.

The CI now fails on that specific VM under PyPy 3.11 with a SEGFAULT in `cheroot/test/test_ssl.py::test_ssl_adapters_with_private_key_password[with-str-password-unencrypted-key-builtin] ` which this PR touches (but another one introduced originally).

The last successful PR CI run wasn't running pypy under macos: https://github.com/cherrypy/cheroot/actions/runs/20009127282. I just forgot to skip it when changing VMs.

This means that the pathc in this PR isn't necessarily related to the problem but if you have a minute, it might be a good idea to look closer in case this is something that might start happening under classic CPython too...

webknjaz: Although.. The surrounding comments suggest pre-existing issues with `pytest-xdist` + PyPy. So never mind that.

webknjaz: With https://github.com/cherrypy/cheroot/commit/41a44f194b3374e6d57675b3b8a467fda880675d, the CI should be fine again.

jatalahd: Ok, hopefully it is fine now. I did not see any test issues when running the unit tests locally with Python 3.11 in MacOS15.

webknjaz: Yeah, this was just PyPy, not CPython.

webknjaz: Yeah, this was just PyPy, not CPython.

PR #799, opened 2025-11-20
Add `TLSSocket` abstraction for uniform SSL/TLS handling

julianz-:
Introduces a new `TLSSocket` class to act as a unified wrapper for SSL/TLS connections, regardless of the underlying adapter (`builtin`, `pyOpenSSL`).

This refactoring aims to:
1. Simplify adapter logic by centralizing common TLS socket properties and methods.
2. Improve consistency when populating WSGI environment variables.
3. Centralize error handling for the adapters.


cheroot/connections.py
```diff
@@ -295,8 +294,17 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
...
             mf = MakeFile
             ssl_env = {}
+
             # if ssl cert and key are set, we try to be a secure HTTP server
             if self.server.ssl_adapter is not None:
+                # 1. Preemptive check for BuiltinSSLAdapter
+                if hasattr(
+                    self.server.ssl_adapter,
+                    '_check_for_plain_http',
+                ) and self.server.ssl_adapter._check_for_plain_http(s):
```
webknjaz: Better move this into the wrap method of the adapter and raise a `NoSSLError` so that it's handled in the same `except` block below for both adapters.
This way, the adapter internals won't leak into this layer.

cheroot/connections.py
```diff
@@ -381,6 +365,38 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
...
+            'Content-Type: text/plain\r\n',
+            f'Content-Length: {len(msg)}\r\n',
+            'Connection: close\r\n',
+            '\r\n',
+            msg,
+        ]
+        response_bytes = ''.join(response_parts).encode('ISO-8859-1')
+
+        try:
+            sock.sendall(response_bytes)
```
webknjaz: Do we need a modern enough PyOpenSSL for `sendall()` to work well?

cheroot/connections.py
```diff
@@ -306,35 +314,11 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
...
-                        if ex.args[0] not in errors.socket_errors_to_ignore:
-                            raise
-                    return None
-                mf = self.server.ssl_adapter.makefile
-                # Re-apply our timeout since we may have a new socket object
-                if hasattr(s, 'settimeout'):
-                    s.settimeout(self.server.timeout)
+                except errors.NoSSLError:
+                    # this only arises for PyOpenSSL as
+                    # we handle http over https pre-emptivley
```
webknjaz:
```suggestion
                    # we handle http over https pre-emptively
```
(although, with the suggestion above we could just drop the entire comment)

cheroot/connections.py
```diff
@@ -381,6 +365,38 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
...
+
+        try:
+            sock.sendall(response_bytes)
+            sock.shutdown(socket.SHUT_WR)
+        except OSError as ex:
+            if ex.args[0] not in errors.socket_errors_to_ignore:
+                raise
+
+        sock.close()
+
```
webknjaz, line 390: This method (or a simplified/standalone variant) could go into a separate PR so that we'd accept it earlier and have a smaller patch here.

cheroot/makefile.py
```diff
@@ -4,6 +4,8 @@
 import _pyio as io
 import socket
 
+from .ssl.tls_socket import TLSSocket
```
webknjaz: I'm not sure we should be bringing TLS internals into the generic makefile. I'll have to think more about this..

cheroot/server.py
```diff
@@ -1522,7 +1522,7 @@ def _close_kernel_socket(self):
                 raise
 
 
-class HTTPServer:
+class HTTPServer:  # noqa: PLR0904
```
webknjaz: Why?

cheroot/server.py
```diff
@@ -1590,14 +1590,26 @@ class HTTPServer:
...
     """
 
+    @property
+    def ssl_adapter(self):
+        """An instance of ssl.Adapter (or a subclass)."""
+        return self._ssl_adapter
+
+    @ssl_adapter.setter
+    def ssl_adapter(self, value):
+        """Set the SSL adapter and establish bidirectional reference."""
```
webknjaz: I'd rather not mutate the attributes post object init. Will have to think about this more.

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,219 @@
-"""Implementation of the SSL adapter base interface."""
+"""Implementation of the SSL adapter base interface.
 
+.. spelling::
+
+   dn
```
webknjaz, line 4: This seems to be not necessary anymore.

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,219 @@
...
+            'wsgi.run_once': False,
+            'CHERPY_SERVER': server.version,
+            'PYTHON_VERSION': sys.version.split()[0],
+            'OS_VERSION': os.name,
+        }
+
+        # Add Cheroot-specific variables
+        env.update(
+            {
+                'CHERPY_SERVER': server.version,
```
webknjaz: Typo? Also why not cheroot?

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,219 @@
...
+        )
+
+        return env
+
+
+class Adapter(SSLEnvironMixin, ABC):
     """Base class for SSL driver library adapters.
 
     Required methods:
-
```
webknjaz, line 10: This will probably break RST rendering.

cheroot/makefile.py
```diff
@@ -59,7 +68,14 @@ class StreamWriter(BufferedWriter):
 
     def __init__(self, sock, mode='w', bufsize=io.DEFAULT_BUFFER_SIZE):
         """Initialize socket stream writer."""
-        super().__init__(socket.SocketIO(sock, mode), bufsize)
+        # If sock is TLSSocket, use it directly as raw I/O
+        if isinstance(sock, TLSSocket):
+            raw_io = sock
+        else:
+            # Standard socket - wrap with SocketIO
+            raw_io = socket.SocketIO(sock, mode)
```
webknjaz: I think this could go into an additional constructor (`@classmethod`) that would perform initialization of socketio outside init and would pass it into the initializer. And the initializer would just use what's passed, no branching. This way, regular where socket creation happens, we'd change the calls to `StreamWriter.from_raw_socket(sock)` and TLS adapters would call `StreamWriter(wrapped_sock)`. No leak of TLS things into this module.

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
-        # strip off any keys by only taking the first certificate
-        cert_start = cert.find(ssl.PEM_HEADER)
-        if cert_start == -1:
-            return
-        cert_end = cert.find(ssl.PEM_FOOTER, cert_start)
-        if cert_end == -1:
-            return
-        cert_end += len(ssl.PEM_FOOTER)
-        self._server_env['SSL_SERVER_CERT'] = cert[cert_start:cert_end]
+        _ = self.context
```
webknjaz: This looks like something unobvious.

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
+
+    def bind(self, sock):
+        """Prepare the server socket."""
+        # Ensure context is initialized
+        _ = self.context
```
webknjaz: ?

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
+
+        Args:
+            sock: Raw socket to wrap with TLS
+
+        Returns:
+            tuple: (TLSSocket, ssl_environ_dict)
+        """
+        try:
+            tls_socket = self._wrap_with_builtin(sock)
+        except Exception as e:
```
webknjaz: We should only handle known/expected errors on this layer.

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
+                ssl_socket=wrapped_ssl_socket,
+                raw_socket=underlying_socket,
+                context=self.context,
+            )
+        except errors.NoSSLError:
+            # For plain HTTP errors, don't close the socket immediately
+            # The response has been sent, let it propagate naturally
+            raise
+        except Exception:
+            with suppress(Exception):
```
webknjaz: This would be a disaster to troubleshoot.

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
+                    ) from e
+                except ssl.SSLError as e:
+                    self._handle_ssl_error(e)
+                except OSError as e:
+                    raise errors.FatalSSLAlert(
+                        f'TCP error during handshake: {e}',
+                    ) from e
+        finally:
+            # Restore original timeout
+            with suppress(Exception):
```
webknjaz, line 286: This could probably just be an oserror, right?

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
+        ):
+            raise errors.NoSSLError(
+                'Client sent plain HTTP request',
+            ) from error
+
+        raise errors.FatalSSLAlert(
+            f'Fatal SSL error during handshake: {error}',
+        ) from error
 
     @property
```
webknjaz, line 327: This could probably be a cached properly.

cheroot/ssl/builtin.py
```diff
@@ -235,174 +121,407 @@ def __init__(
...
 
-    def wrap(self, sock):
-        """Wrap and return the given socket, plus WSGI environ entries."""
         try:
-            s = self.context.wrap_socket(
-                sock,
-                do_handshake_on_connect=True,
-                server_side=True,
+            with open(self.certificate, 'rb') as cert_file:
+                cert_data = cert_file.read()
```
webknjaz, line 424: We should just use `pathlib` for these.

cheroot/ssl/builtin.py
```diff
@@ -482,8 +601,10 @@ def _make_env_dn_dict(self, env_prefix, cert_value):
                 dn_attrs.setdefault(attr_code, [])
                 dn_attrs[attr_code].append(val)
 
+        dn_string = '/'.join(dn)
+
         env = {
-            env_prefix: ','.join(dn),
+            env_prefix: '/%s' % (dn_string,),
```
webknjaz, line 584: Use f-strings instead of old-style interpolation.

cheroot/ssl/builtin.pyi
```diff
@@ -22,5 +22,5 @@ class BuiltinSSLAdapter(Adapter):
     def context(self, context) -> None: ...
     def bind(self, sock): ...
     def wrap(self, sock): ...
-    def get_environ(self, sock): ...
+    def get_environ(self, conn) -> dict: ...
```
webknjaz, line 23: This public API change might be dangerous for CherryPy. Will have to check if it uses this.

cheroot/ssl/tls_socket.py
```diff
@@ -0,0 +1,492 @@
+"""
+A unified SSL/TLS socket layer for Cheroot.
+
+This module provides a TLSSocket class that abstracts over
+different SSL/TLS implementations, such as Python's built-in ssl module
+and pyOpenSSL. It offers a consistent interface for the rest of
```
webknjaz, line 6: This module should only have a generic base class and sockets for different libs would be subclasses. Trying to put multiple corner cases from multiple libs is going to be difficult to maintain or extend.

webknjaz (commented):
These are a few superficial observations. We'll have to work towards moving as much as possible into multiple small PRs so that this one will be reviewable. But one thing I'd want to request already is avoiding intercepting broad exception class and sticking to what's known and expected. I marked some things that are mostly ready for extraction out of this patch.

cheroot/connections.py
julianz-: Much better yes!

cheroot/connections.py
julianz-: I changed to:

```python
if hasattr(sock, 'sendall'):
    sock.sendall(response_bytes)
else:
    # Fallback for older PyOpenSSL or SSL objects
    sock.send(response_bytes)
```

Is that safe enough?

cheroot/connections.py
julianz-: Yes gone now

cheroot/connections.py
julianz-: Sure. Will do.

cheroot/makefile.py
julianz-: Good point. Maybe this is better?

```
# Check for wrapped socket
if hasattr(sock, 'read') and hasattr(sock, 'write'):
    raw_io = sock  # Already wrapped (could be TLSSocket) 
else:
    # Standard socket - wrap with SocketIO
    raw_io = socket.SocketIO(sock, mode)
```

cheroot/server.py
julianz-: You're right. I thought I needed the reference to the server on the adapter so that I could add server variables in get_environ() somehow overlooking the server was already doing this. Totally unnecessary and removed.

cheroot/ssl/__init__.py
julianz-: When I remove I get a linter spell error I assume because of the method: _parse_dn_components()?

../cheroot/ssl/__init__.py:docstring of cheroot.ssl._parse_dn_components:6: : Spell check: dn: ["db", "dm", "d", "n", "den", "din", "don", "dun", "de", "en", "di", "in", "an", "dr", "tn", "dev", "py", "fd"]: dn_type: 'S' for subject or 'I' for issuer.

https://github.com/cherrypy/cheroot/actions/runs/19599932399/job/56130008017?pr=799

cheroot/ssl/__init__.py
julianz-: Gone.

cheroot/ssl/__init__.py
julianz-: Yes it did. Thanks for spotting!

cheroot/makefile.py
webknjaz: With the `@classmethod` suggestion (https://github.com/cherrypy/cheroot/pull/799#discussion_r2548230146), this will be redundant.

webknjaz: With the `@classmethod` suggestion (https://github.com/cherrypy/cheroot/pull/799#discussion_r2548230146), this will be redundant.

cheroot/makefile.py
julianz-: I have implemented something approximating this idea but now the if statement moves up to MakeFile. I guess we could have a different MakeFile call for the wrapped socket but then we would need to modify HTTPConnection and then in turn the call _from_server_socket() in connections.py? 

By checking for an instance of io.RawIOBase we can at least avoid MakeFile needing to know about TLSSocket. But maybe you had something else in mind?

```python
import io as stdlib_io

def MakeFile(sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
    """File object attached to a raw socket object."""
    cls = StreamReader if 'r' in mode else StreamWriter

    # If it's already a file-like object (TLSSocket, etc), use directly
    if isinstance(sock, (io.RawIOBase, stdlib_io.RawIOBase)):
        return cls(sock, bufsize)
    # Otherwise wrap it
    return cls.from_raw_socket(sock, mode, bufsize)
```

cheroot/ssl/builtin.py
julianz-: Removed. Was meant to be something like the lazy initialization method on the pyopenSSL adapter but I see we aren't supporting that here.

cheroot/ssl/builtin.py
julianz-: Ditto

cheroot/ssl/builtin.py
julianz-: Replaced.

cheroot/ssl/builtin.py
julianz-: Gone

cheroot/makefile.py
webknjaz: Oh, I thought we'd get rid of the makefile function for good and initialize the objects directly. The built-in adapter already does that. I vaguely recall struggling with removing it from other places and postponing.

webknjaz: Oh, I thought we'd get rid of the makefile function for good and initialize the objects directly. The built-in adapter already does that. I vaguely recall struggling with removing it from other places and postponing.

julianz-: Ah okay. I have removed `makefile` methods so now the `StreamReader` and `StreamWriter` classes are called directly although they still live in a file called `makefile.py`. Maybe we should rename this to `streams.py` or `buffering.py`?

Also I have, for now, put `if` statements in the init calls to decide whether they are receiving wrapped sockets. But did you to push that logic higher up making changes in `server.py` and `connections.py`?

cheroot/connections.py
webknjaz: I don't know. The fallback might need a loop or something. I was just mostly asking because you worked with that lib recently and I didn't. If it needs a newer PyOpenSSL, I'd maybe think of starting to provide a `pyopenssl` extra in the core packaging metadata.

cheroot/test/conftest.py
```diff
@@ -107,3 +118,90 @@ def make_http_server(bind_addr):
...
+
+@pytest.fixture
+def tls_certificate_private_key_pem_path(tls_certificate):
+    """Provide a certificate private key PEM file path via fixture."""
+    with tls_certificate.private_key_pem.tempfile() as cert_key_pem:
+        yield cert_key_pem
+
+
+@pytest.fixture
+def tls_certificate_passwd_private_key_pem_path(
```
webknjaz, line 183: I think this is only used in one test module. Unless it's used in more places, it should remain in that module.

cheroot/test/conftest.py
webknjaz, line 1: Let's put moving these into a separate PR.

cheroot/test/ssl/test_ssl_builtin.py
webknjaz, line 1: Rename to `cheroot/test/ssl/test_builtin.py`

cheroot/test/ssl/test_ssl_pyopenssl.py
webknjaz, line 1: Rename to `cheroot/test/ssl/test_ssl_pyopenssl.py`

cheroot/test/test_ssl.py
```diff
@@ -32,9 +32,7 @@
     IS_LINUX,
     IS_MACOS,
     IS_PYPY,
-    IS_SOLARIS,
```
webknjaz, line 35: If this var isn't used elsewhere, we might want to drop it from compat.

cheroot/makefile.py
webknjaz: Interesting.. I didn't expect you to remove the methods. Just the function.

I wouldn't look into renaming modules as that might be seen as public API. For now, I'd avoid making a lot of entangled changes since they'll all have to go through separate PRs eventually since it's difficult to make sense of such large diffs. We can return to this idea later.

Let's keep the `makefile()` methods on adapters since this may affect CherryPy. I think up to this point, the adapters were expected to maintain a socket-like interface (hence [`makefile()`](https://docs.python.org/3/library/socket.html#socket.socket.makefile)). I cannot foresee what removing it would affect. Looks like there's one test in CherryPy that might be hitting the removed methods: https://github.com/cherrypy/cherrypy/blob/1f75bc9eed8e0e385f64f368bd69f58d96fb8c2b/cherrypy/test/test_conn.py#L447. So let's postpone dropping it for good.

webknjaz: So for now, let's stick with the `@classmethod` changes and work towards making the diff smaller.

.mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-python_version = 3.8
+python_version = 3.9
```
webknjaz, line 2: This would have to go into a separate PR and be coupled with changes to the core packaging metadata, linting configuration, dev env and CI setup.

cheroot/makefile.py
```diff
@@ -2,6 +2,7 @@
 
 # prefer slower Python-based io module
 import _pyio as io
+import io as stdlib_io
```
webknjaz, line 5: Why is this needed?

cheroot/makefile.py
```diff
@@ -38,9 +39,16 @@ def _flush_unlocked(self):
...
+    def __init__(self, sock, bufsize=io.DEFAULT_BUFFER_SIZE):
+        """Initialize with socket or raw IO object."""
+        # If already a RawIOBase (like TLSSocket), use directly
+        if isinstance(sock, (io.RawIOBase, stdlib_io.RawIOBase)):
+            raw_io = sock
+        else:
+            # Wrap raw socket with SocketIO
+            raw_io = socket.SocketIO(sock, 'rb')
+
+        super().__init__(raw_io, bufsize)
```
webknjaz, line 51: This is still checking obscure properties of objects to infer TLS relation. We shouldn't be relying on whether something is TLS or not in here. This is an abstraction leak no matter the method of checking.
The callers should decide if something needs wrapping and this is best done through providing alternative constructors.

cheroot/makefile.py
```diff
@@ -57,19 +65,20 @@ def has_data(self):
 class StreamWriter(BufferedWriter):
     """Socket stream writer."""
 
-    def __init__(self, sock, mode='w', bufsize=io.DEFAULT_BUFFER_SIZE):
-        """Initialize socket stream writer."""
-        super().__init__(socket.SocketIO(sock, mode), bufsize)
+    def __init__(self, sock, bufsize=io.DEFAULT_BUFFER_SIZE):
```
webknjaz, line 68: Removing arguments or changing their order is a backwards incompatible change. We should avoid those and at least have a deprecation period to allow for downstream adoption. Deprecations would have to be dedicated multi-stage coordinated processes.

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,162 @@
...
+    """
+    Parse Distinguished Name components into environ dict.
+
+    Args:
+        components: Iterable of (key, value) tuples
+        key_prefix: 'SSL_CLIENT' or 'SSL_SERVER'
+        dn_type: 'S' for subject or 'I' for issuer
+
+    Returns:
+        dict: ``DN`` and ``CN`` environment variables
```
webknjaz, line 25: Don't use napoleon that is not integrated. Use Sphinx-native param lists.

```suggestion
    :param components: Iterable of (key, value) tuples
    :param key_prefix: 'SSL_CLIENT' or 'SSL_SERVER'
    :param dn_type: 'S' for subject or 'I' for issuer

    :returns: ``DN`` and ``CN`` environment variables
    :rtype: dict
```

Currently, it's not rendered well and Sphinx objects aren't recorded for params: https://cheroot--799.org.readthedocs.build/en/799/pkg/cheroot.ssl/#cheroot.ssl._parse_dn_components.

When done correctly, the params are clearly marked and well-formatted. As well as return values and raised exceptions. Example: https://cheroot--799.org.readthedocs.build/en/799/pkg/cheroot._compat/#cheroot._compat.extract_bytes.

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,162 @@
...
+from contextlib import suppress
+
+
+DN_SEPARATOR = '/'  # Distinguished Name separator
+UTF8_ENCODING = 'utf-8'
+
+
+def _parse_dn_components(components, key_prefix, dn_type):
+    """
+    Parse Distinguished Name components into environ dict.
```
webknjaz, line 17: Use full words in docstrings. Not "environ" or "dict", because docstrings are user-facing. Additionally, it's possible to link objects Sphinx knows of, like `` :class:`dict` ``.

Within the scope of this function, it doesn't really matter how the caller will use the dict. So claiming that it's an "environ dict" doesn't make sense. It's just a dict and that's it. The caller might put it into a context.

Perhaps, this would be better:
```suggestion
    Transform Distinguished Name components list into a dictionary.
```

cheroot/server.py
```diff
@@ -1275,19 +1275,18 @@ class HTTPConnection:
     # Fields set by ConnectionManager.
     last_used = None
 
-    def __init__(self, server, sock, makefile=MakeFile):
```
webknjaz, line 1278: We'll probably have to keep this for a while (with a sentinel default). And issue a warning when the value isn't that sentinel.

cheroot/server.py
```diff
@@ -1363,7 +1362,7 @@ def _handle_no_ssl(self, req):
         except AttributeError:
             # self.socket is of OpenSSL.SSL.Connection type
```
webknjaz, line 1363: Oh, look — another abstraction leak to fight..

cheroot/ssl/__init__.py
```diff
@@ -1,16 +1,162 @@
...
+        issuer = cert.issuer
+        components = [(attr.oid._name, attr.value) for attr in issuer]
+        env.update(_parse_dn_components(components, key_prefix, 'I'))
+
+    # Version and Serial
+    with suppress(Exception):
+        env[f'{key_prefix}_M_VERSION'] = str(cert.version.value)
+        env[f'{key_prefix}_M_SERIAL'] = str(cert.serial_number)
+
+    return env
```
webknjaz, line 94: These helpers look like they could be a separate refactoring PR. Also, I'm not sure if it's a good idea to put adapter-specific things into a shared module.

webknjaz:
@julianz- let's proceed here by splitting out the makefile function removal into a separate PR. And another PR for the certificate parsing helpers. This should help use make this one as small as possible. Plus cleaning up unrelated config changes will be helpful too.

julianz-:
After checking, removing the makefile really depends on using a unified class to represent the adpaters - i.e. the one I added here: `TLSSocket`. At least that's the way I was able to remove makefile in this PR. Not sure whether that is what you had in mind? I can probably split out the getenviron() stuff.

webknjaz:
Having looked into it, I see that it's not as simple to get rid of `makefile`. I don't have a vision of how to handle that properly but at least "get environ" doesn't seem too complicated so feel free to focus on it first.

Once that's done, we you can rebase this and we'd strategize further. Sounds good?

PR #800, opened 2025-11-23
Unify handling of http over https errors

julianz-:
Added an early probe to identify when a client attempts to speak unencrypted HTTP on an HTTPS port.

When a plaintext connection is detected, the server now attempts to use the raw TCP socket (if wrapped) to send a "400 Bad Request" error response before any TLS failure. This unifies the behavior of the 'builtin' and 'pyopenssl' adapters for this specific error condition.

PR #801, opened 2025-11-25
Align SSL adapters by removing `bind()`

julianz-:
Relocated `SSLConnection` instantiation from `bind()` to `wrap()` to provide a uniform adapter interface, in which `wrap()` is responsible for converting raw sockets to SSL connections. With this change `bind()` serves no purpose and so is deprecated. This change also fixes the pyOpenSSL adapter so that it doesn't prematurely wrap the raw socket, an issue in the old `bind()` implementation.


cheroot/server.py
```diff
@@ -2112,7 +2110,6 @@ def prepare_socket(  # pylint: disable=too-many-positional-arguments
         type,
         proto,
         nodelay,
-        ssl_adapter,
```
webknjaz, line 2115: Since this changes the public method signature, I wonder if we should try deprecating it first. Although, with the scale of refactoring you're doing, it'll require a major release anyway. So we're probably fine.

cheroot/ssl/__init__.py
```diff
@@ -31,11 +31,6 @@ def __init__(
         self.private_key_password = private_key_password
         self.context = None
 
-    @abstractmethod
-    def bind(self, sock):
```
webknjaz, line 35: In https://github.com/cherrypy/cheroot/pull/800#discussion_r2561059012, I was suggesting that we deprecate the method first and remove later. Although, I doubt anything external would be using it...

cheroot/ssl/pyopenssl.py
```diff
@@ -442,7 +453,9 @@ def makefile(self, sock, mode='r', bufsize=-1):
             if 'r' in mode
             else SSLFileobjectStreamWriter
         )
-        if SSL and isinstance(sock, ssl_conn_type):
+        # sock is an pyopenSSL.SSLConnection instance here
+        # so checking against ssl_conn_type doesn't work
+        if SSL and isinstance(sock, (ssl_conn_type, SSLConnection)):
```
webknjaz: Should we just drop `ssl_conn_type` then?

cheroot/ssl/pyopenssl.py
```diff
@@ -270,6 +272,17 @@ def __init__(self, *args):
         self._ssl_conn = SSL.Connection(*args)
         self._lock = threading.RLock()
 
+    @property
+    def _socket(self):
+        """
+        Expose underlying raw socket.
+
+        This is needed for times when the cheroot server needs access to the
```
webknjaz, line 280: Can we go to the places that are trying to access `_socket` and just use the raw socket that should already be available in there? This way, we wouldn't need to add this property.

cheroot/ssl/pyopenssl.py
```diff
@@ -320,20 +333,18 @@ def __init__(
...
         """Wrap and return the given socket, plus WSGI environ entries."""
         # pyOpenSSL doesn't perform the handshake until the first read/write
         # forcing the handshake to complete tends to result in the connection
         # closing so we can't reliably access protocol/client cert for the env
-        return sock, self._environ.copy()
+        if self.context is None:
+            self.context = self.get_context()
+        conn = SSLConnection(self.context, sock)
+
+        conn.set_accept_state()  # Tell OpenSSL this is a server connection
```
webknjaz, line 344: This looks like a new behavior — is it necessary in the scope of this PR?

cheroot/test/test_ssl.py
```diff
@@ -920,16 +920,14 @@ def test_openssl_adapter_with_false_key_password(
...
-
     with expected_warn, pytest.raises(
         OpenSSL.SSL.Error,
         # Decode error has happened very rarely with Python 3.9 in MacOS.
         # Might be caused by a random issue in file handling leading
         # to interpretation of garbage characters in certificates.
         match=r'.+\'(bad decrypt|decode error)\'.+',
     ):
-        httpserver.prepare()
+        tls_adapter.get_context()
```
webknjaz: This looks like an unrelated change. If we were to refactor this, we'd probably also have to get rid of `httpserver`. Currently, the test seems to be checking for the HTTP server side effects too, not just the adapter.

cheroot/ssl/pyopenssl.py
```diff
@@ -216,6 +216,8 @@ def __new__(mcl, name, bases, nmspc):
             'settimeout',
             'gettimeout',
             'shutdown',
+            'recv_into',
+            '_decref_socketios',
```
webknjaz, line 220: Why are these needed?

cheroot/ssl/pyopenssl.py
```diff
@@ -320,20 +333,18 @@ def __init__(
...
         # forcing the handshake to complete tends to result in the connection
         # closing so we can't reliably access protocol/client cert for the env
-        return sock, self._environ.copy()
+        if self.context is None:
+            self.context = self.get_context()
+        conn = SSLConnection(self.context, sock)
+
+        conn.set_accept_state()  # Tell OpenSSL this is a server connection
+        self._environ = self.get_environ()
+        return conn, self._environ
```
webknjaz: Why does this stop copying the environ? It's mutable after all, we don't want it to be changed externally.

cheroot/ssl/pyopenssl.py
```diff
@@ -320,20 +333,18 @@ def __init__(
...
         # pyOpenSSL doesn't perform the handshake until the first read/write
         # forcing the handshake to complete tends to result in the connection
         # closing so we can't reliably access protocol/client cert for the env
-        return sock, self._environ.copy()
+        if self.context is None:
+            self.context = self.get_context()
+        conn = SSLConnection(self.context, sock)
+
+        conn.set_accept_state()  # Tell OpenSSL this is a server connection
+        self._environ = self.get_environ()
```
webknjaz: Should this be running on each `wrap()` call? I imagine a better place for this could be the initializer where it'd run once.

cheroot/ssl/pyopenssl.py
julianz-: Currently _handle_no_ssl() in HTTPConnection deals with sending the 400 Bad Request message.  It doesn't have access to the raw socket without the unwrapping that's been there for at least 9 years. However, we should be able to get rid of this method once we do what I was proposing in #800. So I would suggest we leave this as is until then.

cheroot/ssl/pyopenssl.py
julianz-: Without this call, I get an 'uninitialized' SSL context error when the handshake occurs:
`OpenSSL.SSL.Error: [('SSL routines', '', 'uninitialized')]`. According to pyOpenSSL documentation, we're supposed to call either `set_accept_state()` (server) or `set_connect_state()` (client) on a fresh `SSL.Connection` before performing I/O. However, I couldn't find `set_accept_state()` in the previous code which is puzzling. Maybe there was an implicit initialization happening that I haven't traced? Regardless, this call seems to be necessary for the code to work correctly now.

cheroot/ssl/pyopenssl.py
julianz-: Without these,  I get errors: AttributeError: 'SSLConnection' object has no attribute 'recv_into' and  '_decref_socketios' running some of the tests such as test_ssl_adapters(). Again, why this wasn't a problem in the old set up is mysterious.

cheroot/ssl/pyopenssl.py
julianz-: The `conn` creation and setting up as a server need to be here for sure - they are per-connection tasks. Looks like `get_environ()` just gets the static info though so have moved this to `init`.

webknjaz: Yep, that was my impression too.

cheroot/ssl/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Implementation of the SSL adapter base interface."""
 
+import warnings
 from abc import ABC, abstractmethod
```
webknjaz: Let's do
```suggestion
from abc import ABC, abstractmethod
from warnings import warn as _warn
```
instead. We've started using this pattern in `pyopenssl.py` already. This is so we don't add any new non-private objects into the module's global namespace.

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
+Deprecated :py:meth:`~cheroot.ssl.__init__.bind` method from
```
webknjaz: I think this is just
```suggestion
Deprecated :py:meth:`~cheroot.ssl.bind` method from
```
*(https://app.readthedocs.org/projects/cheroot/builds/30463447/#295792903--60)*

cheroot/ssl/pyopenssl.py
webknjaz: Okay, with the plan to remove the proxy in the future, it's probably not worth it trying to understand it deeper. Marking as resolved.

cheroot/ssl/__init__.py
```diff
@@ -31,9 +32,20 @@ def __init__(
...
 
-    @abstractmethod
     def bind(self, sock):
-        """Wrap and return the given socket."""
+        """
+        Wrap and return the given socket.
+
+        Deprecated:
+        This method no longer performs any SSL-specific operations.
+        SSL wrapping now happens in wrap(). This method will be
```
webknjaz:
```suggestion
        SSL wrapping now happens in :meth:`.wrap`. This method will be
```

cheroot/ssl/__init__.py
```diff
@@ -31,9 +32,20 @@ def __init__(
...
+        removed in a future version.
+        """
+        warnings.warn(
+            'SSLAdapter.bind() is deprecated and will be removed in a future version.',
+            DeprecationWarning,
```
webknjaz, line 46: Could you add a test for this with `pytest.deprecated()`?

cheroot/ssl/__init__.py
```diff
@@ -31,9 +32,20 @@ def __init__(
         self.private_key_password = private_key_password
         self.context = None
 
-    @abstractmethod
     def bind(self, sock):
-        """Wrap and return the given socket."""
+        """
+        Wrap and return the given socket.
```
webknjaz:
```suggestion
        Return the given socket.
```

cheroot/ssl/__init__.py
webknjaz: Marking this as resovled.

cheroot/ssl/pyopenssl.py
webknjaz: Agreed. Marking as resolved then.

cheroot/ssl/pyopenssl.py
webknjaz: Yes, this is confusing to me too. I suppose, there was something implicit indeed. Marking as resolved then.

cheroot/ssl/pyopenssl.py
webknjaz: I was a bit worried that CherryPy sets these attributes directly but looks like we're good: https://github.com/cherrypy/cherrypy/blob/1f75bc9eed8e0e385f64f368bd69f58d96fb8c2b/cherrypy/_cpwsgi_server.py#L92-L108 / https://github.com/cherrypy/cherrypy/blob/1f75bc9eed8e0e385f64f368bd69f58d96fb8c2b/cherrypy/_cpnative_server.py#L170-L187.

Marking this thread as resolved.

cheroot/ssl/__init__.py
```diff
@@ -31,9 +32,20 @@ def __init__(
...
+
+        Deprecated:
+        This method no longer performs any SSL-specific operations.
+        SSL wrapping now happens in wrap(). This method will be
+        removed in a future version.
+        """
+        warnings.warn(
```
webknjaz:
```suggestion
        _warn(
```

cheroot/server.py
webknjaz: @julianz- do you think we could keep it here and issue a deprecation warning instead? + test it.

cheroot/ssl/pyopenssl.py
```diff
@@ -270,6 +272,17 @@ def __init__(self, *args):
         self._ssl_conn = SSL.Connection(*args)
         self._lock = threading.RLock()
 
+    @property
+    def _socket(self):
+        """
+        Expose underlying raw socket.
+
+        This is needed for times when the cheroot server needs access to the
+        original socket object, e.g. in replying to a client attempting
```
```suggestion
        original socket object, e.g. in response to a client attempting
```

cheroot/test/test_ssl.py
webknjaz: @julianz- could you explain why this whole refactoring of the test is needed in the first place?

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
...
+SSLAdapter interface and implementations.
```
webknjaz:
```suggestion
the :py:class:`~cheroot.ssl.SSLAdapter` interface and respective implementations.
```

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
...
+Previously ``bind()`` was doing nothing in the ``builtin`` SSLAdapter
```
webknjaz:
```suggestion
Previously :py:meth:`~cheroot.ssl.bind` was doing nothing in the :py:class:`builtin TLS adapter <cheroot.ssl.builtin.BuiltinSSLAdapter>`
```

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
...
+but was being used in the ``pyOpenSSLAdapter`` to wrap the socket.
```
webknjaz:
```suggestion
but was being used in the :py:class:`~cheroot.ssl.builtin.pyOpenSSLAdapter` to wrap the socket.
```

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
...
+Socket wrapping is now done exclusively in the ``wrap()``
```
webknjaz:
```suggestion
Socket wrapping is now done exclusively in the :py:meth:`~cheroot.ssl.wrap`
```

cheroot/test/test_ssl.py
julianz-: I changed the test as it no longer worked. The test used to rely on `httpserver.prepare()` calling `get_context()` but since moving the latter from `bind()` to `wrap(),` `prepare()` no longer does this.  However, this raises a point you mentioned elsewhere. We obviously don't want to validate the server certificate on every call so I have moved the `get_context()` call to `init`. But this make the whole `http_server` set up redundant.  The bad password error is raised now as soon as we initialize the adapter.

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,8 @@
+Deprecated :py:meth:`~cheroot.ssl.Adapter.bind` method from the :py:class:`~cheroot.ssl.Adapter`
+interface and respective implementations. Previously :py:meth:`~cheroot.ssl.Adapter.bind` was
+doing nothing in the `:py:class:`builtin TLS adapter <cheroot.ssl.builtin.BuiltinSSLAdapter>`
+but was being used in the :py:class:`~cheroot.ssl.pyOpenSSLAdapter` to wrap the socket.
```
webknjaz: Looks like I made some typos making suggestiosn from browser.
```suggestion
but was being used in the :py:class:`~cheroot.ssl.pyopenssl.pyOpenSSLAdapter` to wrap the socket.
```

*(looks like this is the only problem in Sphinx build: https://app.readthedocs.org/projects/cheroot/builds/30475668/#295926113--60)*

cheroot/test/test_ssl.py
```diff
@@ -908,28 +911,38 @@ def test_openssl_adapter_with_false_key_password(
...
+    # pylint: disable=abstract-class-instantiated
+    return Adapter(
+        certificate='cert.pem',
+        private_key='key.pem',
+    )
+
+
+def test_bind_deprecated_call(dummy_adapter):
+    """Test deprecated bind() method issues a warning and returns socket."""
+    s = socket.socket()
```
webknjaz: We should close the socket so that it doesn't cause resource warnings.

cheroot/ssl/__init__.py
webknjaz: Resolved.

cheroot/test/test_ssl.py
```diff
@@ -908,28 +911,38 @@ def test_openssl_adapter_with_false_key_password(
...
+    monkeypatch.setattr(Adapter, '__abstractmethods__', set())
+    # pylint: disable=abstract-class-instantiated
+    return Adapter(
+        certificate='cert.pem',
+        private_key='key.pem',
+    )
+
+
+def test_bind_deprecated_call(dummy_adapter):
+    """Test deprecated bind() method issues a warning and returns socket."""
```
```suggestion
    """Test deprecated ``bind()`` method issues a warning and returns socket."""
```

julianz-:
@webknjaz This PR is failing in some instances running `test_http_over_https_error` e.g.: ["ERROR at teardown of test_http_over_https_error[::-builtin]." ](https://github.com/cherrypy/cheroot/actions/runs/19726077067/job/56517763645?pr=801) The problem seems to arise only with the builtin adapter because this forces a handshake in `wrap()` and after the handshake fails the state of the socket is dead running on my Mac (fd=-1). In some other environments the test seems to pass in the CI, so I guess the socket can still be usable but it depends - maybe there is a race condition.  Anyway, I can fix this issue by overriding `_flush_unlocked()` in `StreamWriter`. This is more of bug fix though so should probably go in a different PR. With that fix, I suspect the test becomes less flaky too. Thoughts?

cheroot/test/test_ssl.py
webknjaz: Alright. Makes sense. Though, maybe it'd be good to have a full e2e test at some point in the future.
Marking as resolved.

cheroot/test/test_ssl.py
```diff
@@ -908,28 +909,108 @@ def test_openssl_adapter_with_false_key_password(
...
+    ]
+    assert len(deprecation_warnings) == 0
 
-    assert not httpserver.requests._threads
-    assert not httpserver.ready
+    # Check that the returned object is indeed a socket
+    assert isinstance(sock, socket.socket)
+    # Check we have a socket configured with file descriptor
+    assert sock.fileno() > 0
+    sock.close()
```
webknjaz: I don't think we need a test for this since pytest is already configured to turn any warnings, including deprecations, into errors. So it'd already explode in many other tests if misconfigured.

cheroot/test/test_ssl.py
```diff
@@ -908,28 +909,108 @@ def test_openssl_adapter_with_false_key_password(
...
+    ``ssl_adapter`` has been deprecated in ``prepare_socket()``.
+    """
+    # Required parameters for prepare_socket (standard IPv4 TCP config)
+    bind_addr = ('127.0.0.1', 8080)
+    family = socket.AF_INET
+    sock_type = socket.SOCK_STREAM
+    proto = socket.IPPROTO_TCP
+    nodelay = True
+
+    with pytest.deprecated_call():
```
webknjaz: Let's try checking the message via `match=`.

cheroot/server.py
```diff
@@ -2115,7 +2119,25 @@ def prepare_socket(  # pylint: disable=too-many-positional-arguments
         ssl_adapter,
```
```suggestion
        ssl_adapter=None,
```

cheroot/server.py
```diff
@@ -1987,7 +1989,9 @@ def bind(self, family, type, proto=0):
             type,
             proto,
             self.nodelay,
-            self.ssl_adapter,
+            # ssl_adapter is passed to HTTPServer, not needed here
+            # this parameter is deprecated and will be removed in future
+            None,
             self.reuse_port,
```
webknjaz: If we set the default in the method signature, we can get away with not passing it at all here:
```suggestion
            reuse_port=self.reuse_port,
```

cheroot/server.py
```diff
@@ -1369,6 +1370,7 @@ def _handle_no_ssl(self, req):
             'this server only speaks HTTPS on this port.'
         )
         req.simple_response('400 Bad Request', msg)
+        self.wfile.flush()
```
webknjaz: This looks like a new behavior not mentioned in change notes. It probably deserves a separate PR.

cheroot/server.py
```diff
@@ -2115,7 +2119,25 @@ def prepare_socket(  # pylint: disable=too-many-positional-arguments
...
-        """Create and prepare the socket object."""
+        """
+        Create and prepare the socket object.
+
+        :param ssl_adapter: Legacy SSL adapter parameter.
+            This argument is now ignored internally.
+
+        :deprecated ssl_adapter: This parameter is now deprecated and will be
+            removed in a future release. Pass the adapter to the 'HTTPServer`
+            constructor instead.
```
webknjaz: This renders a little weird: https://cheroot--801.org.readthedocs.build/en/801/pkg/cheroot.server/#cheroot.server.HTTPServer.prepare_socket.

I don't think `:deprecated:` exists in this context: https://github.com/sphinx-doc/sphinx/blob/4ea0d9fef464390694cc600fda89a924c2336b69/sphinx/domains/python/_object.py#L187-L232 / https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists. It only exists for module-global definitions, it appears: https://www.sphinx-doc.org/en/master/usage/domains/python.html#directive-option-py-module-deprecated.

Plus, there's a single quote + backtick mismatch.

```suggestion
        :param ssl_adapter: Legacy SSL adapter parameter.
            This argument is now ignored internally. It is now deprecated
            and will be removed in a future release. Pass the adapter to
            the :class:`HTTPServer` initializer instead.
```

cheroot/server.py
julianz-: Oops - that was the beginning of some experiments I did to figure out the source of those bad socket errors. Gone now.

julianz-:
@webknjaz 

> > If you're worried about the 3.14 jobs — they all are marked as allowed to fail and fixing that would be a separate effort. 

Interestingly, I am able to repro the error with 100% reliability on MacOS using Python v3.12. I notice that the test excludes MacOS with the builtin adapter for reasons that were not well explained in #225 apart from the fact that the test wouldn't pass! I am tempted to do a new PR and you can see for yourself. The fix is pretty simple. There's no reason to exclude MacOS anymore too.

webknjaz:
> I tempted to do a new PR and you can see for yourself. The fix is pretty simple. There's no reason to exclude MacOS anymore too.

Looking forward!

docs/changelog-fragments.d/801.deprecation.rst
```diff
@@ -0,0 +1,9 @@
+Deprecated :py:meth:`~cheroot.ssl.Adapter.bind` method from the :py:class:`~cheroot.ssl.Adapter`
+interface and respective implementations. Previously :py:meth:`~cheroot.ssl.Adapter.bind` was
+doing nothing in the `:py:class:`builtin TLS adapter <cheroot.ssl.builtin.BuiltinSSLAdapter>`
```
webknjaz: A backtick sneaked in
```suggestion
doing nothing in the :py:class:`builtin TLS adapter <cheroot.ssl.builtin.BuiltinSSLAdapter>`
```
*This is the only thing that seems to cause odd render @ 
https://cheroot--801.org.readthedocs.build/en/801/history/#deprecations-removal-in-next-major-release*

webknjaz (approved):
Wonderful, thanks!

PR #802, opened 2025-11-28
Handle socket errors when sending HTTP-over-HTTPS error responses

julianz-:
When a client sends plaintext HTTP to a TLS-only port, the SSL layer detects the invalid handshake and may close the underlying socket. The server attempts to send a 400 Bad Request error response, but the socket may already be closed, causing OSError during the flush.

With pyOpenSSL, the response usually succeeds. With the builtin SSL adapter, the socket is typically closed before the write can occur.

This fix overrides `_flush_unlocked()` in `StreamWriter` to catch `OSError` and clear the buffer. This allows:
- The explicit flush to fail gracefully when sending the 400 response
- Object finalization (`__del__`) to complete without errors

Note that swallowing OSErrors here should not affect normal communication: `write()` calls upon `_flush_unlocked()` in `StreamWriter` which in turn calls the base class implementation `_flush_unlocked` in BufferedWriter. The base implementation handles errors of type `BlockingIOError`  such as `WantWriteError` and `WantReadsError`. But if the socket is dead an OSError is raised which propagates up to `StreamWriter` where `_flush_unlocked()` now handles the error.

### DIFF
--- a/cheroot/connections.py
+++ b/cheroot/connections.py
@@ -327,6 +327,8 @@ class ConnectionManager:
                     wfile = mf(s, 'wb', io.DEFAULT_BUFFER_SIZE)
                     try:
                         wfile.write(''.join(buf).encode('ISO-8859-1'))
+                        wfile.flush()
+                        wfile.close()
                     except OSError as ex:
                         if ex.args[0] not in errors.socket_errors_to_ignore:
                             raise
--- a/cheroot/makefile.py
+++ b/cheroot/makefile.py
@@ -68,6 +68,26 @@ class StreamWriter(BufferedWriter):
         self.bytes_written += len(val)
         return res
 
+    def _flush_unlocked(self):
+        """
+        Flush buffered output to the underlying socket.
+
+        We override this method because when a client sends plaintext HTTP
+        to a TLS-only port, SSL detects a bad handshake and may
+        invalidate the underlying socket. At that point, the socket
+        may not be writable. Attempting to write (e.g., sending a
+        400 Bad Request) may succeed or raise OSError. This override
+        prevents OSError from propagating.
+        """
+        try:
+            super()._flush_unlocked()
+        except OSError:
+            # The socket is already closed or otherwise unusable (e.g. TLS
+            # fatal alert triggered by invalid handshake).
+            # Clearing the buffer prevents the error from happening
+            # again during cleanup.
+            self._write_buf.clear()
+
 
 def MakeFile(sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
     """File object attached to a socket object."""
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -40,6 +40,7 @@ netloc
 noqa
 PIL
 pipelining
+plaintext
 positionally
 pre
 preconfigure
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1370,6 +1370,7 @@ class HTTPConnection:
             'this server only speaks HTTPS on this port.'
         )
         req.simple_response('400 Bad Request', msg)
+        self.wfile.flush()
         self.linger = True
 
     def _conditional_error(self, req, response):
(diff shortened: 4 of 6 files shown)


cheroot/connections.py
```diff
@@ -327,6 +327,8 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
                     wfile = mf(s, 'wb', io.DEFAULT_BUFFER_SIZE)
                     try:
                         wfile.write(''.join(buf).encode('ISO-8859-1'))
+                        wfile.flush()
+                        wfile.close()
```
webknjaz, line 331: plz avoid having more than one instruction in try-blocks.

cheroot/makefile.py
```diff
@@ -68,6 +68,26 @@ def write(self, val, *args, **kwargs):
...
+    def _flush_unlocked(self):
+        """
+        Flush buffered output to the underlying socket.
+
+        We override this method because when a client sends plaintext HTTP
+        to a TLS-only port, SSL detects a bad handshake and may
+        invalidate the underlying socket. At that point, the socket
+        may not be writable. Attempting to write (e.g., sending a
+        400 Bad Request) may succeed or raise OSError. This override
+        prevents OSError from propagating.
```
webknjaz, line 80: This part looks like a code comment, not a docstring.

docs/spelling_wordlist.txt
```diff
@@ -40,6 +40,7 @@ netloc
 noqa
 PIL
 pipelining
+plaintext
```
webknjaz, line 43: We might not need this if we don't put that comment into a docstring but have it as a comment.

cheroot/makefile.py
```diff
@@ -68,6 +68,26 @@ def write(self, val, *args, **kwargs):
...
+        """
+        try:
+            super()._flush_unlocked()
+        except OSError:
+            # The socket is already closed or otherwise unusable (e.g. TLS
+            # fatal alert triggered by invalid handshake).
+            # Clearing the buffer prevents the error from happening
+            # again during cleanup.
+            self._write_buf.clear()
```
webknjaz, line 89: I'm not sure if we should be suppressing problems on the low level like this. It's likely the caller that should perform the suppression since the stream writer does not know the context it's used in (nor should it).

cheroot/connections.py
webknjaz: Also, this should probably be a `with`-block anyway.

cheroot/makefile.py
webknjaz, line 1: I don't like that we mess with private attributes (although, the existing code already accesses it). In particular, it's weird how TLS quirks leak into makefile. I'd wait for other bits of refactoring, especially the removal of makefile in favor of classmethods before doing anything about this.

webknjaz:
@julianz- did this become redundant after the last merge? Anything worth salvaging?

julianz-:
I think this fix is still useful. I saw on my last commit on my fork for #800, "OSError: [Errno 9] Bad file descriptor" coming from one environment because `StreamWriter` was still trying to complete a `write()` to a dead socket. The main idea in this PR is to make StreamWriter more resilient to dealing with sockets that are being torn down. I can work on rebasing and updating this PR but I wasn't sure whether you were dead set against the idea because of the need to override private methods?

webknjaz:
It's possible that #779 was closer to the solution partially, but incomplete. I'd still postpone this one for now, though.

avinashkamat48 (commented):
I think the StreamWriter._flush_unlocked() override is too broad. It catches every OSError from super()._flush_unlocked() and clears _write_buf, which means normal response writes can now silently drop buffered bytes for any socket write error, not just the HTTP-over-HTTPS error path this PR is trying to handle. That changes the behavior of all MakeFile(..., 'w') users and can hide real connection/write failures that callers currently see. Could the suppression be scoped to the HTTP-over-HTTPS error response path instead of changing the general writer primitive?

julianz-:
I think this is superseded by #820 which handles the `OSError` suppression in `_send_bad_request_plain_http_error` directly on the raw socket via `_suppress_socket_io_errors`, avoiding the need to touch makefile.py. Planning to close this assuming #820 is a go.

PR #805, opened 2025-12-08
Deprecate `MakeFile()` in favor of `StreamReader` and `StreamWriter`

julianz-:
We replace the `MakeFile` factory function with direct class instantiation. The `MakeFile` function now emits a `DeprecationWarning` and will be removed in a future release.

The `makefile()` methods in the SSL adapters and the base `Adapter` class are also deprecated. These methods are no longer necessary because the adapters now return wrapped sockets that work directly with `StreamReader` and `StreamWriter`.

The `pyOpenSSL` adapter wraps its `SSL.Connection` objects in a new `TLSSocket` class. `TLSSocket` implements a socket interface that handles OpenSSL-specific errors (`WantReadError`, `WantWriteError`, etc.). This allows plain `StreamReader` and `StreamWriter` to work transparently with `pyOpenSSL` connections.

Although its `makefile` is no longer used, the `builtin` SSL adapter continues to work essentially as before - Python's `ssl.SSLSocket` already provided a compatible interface that works directly with `StreamReader` and `StreamWriter`.

This change eliminates the abstraction leak where `HTTPConnection` needed to call upon an SSL adapter custom `makefile()` method. Now, `HTTPConnection` simply creates `StreamReader` and `StreamWriter` instances with whatever socket it receives, whether it's a plain socket, `ssl.SSLSocket`, or `TLSSocket`.

This is a step towards unifying the two SSL adapters while simplifying higher-level logic.


cheroot/connections.py
```diff
@@ -123,6 +124,15 @@ def _suppress_socket_io_errors(socket, /):
             raise
 
 
+def _make_file_object(sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
```
webknjaz: I was hoping this trampoline would be made redundant. Why is it necessary? What are the challenges in using the right stream constructors right away?

julianz-: Ok maybe this is better. I have jettisoned `_make_file_object`. `HTTPConnection` now checks `server.ssl_adapter` to decide whether to use plain StreamReader/StreamWriter or, if there is one, the SSL adapter's makefile() method. pyopenSSL adapter still needs such a method so that it can offer wrapped streaming methods that know how to deal with `WantReadError`, `WantWriteError` etc. The builtin adapter is fine relying on the plain streaming classes.

cheroot/server.py
```diff
@@ -1287,8 +1287,18 @@ def __init__(self, server, sock, makefile=MakeFile):
...
         self.socket = sock
-        self.rfile = makefile(sock, 'rb', self.rbufsize)
-        self.wfile = makefile(sock, 'wb', self.wbufsize)
+
+        # Check if SSL adapter needs custom wrapped streaming
+        # objects (pyOpenSSL does, builtin doesn't)
+        if server.ssl_adapter is not None and hasattr(
+            server.ssl_adapter,
+            'makefile',
+        ):
```
webknjaz: This looks like an abstraction leak. We shouldn't make TLS adapters and HTTP connections coupled. Let's think of a better solution. I wonder if this initializer could work with a pre-wrapped socket or something. It doesn't feel right that HTTP layer would care whether TCP's got a TLS tunnel or not.

Also, having a conditionally existing method on adapters seems fragile.

cheroot/makefile.py
```diff
@@ -70,6 +71,15 @@ def write(self, val, *args, **kwargs):
 
 
 def MakeFile(sock, mode='r', bufsize=io.DEFAULT_BUFFER_SIZE):
-    """File object attached to a socket object."""
+    """
+    File object attached to a socket object.
+
+    This method is now deprecated: Use StreamReader or StreamWriter directly.
```
webknjaz: It's a function, not a method.

cheroot/server.py
```diff
@@ -1276,7 +1276,7 @@ class HTTPConnection:
     # Fields set by ConnectionManager.
     last_used = None
 
-    def __init__(self, server, sock, makefile=MakeFile):
+    def __init__(self, server, sock):
```
webknjaz: Removing args from method signatures is a breaking API change. They'd need to be deprecated first. Although, I'm not entirely sure how to handle the constructor split beyond that.

cheroot/ssl/__init__.py
```diff
@@ -108,8 +111,3 @@ def wrap(self, sock):
     def get_environ(self):
         """Return WSGI environ entries to be merged into each request."""
         raise NotImplementedError  # pragma: no cover
-
-    @abstractmethod
-    def makefile(self, sock, mode='r', bufsize=-1):
```
webknjaz, line 114: This is a public API change, it'd need to be deprecated.

cheroot/server.py
webknjaz: I'm starting to wonder if the server instance would need a `makefile()` method that would hold the callable selection logic...

cheroot/ssl/builtin.py
```diff
@@ -11,23 +11,15 @@
...
 except ImportError:
     ssl = None
 
-try:
-    from _pyio import DEFAULT_BUFFER_SIZE
-except ImportError:
-    try:
-        from io import DEFAULT_BUFFER_SIZE
-    except ImportError:
-        DEFAULT_BUFFER_SIZE = -1
```
webknjaz, line 27: This public attr would probably need to be deprecated (through `__getattr__()`).

cheroot/server.py
```diff
@@ -1358,12 +1370,8 @@ def communicate(self):  # noqa: C901  # FIXME
...
         if not req or req.sent_headers:
             return
-        # Unwrap wfile
-        try:
-            resp_sock = self.socket._sock
-        except AttributeError:
-            # self.socket is of OpenSSL.SSL.Connection type
-            resp_sock = self.socket._socket
+        # Unwrap to get raw TCP socket
+        resp_sock = self.socket._sock
```
webknjaz, line 1380: Does this attr always exist?

cheroot/server.py
julianz-: This is largely what lead to my earlier `TLSSocket` approach but I've revisited that implementing a more focussed version. The TLSSocket class now wraps the pyOpenSSL SSL.Connection to handle OpenSSL-specific errors like WantReadError internally. This allows plain StreamReader/StreamWriter to work directly now with pyOpenSSL connections. So we get rid of `makefile` without any abstraction leaks hopefully. See what you think.

cheroot/server.py
julianz-: Currently `handle_no_ssl()` will only ever get called when using `PyOpenSSLAdapter` so yes provided we make sure it's there on that adapter this should work.  However, the property mimics the same pattern that `ssl.SSLSocket` uses and also `socket.socket`.

cheroot/server.py
```diff
@@ -1276,19 +1276,31 @@ class HTTPConnection:
...
+    def __init__(self, server, sock, makefile=None):
         """Initialize HTTPConnection instance.
 
         Args:
             server (HTTPServer): web server object receiving this request
             sock (socket._socketobject): the raw socket object (usually
                 TCP) for this connection
-            makefile (file): a fileobject class for reading from the socket
+            makefile (file): Now deprecated
+                Use to be a fileobject class for reading from the socket
```
```suggestion
                Used to be a fileobject class for reading from the socket
```

cheroot/server.py
```diff
@@ -1276,19 +1276,31 @@ class HTTPConnection:
...
-    def __init__(self, server, sock, makefile=MakeFile):
+    def __init__(self, server, sock, makefile=None):
         """Initialize HTTPConnection instance.
 
         Args:
             server (HTTPServer): web server object receiving this request
             sock (socket._socketobject): the raw socket object (usually
                 TCP) for this connection
-            makefile (file): a fileobject class for reading from the socket
+            makefile (file): Now deprecated
```
```suggestion
            makefile (file): Now deprecated.
```

cheroot/ssl/__init__.py
```diff
@@ -56,13 +56,24 @@ def _ensure_peer_speaks_https(raw_socket, /) -> None:
...
+    """
+    Base class for SSL driver library adapters.
 
     Required methods:
-
-        * ``wrap(sock) -> (wrapped socket, ssl environ dict)``
-        * ``makefile(sock, mode='r', bufsize=DEFAULT_BUFFER_SIZE) ->
-          socket file object``
+        ``wrap(sock)``: Wrap a socket with SSL. Also returns ssl environ dict.
+        ``get_environ()``: Get SSL environment variables as a dict.
```
webknjaz: Keep leading asterisks and an empty line after the colon, so RST would parse a proper list. By removing those, you've forced everything into a single line w/o paragraphs or other visible delimiters. Use this RTD preview for checking if your changes look as expected: https://cheroot--805.org.readthedocs.build/en/805/pkg/cheroot.ssl/#cheroot.ssl.Adapter.

cheroot/ssl/__init__.py
```diff
@@ -112,7 +123,11 @@ def get_environ(self):
 
     @abstractmethod
```
webknjaz: If it's deprecated, there's no need to force its presence by marking the method as abstract.

cheroot/ssl/builtin.py
```diff
@@ -7,31 +7,42 @@
 ``BuiltinSSLAdapter``.
 """
 
+import io
```
webknjaz: Let's do this to keep the import out of the global namespace:
```suggestion
import io as _io
```

cheroot/ssl/builtin.py
```diff
@@ -7,31 +7,42 @@
...
+    if name == 'DEFAULT_BUFFER_SIZE':
+        _warn(
+            (
+                '`DEFAULT_BUFFER_SIZE` is deprecated and '
+                'will be removed in a future release.'
+            ),
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return io.DEFAULT_BUFFER_SIZE
```
webknjaz: If that import is only needed here, it's okay to move it to this place as an exception. This way, `from cheroot.ssl.builtin import *` won't load `io`.

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +455,91 @@ def get_environ(self):
...
+        """Shutdown the connection."""
+        return self._ssl_conn._decref_socketios()
+
+    def shutdown(self, how):
+        """Shutdown the connection."""
+        # SSL shutdown can fail if the handshake didn't complete or during
+        # error conditions. Since shutdown is a cleanup operation, we suppress
+        # all errors - the connection will be closed anyway.
+        with _suppress(SSL.Error, socket.error, OSError):
+            return self._ssl_conn.shutdown(how)
```
webknjaz: This method doesn't accept any args: https://www.pyopenssl.org/en/latest/api/ssl.html#OpenSSL.SSL.Connection.shutdown. If called like this, it'll always trigger a `TypeError`. It should be called w/o `how` or alternatively, we'd need to call `.sock_shutdown(how)` instead (provided that actually works).

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +455,91 @@ def get_environ(self):
...
+    def _decref_socketios(self):
+        """Shutdown the connection."""
+        return self._ssl_conn._decref_socketios()
+
+    def shutdown(self, how):
+        """Shutdown the connection."""
+        # SSL shutdown can fail if the handshake didn't complete or during
+        # error conditions. Since shutdown is a cleanup operation, we suppress
+        # all errors - the connection will be closed anyway.
+        with _suppress(SSL.Error, socket.error, OSError):
```
webknjaz: The last two items are the same:
```pycon
$ python    
Python 3.13.0 (main, Nov 18 2024, 05:02:35) [GCC 14.2.1 20241026] on linux
>>> import socket
>>> socket.error is OSError
True
```
`socket.error` is an alias of `OSError` since Python 3.3 so there's no need to list it twice. It's still present in a few places in the repo but that's old code that should eventually be updated. For any new code, we don't need to use deprecated names for things.

---

Also, we should probably be able to intercept `ConnectionError` or its subclasses instead of the generic `OSError` which might suppress more things than we actually want. The suppression should be as narrow as possible (within the limits of actually known/expected errors).

---

Not sure if we need to dig into subclasses of `SSL.Error` but maybe for context.

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +455,91 @@ def get_environ(self):
...
+
+    def _decref_socketios(self):
+        """Shutdown the connection."""
+        return self._ssl_conn._decref_socketios()
+
+    def shutdown(self, how):
+        """Shutdown the connection."""
+        # SSL shutdown can fail if the handshake didn't complete or during
+        # error conditions. Since shutdown is a cleanup operation, we suppress
+        # all errors - the connection will be closed anyway.
```
webknjaz: I'm not entirely convinced that this should just be suppressed on the lower level. It should probably be transformed into a framework-specific exception to communicate this type of failure, allowing the server to log an error on the calling side and decide to suppress there.
So in order for us to be tracing-friendly, we I'd say that we should make sure that anything in TLS adapters processes own exceptions, stores the context and re-raises them as our own exception type with inspectable/loggable context attached. This would unify interfacing with the TLS layer from the server.

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +455,91 @@ def get_environ(self):
...
+        with _suppress(SSL.Error, OSError):
+            self._ssl_conn.shutdown()
+
+        # 2. Close the SSL connection object
+        with _suppress(SSL.Error, OSError):
+            self._ssl_conn.close()
+
+        # 3. Close the raw TCP socket
+        with _suppress(OSError):
+            self._sock.close()
```
webknjaz: Same sentiment here. If we want to perform multiple post-error actions, we could store the exceptions and reraise them as our own + maybe an exception group for never Pythons.

cheroot/ssl/pyopenssl.py
julianz-: For now, I added all exceptions of type `SSL.Error` into `acceptable_sock_shutdown_exceptions` which are swallowed by `_close_kernel_socket()` in `HTTPConnection` without logging  - following the way `BrokenPipeError`, `ConnectionResetError` are handled.

cheroot/errors.py
```diff
@@ -87,4 +87,22 @@ def plat_specific_errors(*errnames):
...
+# WPS229 too long ``try`` body length:
+try:  # noqa: WPS229
+    from OpenSSL import SSL
+
+    # If import succeeds, create tuple with SSL.Error
+    acceptable_sock_shutdown_exceptions = (
+        BrokenPipeError,
+        ConnectionResetError,
+        SSL.Error,
+        SSL.SysCallError,  # need this for (10058, 'WSAESHUTDOWN')
```
webknjaz: This should remain in `pyopenssl.py`. It'd re-raise something like `FatalSSLAlert` (or a new exception if that one doesn't fit). And with that, things outside `cheroot.ssl` wouldn't need to know or understand that `OpenSSL.SSL` even exists.

cheroot/ssl/pyopenssl.py
```diff
@@ -168,14 +167,6 @@ def send(self, *args, **kwargs):
         )
 
 
-class SSLFileobjectStreamReader(SSLFileobjectMixin, StreamReader):
-    """SSL file object attached to a socket object."""
-
-
-class SSLFileobjectStreamWriter(SSLFileobjectMixin, StreamWriter):
-    """SSL file object attached to a socket object."""
```
webknjaz, line 176: We probably should deprecate these too.

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
-            wrapped_socket.ssl_timeout = sock.gettimeout()
-            return wrapped_socket
-        # This is from past:
-        # TODO: figure out what it's meant for
-        return cheroot_server.CP_fileobject(sock, mode, bufsize)
+
+        return sock.makefile(mode, bufsize)
+
+
+class TLSSocket(SSLFileobjectMixin):
```
webknjaz: I don't know if we should expose this type as a public importable. Perhaps, it's best to start with this being an underscored private class and then, we'd see if anything needs it. Would it work with typing?

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+        """Receive data into a buffer (socket interface)."""
+        with self._lock:
+            result = self._safe_call(
+                True,
+                self._ssl_conn.recv_into,
+                buffer,
+                nbytes,
+            )
+            # Handle the case where _safe_call returns b'' for EOF
+            if result == b'':
```
webknjaz: Could you talk more about the need to handle this? Why isn't this an exception, for example? What's the semantic meaning of an empty byte string?

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+    def _decref_socketios(self):
+        """Shutdown the connection."""
+        return self._ssl_conn._decref_socketios()
+
+    def shutdown(self, how):
+        """Shutdown the connection."""
+        # SSL shutdown can fail if the handshake didn't complete or during
+        # other error conditions. Let errors propagate to be handled by
+        # the server
+        return self._ssl_conn.sock_shutdown(how)
```
webknjaz: But this isn't an SSL shutdown, it's the raw socket shutdown, right?

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+
+    # C901 too complex
+    def close(self):  # noqa: C901
+        """
+        Close the TLS socket and underlying connection.
+
+        Best-effort close: attempts all close operations even if some fail.
+        Let the first error propagate.
+        """
+        first_error = None
```
webknjaz: Make it a list, not a single value. This way, we'll be able to raise an exception group under Python 3.11 and just the first one in older runtimes.

cheroot/ssl/pyopenssl.py
webknjaz: Would it make sense to do something like
```suggestion
        self._ssl_conn.shutdown()
        return self._ssl_conn.sock_shutdown(how)
```
?

webknjaz: Do I understand correctly that the above ^ mimics the stdlib ssl behavior?

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+        except (SSL.Error, OSError) as err:
+            if first_error is None:
+                first_error = err
+
+        # 3. Close the raw TCP socket
+        try:
+            self._sock.close()
+        except OSError as sock_err:
+            if first_error is None:
+                first_error = sock_err
```
webknjaz: As discovered in https://github.com/pyca/pyopenssl/issues/1465, `SSL.Connection` objects don't actually have own `close()`. The first one you're calling is just proxied into the underlying socket object through `__getattr__()`.

And so this would mean that (2) and (3) is actually exactly the same call repeated. And this also means that they would never raise `SSL.Error` because these don't actually interact with PyOpenSSL at all. Hence no need to shield from them. Only `ConnectionError` would make some sense.

docs/changelog-fragments.d/805.deprecation.rst
```diff
@@ -0,0 +1,12 @@
+Deprecated :py:meth:`~cheroot.makefile.MakeFile` in favor of :py:class:`~cheroot.makefile.StreamReader`
+and :py:class:`~cheroot.makefile.StreamWriter`.
+
+The ``MakeFile`` factory function is replaced with direct class instantiation
+and now emits a ``DeprecationWarning``. The function will be removed in
```
webknjaz: This is linkable.
```suggestion
and now emits a :py:exc:`DeprecationWarning`. The function will be removed in
```

docs/changelog-fragments.d/805.deprecation.rst
```diff
@@ -0,0 +1,12 @@
+Deprecated :py:meth:`~cheroot.makefile.MakeFile` in favor of :py:class:`~cheroot.makefile.StreamReader`
+and :py:class:`~cheroot.makefile.StreamWriter`.
+
+The ``MakeFile`` factory function is replaced with direct class instantiation
```
webknjaz: IIRC, this should let Sphinx check that the referenced object is correct without actually adding a hyperlink:
```suggestion
The :py:func:`!MakeFile` factory function is replaced with direct class instantiation
```

docs/changelog-fragments.d/805.deprecation.rst
```diff
@@ -0,0 +1,12 @@
+Deprecated :py:meth:`~cheroot.makefile.MakeFile` in favor of :py:class:`~cheroot.makefile.StreamReader`
```
webknjaz: Pretty sure it's a function:
```suggestion
Deprecated :py:func:`~cheroot.makefile.MakeFile` in favor of :py:class:`~cheroot.makefile.StreamReader`
```
*(although, Sphinx doesn't seem to mind: https://cheroot--805.org.readthedocs.build/en/805/history/#deprecations-removal-in-next-major-release)*

docs/changelog-fragments.d/805.deprecation.rst
```diff
@@ -0,0 +1,12 @@
...
+and :py:class:`~cheroot.makefile.StreamWriter`.
+
+The ``MakeFile`` factory function is replaced with direct class instantiation
+and now emits a ``DeprecationWarning``. The function will be removed in
+a future release.
+
+Adapter specific ``makefile()`` methods are also deprecated in :py:mod:`~cheroot.ssl.builtin.BuiltinSSLAdapter`,
```
webknjaz:
```suggestion
Adapter-specific ``makefile()`` methods are also deprecated in :py:mod:`~cheroot.ssl.builtin.BuiltinSSLAdapter`,
```

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,82 @@ def test_prepare_socket_emits_deprecation_warning(
...
+def test_default_buffer_size_deprecated():
+    """Test accessing ``DEFAULT_BUFFER_SIZE`` raises warning."""
+    with pytest.warns(
+        DeprecationWarning,
+        match='`DEFAULT_BUFFER_SIZE` is deprecated',
+    ):
+        val = builtin_adapter.DEFAULT_BUFFER_SIZE
+
+    # Check that the returned value is correct
+    assert val == io.DEFAULT_BUFFER_SIZE, 'DEFAULT_BUFFER_SIZE value mismatch'
```
webknjaz: Does that second `assert` argument add anything of value that `pytest` itself doesn't? It usually is better at this since it's got internal special cases and rewrites for `assert` bytecodes that allow `pytest` extract more detail about the expressions being passed to `assert` and the surrounding variable values + stack frames.
```suggestion
    assert val == io.DEFAULT_BUFFER_SIZE
```

docs/changelog-fragments.d/805.deprecation.rst
webknjaz: *(the `!` idea is borrowed from CPython Style Guide, FTR: https://devguide.python.org/documentation/style-guide/#links)*

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,82 @@ def test_prepare_socket_emits_deprecation_warning(
...
+    ),
+)
+def test_adapter_makefile_deprecation(
+    mocker,
+    adapter_type,
+    tls_certificate_chain_pem_path,
+    tls_certificate_private_key_pem_path,
+):
+    """Test the adapter's makefile() method emits a deprecation warning."""
+    # Mock the adapter
```
webknjaz: This comment is probably misleading — you're not actually mocking an adapter. The following few instructions initialize a fully functional truely usable adapter object. They aren't even stubs.

If I were to pedantically classify them, I'd probably use the term "fakes" according to Martin Fowler: 
https://martinfowler.com/articles/mocksArentStubs.html (https://martinfowler.com/bliki/TestDouble.html).

---

This is to say that most of the time we don't need comments and they can bring more harm. We usually don't need to re-tell what the code is doing as a comment because:
1) the comment and code coupled with it will eventually become out-of-sync due to human factor
2) comment may drift many lines away over time due to more lines appearing in-between (this is almost the same but it slightly different)
3) the comment may be inaccurate from the beginning

One legitimate case for writing code comments is explaining motivation behind something weird/unobvious/entangled/out-of-place that you cannot change ­— do so after you've exhausted all the avenues of refactoring and making code speak for itself (for example, through better abstraction separation or naming variables for what they are).

Another case is related — explaining how something difficult-to-grasp works (again, if it cannot be made simple with code).

Usually, everybody understands that the tests are structured following the AAA pattern so it's not necessary to add those Arrange-Act-Assert comments all over. Where possible, I'd even move the Arrange phase into a fixture. But there's not many pre-existing examples of this here so we don't need to start doing it here right now — this would need a focused standalone refactoring.

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,82 @@ def test_prepare_socket_emits_deprecation_warning(
...
+
+    expected_message = r'makefile.*deprecated'
+
+    # Act & Assert
+    with pytest.deprecated_call(match=expected_message):
+        # Instantiate HTTPConnection, passing the deprecated 'makefile'
+        conn = HTTPConnection(
+            server=dummy_server,
+            sock=dummy_sock,
+            makefile=dummy_makefile_value,  # This triggers the warning
```
webknjaz, line 1208: ^ this is a good use of a comment

cheroot/server.py
```diff
@@ -1521,6 +1535,14 @@ def _close_kernel_socket(self):
...
             if e.errno not in errors.acceptable_sock_shutdown_error_codes:
                 raise
+        except Exception as exc:
+            if SSL is not None and isinstance(
+                exc,
+                (SSL.Error, SSL.SysCallError),
+            ):
+                pass
+            else:
+                raise
```
webknjaz: I'm assuming this is for temporary experimentation and will be removed? This should never happen once the adapters convert exceptions into a common shape like `errors.FatalSSLAlert`.

webknjaz: I'm starting to thing that it may make sense to make that exception inherit from `ConnectionError` as it's semantically a subtype of what might happen with a network connection.

cheroot/ssl/builtin.py
webknjaz: Marking as resolved. And also marking this module as reviewed.

cheroot/ssl/__init__.py
```diff
@@ -56,13 +56,27 @@ def _ensure_peer_speaks_https(raw_socket, /) -> None:
 
 
 class Adapter(ABC):
-    """Base class for SSL driver library adapters.
+    """
+    Base class for SSL driver library adapters.
```
webknjaz: I wouldn't update this. The original same-line style is a pretty popular convention and its fully PEP 257-compliant.
```suggestion
    """Base class for SSL driver library adapters.
```

Keeping it as is will also make the diffs smaller and thus easier to reason about.

cheroot/ssl/pyopenssl.py
```diff
@@ -345,9 +336,11 @@ def wrap(self, sock):
         )
 
         conn = SSLConnection(self.context, sock)
-
         conn.set_accept_state()  # Tell OpenSSL this is a server connection
-        return conn, self._environ.copy()
+
+        # Wrap the SSLConnection to provide standard socket interface
+        tls_socket = TLSSocket(conn)
```
webknjaz: I don't like that we're attempting to access the private attribute `SSLConnection._socket`. Perhaps, we could ask PyOpenSSL to expose it. But for now, we do have that true socket object visible in this scope already
Let's just pass it to the initializer as its own argument and that way we wouldn't be trying to make use potentially unreliable API.

```suggestion
        tls_socket = TLSSocket(underlying_socket=sock, tls_connection=conn)
```

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+    Wrap ``SSL.Connection`` to work with ``StreamReader``/``StreamWriter``.
+
+    Handles OpenSSL-specific errors internally so that standard Python I/O
+    classes can use it transparently.
+    """
+
+    def __init__(self, ssl_conn, ssl_timeout=None, ssl_retry=0.01):
+        """Initialize with an SSL.Connection object."""
+        self._ssl_conn = ssl_conn
+        self._sock = ssl_conn._socket  # Store reference to raw TCP socket
```
webknjaz: Taking into account https://github.com/cherrypy/cheroot/pull/805/files#r2617584138, and the fact that the number of args is growing, let's make them kw-only:

```suggestion
    def __init__(self, underlying_socket, tls_connection, ssl_timeout=None, ssl_retry=0.01):
        """Initialize with an :class:`SSL.Connection <pyopenssl:OpenSSL.SSL.Connection>` object."""
        self._ssl_conn = tls_connection
        self._sock = underlying_socket  # Store reference to raw TCP socket
```

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
-        # This is from past:
-        # TODO: figure out what it's meant for
-        return cheroot_server.CP_fileobject(sock, mode, bufsize)
+
+        return sock.makefile(mode, bufsize)
+
+
+class TLSSocket(SSLFileobjectMixin):
+    """
+    Wrap ``SSL.Connection`` to work with ``StreamReader``/``StreamWriter``.
```
webknjaz: This should work
```suggestion
    Wrap :class:`!SSL.Connection` to work with :class:`StreamReader`/:class:`StreamWriter`.
```
*(https://cheroot--805.org.readthedocs.build/en/805/pkg/cheroot.ssl.pyopenssl/#cheroot.ssl.pyopenssl.TLSSocket)*

cheroot/ssl/pyopenssl.py
```diff
@@ -461,17 +454,110 @@ def get_environ(self):
...
+        """Shutdown the connection."""
+        # SSL shutdown can fail if the handshake didn't complete or during
+        # other error conditions. Let errors propagate to be handled by
+        # the server
+        return self._ssl_conn.sock_shutdown(how)
+
+    # C901 too complex
+    def close(self):  # noqa: C901
+        """
+        Close the TLS socket and underlying connection.
```
webknjaz: Does stdlib `ssl` do both in this method too?

cheroot/test/test_makefile.py
```diff
@@ -48,6 +50,21 @@ def test_bytes_written():
...
     wfile.write(b'bar')
     assert wfile.bytes_written == 3
+
+
+def test_makefile_deprecated():
+    """MakeFile function should emit a deprecation warning."""
+    sock = MockSocket()
+    sock.messages.append(b'foo')
+
+    with pytest.warns(
```
webknjaz:
```suggestion
    with pytest.deprecated_call(
```

cheroot/server.py
webknjaz: Alright, it's probably okay for right now. But I hope that in a follow-up we'll be able to have a better way to reach the underlying socket and won't be accessing the private attributes.

cheroot/server.py
webknjaz: Looks like  with the `TLSSocket` facade in place, we no longer need to reach for `self.socket.sock_shutdown()` and always call `self.socket.shutdown(socket.SHUT_RDWR)`. No need for `getattr()`. Just call the method that will now be common (and hopefully behave the same). After all, this hack is just trying to call the same method on the same underlying raw socket anyway.

webknjaz:
I like the general direction so far, just need to polish a few places I mentioned in the comments above.

cheroot/ssl/pyopenssl.py
julianz-: The problem is `SSLFileobjectMixin` was configured to deal with file-like methods such as 'recv' which returns an array of bytes.  Hence, the error handling would return b'' for EOF.  What we really need for `TLSSocket` is a mixin to support socket-like methods such as 'recv_into' which should return an `int`. For the moment, I have modified SSLFileobjectMixin to behave like a socket-like interface - which means it now returns 0 instead of b'' for EOF and no longer needs `recv()` and `readline()`. It should really be renamed `TLSSocketMixin`. But maybe for a later release?

cheroot/ssl/pyopenssl.py
julianz-: I did some rethinking on `shutdown()` and `close()` for `_TLSSocket`. A challenge is that regular and TLS sockets need different close ordering:

- Non-TLS: shutdown(SHUT_RDWR) sends FIN → close() releases FD
- TLS: SSL close_notify → TCP close/FIN

The server's `HTTPConnection.close()` does:

- `self.rfile.close()`  closes `StreamReader` wrapper, calls `_decref_socketios()`
- `self._close_kernel_socket()`  calls `shutdown(SHUT_RDWR)`
- `self.socket.close()`

This works for regular sockets but creates the wrong order for TLS (TCP shutdown before SSL shutdown).

One solution (without making the server SSL aware) is to turn `_TLSSocket.shutdown()` into a no-op and handle everything in `_TLSSocket.close()`:

- `shutdown(how)` → no-op (for interface compatibility)
- `close()` → SSL shutdown, then TCP close (correct order)

This lets the server's existing code work correctly with TLS. I think stdlib's SSLSocket.shutdown() takes a similar approach but it's less graceful - it abandons the SSL state with `self._sslobj = None` and then does TCP shutdown, skipping SSL `close_notify`. Our implementation attempts graceful SSL shutdown with more complete error handling.

cheroot/server.py
julianz-: I tried changing `FatalSSLAlert` to inherit from `ConnectionError` but this caused issues in `HTTPConnection.communicate()` with the placement of except errors.FatalSSLAlert as it needs to move higher up the list of exceptions.  I tried moving it but this then gave me a linter error complaining about there being no comment.  Some of these changes I find myself going round in circles with endless linter errors for no particularly good reason. So I am leaving this for now.

cheroot/ssl/pyopenssl.py
julianz-: Changed to _TLSSocket.

cheroot/ssl/pyopenssl.py
julianz-: See my comment about `shutdown()` and `close()` above.

Issue #767, opened 2025-09-22
Python 3.14 compatibility


webknjaz:
Currently, a bunch of tests are failing in CI under Python 3.14. They need to be inspected and addressed. I don't know if the patch would need test-only changes or more runtime compatibility fixes.

This might be a good first issue for somebody with network programming + multi-env testing + multi-python compat experience.

thearpankumar:
@webknjaz assign me this issue

Agentx402:
Thanks for bringing this issue to light! It sounds like a great opportunity for someone looking to dive into multi-Python compatibility. I recommend checking out the Python documentation on [version-specific changes](https://docs.python.org/3/whatsnew/3.14.html) to help identify any potential breaking changes that could be affecting your tests. Keep up the good work!

thearpankumar:
@webknjaz so i checked the CI run, tox in 3.12 windows-2025 is also failing with 3.14 builds, i need to check what is up with that locally, what are version changes but any directions which way i should look ??

webknjaz:
Some tests are flaky. And part of them may be addressed by solving #768 (smbd tried and got stuck in #772, it seems). There's also #770 and #777 that might help once complete.

When a flaky test fails a pytest run (through tox), the CI job will rerun it with more verbosity (and purposefully fail that rerun step even if the test succeeds).
When a job fails (and is not marked as allowed to fail), the CI will auto-cancel other jobs in the same matrix.

Some of the suppressed warnings turned into errors in Python 3.14, I think. This might be the root cause for those failures. Somebody need to check that.

This seems like a legitimate failure you should be able to reproduce locally: https://github.com/cherrypy/cheroot/actions/runs/18230474859/job/51912515655#step:21:423 / https://github.com/cherrypy/cheroot/actions/runs/18230474859/job/51912516257#step:20:479.

I recommend you to focus on one small thing at a time and try to understand what causes it, whether it's reproducible consistently and so on. Document your findings and suggest small fixes where possible. Since some problems are difficult to grasp, the fixes might be difficult to judge and you'll need to get into detail when explaining what's broken and why you think the solution is right.

webknjaz:
Also, do you have any proof that xdist is problematic? Any upstream explanations/fixes?

PR #820, opened 2026-04-07
Refactor  _from server socket

julianz-:
This function in connections.py had become highly nested and complex, triggering linter warnings (C901, WPS505). Decomposed the logic into several smaller private methods.

* [ ] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [x] 📋 refactoring
* [ ] 💥 other

Resolves #

None

📋 **Other information**:
📋 **Contribution checklist:**

Will squash on approval


avinashkamat48 (commented):
In `_wrap_socket_for_tls`, the `FatalSSLAlert` branch now returns `(None, {})` immediately, so the cleanup block below never closes `raw_socket` for that error path. The old `_from_server_socket` path simply returned, but with this refactor the helper comment says both TLS failure paths should close the socket. The `NoSSLError` branch reaches the close block, while `FatalSSLAlert` leaks the accepted socket unless the adapter already closed it. Could you either close before returning in that branch or move the common cleanup into a `finally`/single exit path?

cheroot/connections.py
```diff
@@ -295,6 +295,27 @@ def _remove_invalid_sockets(self):
             with _cm.suppress(OSError):
                 conn.close()
 
+    def _setup_conn_addr(self, conn, s, addr):
```
webknjaz: Let's use full words for variable names (things like i/j/k/s aren't really descriptive, nor are the shortened words). It'd also be a good idea to settle for all the args being keyword-only or positional-only (seems like in this case positional args of different types make up bad DX).

cheroot/connections.py
```diff
@@ -316,6 +316,23 @@ def _setup_conn_addr(self, conn, s, addr):
         conn.remote_addr = addr[0]
         conn.remote_port = addr[1]
 
+    def _ignore_socket_oserror(self, exc):
+        if self.server.stats['Enabled']:
+            self.server.stats['Socket Errors'] += 1
```
webknjaz, line 326: This probably belongs outside a function saying "ignore". Though, maybe the helper could be renamed since it doesn't actually ignore anything but checks for expected exceptions.

How about something like `def _is_suppressable_exc(exc: OSError, /) -> bool:`? OTOH, could you check if specific `OSError` subclasses aren't being raised? Maybe, we could replace checking the low-level codes with suppressing the corresponding specific exceptions like `InterruptedError` and similar?

cheroot/connections.py
```diff
@@ -316,6 +316,23 @@ def _setup_conn_addr(self, conn, s, addr):
...
+        is_nonblocking = err_code in errors.socket_errors_nonblocking
+
+        # Our socket was closed. See https://github.com/cherrypy/cherrypy/issues/686.
+        is_ignored = err_code in errors.socket_errors_to_ignore
+
+        return is_eintr or is_nonblocking or is_ignored
+
     def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
         try:
             s, addr = server_socket.accept()
```
webknjaz: I wonder why can't we wrap this line with a suppression of a single exception instead of handling a bunch of errors that could come from a lot of arbitrary places within the try-block..

cheroot/connections.py
```diff
@@ -317,6 +317,7 @@ def _setup_conn_addr(self, conn, s, addr):
         conn.remote_port = addr[1]
 
     def _ignore_socket_oserror(self, exc):
+        """Determine if an OSError during socket operations should be ignored."""
```
webknjaz: "determine" is another clue that the helper probably shouldn't say "ignore".

cheroot/connections.py
```diff
@@ -346,25 +376,10 @@ def _from_server_socket(self, server_socket):  # noqa: C901  # FIXME
...
-                except errors.NoSSLError as http_over_https_err:
-                    self.server.error_log(
-                        f'Client {addr!s} attempted to speak plain HTTP into '
-                        'a TCP connection configured for TLS-only traffic — '
-                        'trying to send back a plain HTTP error response: '
-                        f'{http_over_https_err!s}',
-                    )
-                    self._send_bad_request_plain_http_error(s)
+                s, ssl_env = self._wrap_socket_for_tls(s, addr)
+                if not s:
```
webknjaz: Why can't we communicate this through raising a unified exception (`ConnectionError`? `RuntimeError`?) instead of relying on turning one of the returned tuple values into a bool for deciding to bail?

cheroot/connections.py
```diff
@@ -333,6 +334,35 @@ def _ignore_socket_oserror(self, exc):
...
+        except errors.NoSSLError as http_over_https_err:
+            self.server.error_log(
+                f'Client {addr!s} attempted to speak plain HTTP into '
+                'a TCP connection configured for TLS-only traffic — '
+                'trying to send back a plain HTTP error response: '
+                f'{http_over_https_err!s}',
+            )
+            self._send_bad_request_plain_http_error(raw_socket)
+
+        # If we hit either exception, close the socket and signal failure
```
webknjaz, line 364: why not do this in the fatal ssl alert case too?

cheroot/connections.py
```diff
@@ -364,45 +364,46 @@ def _wrap_socket_for_tls(self, raw_socket, addr):
...
-
-            conn = self.server.ConnectionClass(self.server, s, mf)
-
-            if not isinstance(self.server.bind_addr, (str, bytes)):
-                self._setup_conn_addr(conn, s, addr)
-
-            conn.ssl_env = ssl_env
-            return conn
-
         except socket.timeout:
```
webknjaz, line 374: We may want to leave a note that this is a subclass of OSError since Python 3.3 and since Python 3.10, it's a deprecated alias of TimeoutError.

switch_mode.sh
webknjaz, line 1: accidental commit?

cheroot/connections.py
```diff
@@ -385,28 +385,37 @@ def _configure_socket(self, s):
...
             s.settimeout(self.server.timeout)
 
-    def _from_server_socket(self, server_socket):
-        # Accept the raw connection
+    def _prepare_socket(self, server_socket):
+        """Handle physical accept and TLS negotiation."""
         result = self._accept_conn(server_socket)
         if result is None:
-            return None
+            return None, None, None, None
```
webknjaz: Having an interface returning a tuple of multiple differently typed values is rather fragile.

cheroot/test/_socket_tracker.txt
webknjaz, line 1: Why `.txt`?

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,189 @@ def test_prepare_socket_emits_deprecation_warning(
...
+    # Simulate an interrupt error
+    exc = OSError(errors.socket_error_eintr[0])
+
+    result = conn_manager._ignore_socket_oserror(exc)
+
+    assert result is True
+    assert conn_manager.server.stats['Socket Errors'] == 1
+
+
+def test_ignore_socket_oserror_disabled_stats(conn_manager):
```
webknjaz: Looks like this is the same test with different values. Use `@pytest.mark.parametrize` in such cases.

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,189 @@ def test_prepare_socket_emits_deprecation_warning(
...
+    expected,
+):
+    """Test the socket OSError handler for ignorable errors."""
+    exc = OSError(err_code)
+    assert conn_manager._ignore_socket_oserror(exc) is expected
+
+
+def _raise_eintr(*args, **kwargs):
+    """Raise an interrupt error."""
+    raise OSError(errno.EINTR, 'Interrupted system call')
```
webknjaz: Wouldn't this be https://docs.python.org/3/library/exceptions.html#InterruptedError IRL?

cheroot/test/test_ssl.py
```diff
@@ -1154,3 +1160,189 @@ def test_prepare_socket_emits_deprecation_warning(
...
+    if hasattr(_connections, 'fcntl'):
+        # Patch the function
+        monkeypatch.setattr(_connections.fcntl, 'fcntl', _dummy_fcntl)
+        monkeypatch.setattr(_connections.fcntl, 'F_GETFD', 1)
+
+    # Use the provided socket or fall back to the fixture
+    actual_client = scenario.client_s or fake_socket
+
+    # Mock the server socket to return our scenario-specific data
+    fake_server_socket = SimpleNamespace(
```
webknjaz: the first half could probably go into a fixture with the rest spread across two tests since the logic here is getting nested and tests should be straightforward.

cheroot/test/test_ssl.py
webknjaz, line 1: In general, we should prefer e2e tests since having a lot of mocks will make them tightly coupled with the internals + may make them less reliable.

cheroot/test/test_ssl.py
```diff
@@ -1252,99 +1219,164 @@ def conn_manager_with_server():
...
         bind_addr=('127.0.0.1', 8080),
+        close=mocker.Mock(),
     )
     return mgr
 
 
-@pytest.fixture
-def fake_socket():
-    """Provide a basic mock socket."""
+@contextlib.contextmanager
```
webknjaz: this could be a fixture instead

docs/changelog-fragments.d/820.contrib.rst
```diff
@@ -0,0 +1,5 @@
+Refactored ``_from_server_socket()`` in :py:class:`~cheroot.connections.ConnectionManager`.
+This function had become highly nested and complex, triggering
+linter warnings (C901, WPS505). Decomposed the logic into
+several smaller private methods.
+-- by :user:`julianz-`
```
webknjaz, line 6:
```suggestion

-- by :user:`julianz-`
```

webknjaz (commented):
newer per-commit comments

julianz-:
@avinashkamat48 Thanks for the catch. I have fixed _wrap_socket_for_tls() so that both FatalSSLAlert and NoSSLError now fall through to the shared close block.

cheroot/connections.py
julianz-: Agreed, though I prefer the slightly more intuitive _is_ignorable_socket_error() as the name. I added the / but since none of the other functions have type annotations I left those out — maybe that's better done in a separate PR?

cheroot/connections.py
julianz-: Refactored to handle two different types of exception - _NoConnectionAvailable, ConnectionError. We keep this separation so that transient accept failures (no connection ready) are kept separate from TLS handshake failures.

cheroot/connections.py
julianz-: Updated

cheroot/connections.py
julianz-: Refactored along those lines

cheroot/connections.py
julianz-: Done

cheroot/test/test_ssl.py
julianz-: revised

cheroot/test/test_ssl.py
julianz-: This may have been addressed by subsequent refactoring — the tests have changed significantly since this comment was made. Could you take another look?

cheroot/test/test_ssl.py
julianz-: Agree in principle, although some of these scenarios (TLS handshake failures, transient socket errors) are difficult to trigger reliably in e2e tests without accessing the internals directly.

cheroot/test/test_ssl.py
julianz-: simplified

cheroot/connections.py
julianz-: Done

cheroot/connections.py
julianz-: Fixed

switch_mode.sh
julianz-: Yes - now removed.

cheroot/connections.py
julianz-: Revamped.

cheroot/test/_socket_tracker.txt
julianz-: Removed

cheroot/test/test_ssl.py
julianz-: Revised

julianz-:
Worth noting that there's follow-on work to fix Python 3.14 resource leaks (#767) that I've been holding off on until this refactor lands, since the cleaner structure makes that fix less of a hill to climb.

julianz-:
One fix in this refactor is superseded by PR #838. Will rebase once we have that merged.

PR #822, opened 2026-05-18
Prevent socket write buffer from being cleared when prematurely

### LINKED ISSUE #821: Socket write buffer can be cleared prematurely causing truncated files
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request

When cheroot.makefile.BufferedWriter._flush_unlocked attempts to write to its stream, and no data is able to be written and None is returned which is assigned to `n`.  Then `del self._write_buf[:n]` is called unconditionally and clears the entire write buffer.  This is causing issues where larger files are being truncated, but Cherrypy/Cheroot think the request completed successfully.

Increase the reliability of serving large data sets

💡 **To Reproduce**
Try to send data larger than the buffer and have its stream blocked.

💡 **Expected behavior**
The request completes successfully in full

📋 **Details**
There are no tracebacks or logs associated to this happening. 

📋 **Environment**
- Cheroot version: 11.1.2
- CherryPy version: 18.8.0
- Python version: Noticed on 3.11, but it is not version specific
- OS: Linux
- Browser: all

📋 **Additional context**
When this happens, there are no logs/tracebacks.  Requests to end users "complete" successfully, but the end users may have gotten a truncated file.  This makes it hard to detect it happening, since you would need to look at proxy logs or some other source to see if users are getting the correct sized file

cbbm142:
Adds a check in cheroot.makefile.BufferedWriter._flush_unlocked to prevent clearing of the write buffer when raw.write() returns None due to a blocked stream. Also limits each write call to SOCK_WRITE_BLOCKSIZE bytes, which was defined but not previously used in this method.

Prevent socket write buffer from being cleared when it is blocked
* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

Resolves #821 

https://github.com/cherrypy/cheroot/issues/821

N/A

📋 **Other information**:

This bug is hard to reproduce and spot in production since there are no tracebacks or log messages associated with it.  It happens because n can be None, and then `del self._write_buf[:n]` becomes `del self._write_buf[:]` which will clear everything left in the buffer.

📋 **Contribution checklist:**

[related squash]:

### DIFF
--- a/cheroot/test/test_makefile.py
+++ b/cheroot/test/test_makefile.py
@@ -53,7 +53,26 @@ def test_bytes_written():
     assert wfile.bytes_written == 3
 
 
-def test_flush_preserves_data_when_raw_write_returns_none():
+class _RawWriteBlockOnce:
+    """Mock raw.write() that returns None on the first call, then writes normally."""
+
+    def __init__(self):
+        """Initialize _RawWriteBlockOnce."""
+        self.call_count = 0
+        self.written = bytearray()
+
+    def __call__(self, chunk):
+        """Return None on first call to simulate a blocked socket write."""
+        self.call_count += 1
+        if self.call_count == 1:
+            return (
+                None  # simulates socket returning None on first blocked write
+            )
+        self.written.extend(chunk)
+        return len(chunk)
+
+
+def test_flush_when_raw_write_returns_none():
     """_flush_unlocked() must not treat None from raw.write() as a byte count.
 
     io.RawIOBase.write() returns None when a non-blocking socket cannot accept
@@ -61,29 +80,17 @@ def test_flush_preserves_data_when_raw_write_returns_none():
     which silently clears the entire buffer, truncating the response without
     raising an exception.
     """
-    data = b'x' * 32768  # larger than SOCK_WRITE_BLOCKSIZE to stress the loop
+    data = b'x' * (makefile.SOCK_WRITE_BLOCKSIZE * 2)  # stress the write loop
 
     sock = MockSocket()
     wfile = makefile.MakeFile(sock, 'w')
     wfile._write_buf.extend(data)
 
-    written = bytearray()
-    call_count = 0
-
-    def mock_raw_write(b):
-        nonlocal call_count
-        call_count += 1
-        if call_count == 1:
-            return (
-                None  # simulates socket returning None on first blocked write
-            )
-        written.extend(b)
-        return len(b)
-
-    wfile.raw.write = mock_raw_write
+    mock = _RawWriteBlockOnce()
+    wfile.raw.write = mock
     wfile._flush_unlocked()
 
-    assert bytes(written) == data, (
-        f'Expected {len(data)} bytes but only {len(written)} reached raw.write(): '
+    assert bytes(mock.written) == data, (
+        f'Expected {len(data)} bytes but only {len(mock.written)} reached raw.write(): '
         'buffer was silently discarded when raw.write() returned None'
     )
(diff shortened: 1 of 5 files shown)


avinashkamat48 (commented):
This fixes the data-loss case when 
aw.write() returns None once, but it can hang if the raw stream keeps returning None for the same non-blocking condition. In _flush_unlocked(), 
 stays None, no bytes are deleted, _write_buf remains non-empty, and the while self._write_buf: loop immediately retries forever. It would be safer to either raise/propagate a BlockingIOError or otherwise return control to the caller when no progress is made, rather than spinning inside flush.

cbbm142:
@avinashkamat48 I have added a check to see if the socket becomes writable within 10 seconds.  If it doesn't it raises the blocking error, otherwise it continues to try to send whatever is in the buffer.

julianz-:
@cbbm142 Thanks for this PR. Overall this looks good. A few suggestions:

1. Hardcoded timeout, `SOCK_WRITE_TIMEOUT = 10`, ignores the server's timeout configuration (`HTTPServer.timeout`). Since `HTTPConnection` has access to self.server.timeout, it might be better to pass this timeout into MakeFile → StreamWriter → BufferedWriter as an optional parameter (defaulting to 10 seconds as you had) and use it in the select() call.

2. Add a guard for `n == 0`. If `write()` returns 0 rather than None (could happen for some `RawIOBase` implementations), `del self._write_buf[:0]` does nothing and the loop spins forever. Worth adding a defensive `if n == 0: break` before the `del`.

3. Is the`SOCK_WRITE_BLOCKSIZE` chunking necesaary? `send()` already returns fewer bytes than requested if it can't write everything, and the loop handles the rest.

cheroot/test/test_makefile.py
```diff
@@ -51,3 +51,100 @@ def test_bytes_written():
...
+
+def test_flush_raises_on_sustained_block(monkeypatch):
+    """_flush_unlocked() raises BlockingIOError after select timeout.
+
+    If the socket stays blocked past SOCK_WRITE_TIMEOUT, the write
+    buffer must be preserved and BlockingIOError raised.
+    """
+    import io
+
+    import pytest
```
julianz-, line 128: imports should be placed at the top of the file

cheroot/test/test_makefile.py
```diff
@@ -51,3 +51,100 @@ def test_bytes_written():
...
+
+    data = b'x' * makefile.SOCK_WRITE_BLOCKSIZE
+
+    sock = MockSocket()
+    wfile = makefile.MakeFile(sock, 'w')
+    wfile._write_buf.extend(data)
+
+    mock = _RawWriteBlockAlways()
+    wfile.raw.write = mock
```
julianz-, line 137: Better to patch instance methods using `mocker` because this restores the original write after the test finishes:

`mocker.patch.object(wfile.raw, 'write', new=mock)`

cheroot/test/test_makefile.py
```diff
@@ -51,3 +51,100 @@ def test_bytes_written():
...
+class _RawWriteBlockAlways:
+    """Mock raw.write() that always returns None."""
+
+    def __init__(self):
+        """Initialize _RawWriteBlockAlways."""
+        self.call_count = 0
+
+    def __call__(self, chunk):
+        """Return None to simulate a permanently blocked socket."""
+        self.call_count += 1
```
julianz-, line 86: For clarity, I would add an explicit `return None`

PR #825, opened 2026-06-11
Fix error logging when sending 503 response fails

### LINKED ISSUE #797: Replace `self.server.error_log()` with `self.error_log()` in `HTTPServer._serve_unservicable()`
There is no `self.server` within `HTTPServer`.

_Originally posted by @martin-vi in https://github.com/cherrypy/cheroot/pull/745#discussion_r2409724873_

cc @itamarst @pythonspeed

----

Solving this involves both doing what $sbj says *and* writing a test that covers the respective line.

avinashkamat48:
## Summary
- replace the invalid self.server.error_log call in HTTPServer._serve_unservicable()
- add a focused regression test for exceptions raised while sending the polite 503 response

## Verification
- python -m pytest cheroot/test/test_server.py::test_serve_unservicable_logs_simple_response_errors -q (test passed; command exited nonzero because single-test coverage is below the repo-wide fail-under threshold)
- git diff --check

Closes #797

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -1888,7 +1888,7 @@ class HTTPServer:
                 # We can't just raise an exception because that will kill this
                 # thread, and prevent 503 errors from being sent to future
                 # connections.
-                self.server.error_log(
+                self.error_log(
                     repr(ex),
                     level=logging.ERROR,
                     traceback=True,
--- a/cheroot/test/test_server.py
+++ b/cheroot/test/test_server.py
@@ -22,10 +22,10 @@ from pypytools.gc.custom import DefaultGc
 
 from .._compat import IS_LINUX, IS_MACOS, IS_WINDOWS, SYS_PLATFORM, bton, ntob
 from ..server import (
+    _STOPPING_FOR_INTERRUPT,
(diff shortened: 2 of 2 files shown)


julianz-:
@avinashkamat48 Thanks for this PR. There are a few housekeeping issues here:

1. The function _serve_unservicable() has a spelling mistake (should be unserviceable) which is not your fault but I have changed the PR title to avoid repeating the same error. I am not suggesting we change the name in the code in this PR - just that we don't repeat its use in comments and titles if we can avoid it. 

2. There is a description template with a checklist for PRs which @webknjaz  would like us to use.

3. You need to add a docs/changelog-fragments.d/825.bugfix.rst describing the bug fix here in the appropriate towncrier style - see docs/changelog-fragments.d/README.rst.

4. The commit is missing a body. You can see this in the CI log:

https://github.com/cherrypy/cheroot/actions/runs/27353939982/job/80822995068?pr=825

```
Commit 5ef1cb8897:
3: B6 Body message is missing
```

5. git diff --check  in your "Verification" section would only check for white space issues that pre-commit already checks for so this is not a meaningful test.

Issue #645, opened 2024-01-10
Failing test with openssl 3.2.0 (test_https_over_http_error)


danigm:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**
`test_https_over_http_error` test is failing when running cheroot tests suite in opensuse with openssl 3.2.

❓ **What is the motivation / use case for changing the behavior?**
Test compatibility with latest version of openssl

💡 **To Reproduce**
1. Install openssl 3.2, you can do that in opensuse from this repository https://download.opensuse.org/repositories/security:tls/openSUSE_Tumbleweed/security:tls.repo
2. Run cheroot tests
3. See error

💡 **Expected behavior**
No errors running the tests

📋 **Details**

Looks like in the latest version the error returned when trying to connect to https with http is different.

This is the current test output:
```
[   36s] _____________________ test_https_over_http_error[0.0.0.0] ______________________
[   36s] [gw2] linux -- Python 3.9.18 /usr/bin/python3.9
[   36s] 
[   36s] http_server = <generator object http_server.<locals>.start_srv at 0x7f5ace79ac10>
[   36s] ip_addr = '0.0.0.0'
[   36s] 
[   36s]     @pytest.mark.parametrize(
[   36s]         'ip_addr',
[   36s]         (
[   36s]             ANY_INTERFACE_IPV4,
[   36s]             ANY_INTERFACE_IPV6,
[   36s]         ),
[   36s]     )
[   36s]     def test_https_over_http_error(http_server, ip_addr):
[   36s]         """Ensure that connecting over HTTPS to HTTP port is handled."""
[   36s]         httpserver = http_server.send((ip_addr, EPHEMERAL_PORT))
[   36s]         interface, _host, port = _get_conn_data(httpserver.bind_addr)
[   36s]         with pytest.raises(ssl.SSLError) as ssl_err:
[   36s]             http.client.HTTPSConnection(
[   36s]                 '{interface}:{port}'.format(
[   36s]                     interface=interface,
[   36s]                     port=port,
[   36s]                 ),
[   36s]             ).request('GET', '/')
[   36s]         expected_substring = (
[   36s]             'wrong version number' if IS_ABOVE_OPENSSL10
[   36s]             else 'unknown protocol'
[   36s]         )
[   36s] >       assert expected_substring in ssl_err.value.args[-1]
[...]

Issue #736, opened 2024-11-30
Cheroot ('s test suite?) is hardcoding error codes specific to x86 Linux


mgorny:
❓ **I'm submitting a ...**
- [X] 🐞 bug report
- [ ] 🐣 feature request
- [ ] ❓ question about the decisions made in the repository

🐞 **Describe the bug. What is the current behavior?**
At least the test suite (haven't checked other code) seems to hardcode a number of `errno.h` error codes, based on the values present on x86 Linux. This causes the tests to fail on other Linux architectures, particularly Alpha, MIPS, PA/RISC, SPARC.

❓ **What is the motivation / use case for changing the behavior?**
A passing test suite would help us verify that the package will work on expected on these architectures that aren't tested upstream.

💡 **To Reproduce**
Run `pytest` on one of the affected architectures.

💡 **Expected behavior**
The package not hardcoding errno values, and instead using the value obtained from the `errno` Python module.

📋 **Details**

For example, on Gentoo Linux SPARC we're seeing the following failures:

<details>

```pytb
=================================== FAILURES ===================================
_______________ test_plat_specific_errors[err_names1-err_nums1] ________________

err_names = ('EPROTOTYPE', 'EAGAIN', 'EWOULDBLOCK', 'WSAEWOULDBLOCK', 'EPIPE')
err_nums = (91, 11, 32)

    @pytest.mark.parametrize(
        ('err_names', 'err_nums'),
        (
            (('', 'some-nonsense-name'), []),
            (
                (
                    'EPROTOTYPE', 'EAGAIN', 'EWOULDBLOCK',
                    'WSAEWOULDBLOCK', 'EPIPE',
                ),
                (91, 11, 32) if IS_LINUX else
                (32, 35, 41) if IS_MACOS else
                (98, 11, 32) if IS_SOLARIS else
                (32, 10041, 11, 10035) if IS_WINDOWS else
                (),
            ),
        ),
    )
    def test_plat_specific_errors(err_names, err_nums):
        """Test that ``plat_specific_errors`` gets correct error numbers list."""
        actual_err_nums = errors.plat_specific_errors(*err_names)
        assert len(actual_err_nums) == len(err_nums)
>       assert sorted(actual_err_nums) == sorted(err_nums)
E       assert [11, 32, 41] == [11, 32, 91]
E         
E         At index 2 diff: 41 != 91
E         
E         Full diff:
E           [
E               11,
E               32,
E         -     91,
E         ?     ^
E         +     41,
[...]

PR #828, opened 2026-06-21
Fix test_https_over_http_error on Windows with OpenSSL 3.1+

julianz-:
`test_https_over_http_error` started failing in the CI on Windows with Python 3.14 after recent GitHub Actions runner image updates bumped Python 3.14.5 → 3.14.6 and OpenSSL 3.6.2 → 3.6.3 (win25/20260614.167 deployed ~June 19, win22/20260616.203 deployed shortly after).

The test expects record layer failure when `IS_ABOVE_OPENSSL31` is `True` (OpenSSL > 3.1), but with Python 3.14.6 bundling OpenSSL 3.6.3 on Windows both windows-2025 and windows-2022 now return `wrong version number` instead. This is a known Windows behavior: the TCP reset is surfaced to userspace before OpenSSL fully processes the server's response, causing a different error code to be reported (documented in [CPython's own test_ssl.py](https://github.com/python/cpython/blob/1b9fe5c7226eccc8b269a7443148033080399f43/Lib/test/test_ssl.py#L5705)).

Linux is currently unaffected — Ubuntu 24.04 uses system OpenSSL 3.0.x, so `IS_ABOVE_OPENSSL31` is `False` there and the test falls through to the elif `IS_ABOVE_OPENSSL10` branch, which already expects `wrong version number`.

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

CI errors:

```python-traceback
>       assert expected_substring in ssl_err.value.args[-1]
E       AssertionError: assert 'record layer failure' in '[SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1082)'

```
e.g.

https://github.com/cherrypy/cheroot/actions/runs/27889445941/job/82595847403?pr=827

📋 **Other information**:

Related past patches:
* #645
* #655
* https://github.com/cherrypy/cheroot/commit/4470ae9849c2d60898f18e0e92bd4b585657cad2
* https://github.com/cherrypy/cheroot/commit/04414124582c06e787b88dc5aae24a3ffc16e293

📋 **Contribution checklist:**

### DIFF
--- a/cheroot/test/test_ssl.py
+++ b/cheroot/test/test_ssl.py
@@ -527,7 +527,9 @@ def test_tls_client_auth(  # noqa: C901, WPS213  # FIXME
                 "'Remote end closed connection without response'))",
             )
 
-        assert any(e in err_text for e in expected_substrings)
+        assert any(e in err_text for e in expected_substrings), (
+            f'Unexpected error text: {err_text!r}'
+        )
 
 
 @pytest.mark.parametrize(  # noqa: C901  # FIXME
@@ -697,14 +699,26 @@ def test_https_over_http_error(http_server, ip_addr):
         http.client.HTTPSConnection(
             f'{interface}:{port}',
         ).request('GET', '/')
-    expected_substring = (
-        'record layer failure'
-        if IS_ABOVE_OPENSSL31
-        else 'wrong version number'
-        if IS_ABOVE_OPENSSL10
-        else 'unknown protocol'
-    )
-    assert expected_substring in ssl_err.value.args[-1]
+    underlying_error_string = str(ssl_err.value)
+    if IS_ABOVE_OPENSSL31 and IS_WINDOWS:
+        # On Windows, the TCP reset (RST) is surfaced before OpenSSL processes
+        # the server's response, yielding 'wrong version number' rather than
+        # the usual 'record layer failure'. See CPython's own test_ssl.py:
+        # https://github.com/python/cpython/blob/\
+        # 1b9fe5c7226eccc8b269a7443148033080399f43\
+        # /Lib/test/test_ssl.py#L5705
+        assert 'wrong version number' in underlying_error_string
+    elif IS_ABOVE_OPENSSL31:
+        # Newer Python returns 'record layer failure'; Python 3.8 on macOS
+        # returns 'wrong version number' despite OpenSSL >= 3.1.
+        assert (
+            'record layer failure' in underlying_error_string
+            or 'wrong version number' in underlying_error_string
+        )
+    elif IS_ABOVE_OPENSSL10:
+        assert 'wrong version number' in underlying_error_string
+    else:  # pragma: no cover
+        assert 'unknown protocol' in underlying_error_string
 
 
 def test_http_over_https_no_data(mocker):
--- a/cheroot/test/test_ssl.py
+++ b/cheroot/test/test_ssl.py
@@ -527,9 +527,7 @@ def test_tls_client_auth(  # noqa: C901, WPS213  # FIXME
                 "'Remote end closed connection without response'))",
             )
 
-        assert any(e in err_text for e in expected_substrings), (
-            f'Unexpected error text: {err_text!r}'
-        )
+        assert any(e in err_text for e in expected_substrings)
 
 
 @pytest.mark.parametrize(  # noqa: C901  # FIXME
@@ -700,17 +698,9 @@ def test_https_over_http_error(http_server, ip_addr):
             f'{interface}:{port}',
         ).request('GET', '/')
     underlying_error_string = str(ssl_err.value)
-    if IS_ABOVE_OPENSSL31 and IS_WINDOWS:
-        # On Windows, the TCP reset (RST) is surfaced before OpenSSL processes
-        # the server's response, yielding 'wrong version number' rather than
-        # the usual 'record layer failure'. See CPython's own test_ssl.py:
-        # https://github.com/python/cpython/blob/\
-        # 1b9fe5c7226eccc8b269a7443148033080399f43\
-        # /Lib/test/test_ssl.py#L5705
-        assert 'wrong version number' in underlying_error_string
-    elif IS_ABOVE_OPENSSL31:
-        # Newer Python returns 'record layer failure'; Python 3.8 on macOS
-        # returns 'wrong version number' despite OpenSSL >= 3.1.
+    if IS_ABOVE_OPENSSL31:
+        # 'record layer failure' is typical, but Windows and macOS/Python 3.8
+        # yield 'wrong version number' instead.
         assert (
             'record layer failure' in underlying_error_string
             or 'wrong version number' in underlying_error_string
--- a/docs/changelog-fragments.d/828.contrib.rst
+++ b/docs/changelog-fragments.d/828.contrib.rst
@@ -1,0 +1,3 @@
+Fixed ``test_https_over_http_error`` failing on Windows with OpenSSL 3.1+,
+where the SSL error message is ``wrong version number`` rather than
+``record layer failure`` -- by :user:`julianz-`.


docs/changelog-fragments.d/828.contrib.rst
webknjaz, line 1: Since this patch doesn't fix a bug in Cheroot's runtime, it shouldn't be listed among bugfixes in the change log. Fixing a test is infra work. So it could be a contrib note. Perhaps, a packaging note when we declare that we support new things (like a new Python version, a new OS or a new dep in this case).

Though, feel free to opt out of attaching a change note if it doesn't feel important to surface to the change log audience. You can add the [`bot:chronographer:skip` label](https://github.com/cherrypy/cheroot/labels/bot:chronographer:skip) and the bot will stop flagging a missing note in the PR.

cheroot/test/test_ssl.py
```diff
@@ -697,14 +697,22 @@ def test_https_over_http_error(http_server, ip_addr):
...
+        # OpenSSL 3.1+ usually reports 'record layer failure', but some builds
+        # report 'wrong version number' instead (e.g. Python 3.14.6 on Windows
+        # where the TCP reset takes a different code path). See CPython:
+        # https://github.com/python/cpython/blob/\
+        # 1b9fe5c7226eccc8b269a7443148033080399f43\
+        # /Lib/test/test_ssl.py#L5705
+        assert (
+            'record layer failure' in actual_error
+            or 'wrong version number' in actual_error
+        ), actual_error
```
webknjaz: This shouldn't be needed since pytest is invoked w/ CLI options that should reveal the vars in scope:
```suggestion
        )
```

webknjaz (commented):
Thanks! I've been meaning to ask you to look into what's changed in the CI env.

See a few notes on the patch inline.

P.S. I wonder how difficult it'd be to have predictable OpenSSL code paths in CI. Is CPython built statically, or can we somehow substitute OpenSSL it uses with something we pre-build? Since we have a few TLS-dependent code paths in runtime, it'd be good to run tests against each of them in CI at some point.

docs/changelog-fragments.d/828.contrib.rst
julianz-: Good points. I changed to a contrib note. 

Regarding CPython, I get your idea but it sounds complicated. I was thinking before the issue was the version of OpenSSL but apparently macOS 15 CI runners use Python 3.14.6 with the same bundled OpenSSL 3.6.3 as Windows, but unlike Windows, macOS was already returning  the correct `record layer failure` error.  So I have modified the code the check for Windows in addition to the OpenSSL version.

webknjaz:
@julianz- I just realized that #655 seems to address the same case: #645. Could you absorb the commit from that PR into your branch, rebasing your structural improvements on top?

If you wish to make a combined commit through interactive rebase, use https://hynek.me/til/easier-crediting-contributors-github/ to give appropriate credit.

docs/changelog-fragments.d/828.contrib.rst
webknjaz: Could you also symlink this note to 655 and 645, mentioning the prior art/contrib by radez?

webknjaz: Yeah, it's probably a good idea to normalize these into a unified check where possible. Maybe, migrating to `requests` in tests over time would be a better solution than trying to rely on low-level stdlib stuff. But this is something to think about separately.

julianz-:
@webknjaz

> I just realized that #655 seems to address the same case: #645. Could you absorb the commit from that PR into your branch, rebasing your structural improvements on top?

I looked into rebasing using radez's commit from #655, but your commit [4470ae98](https://github.com/cherrypy/cheroot/commit/4470ae9849c2d60898f18e0e92bd4b585657cad2) ("🧪 Add OpenSSL 3.2 expectations to tests", pushed directly to main in Feb 2025) already incorporated similar changes — `IS_ABOVE_OPENSSL31` in `_compat.py` and the 3-level test logic in `test_ssl.py`. So rebasing produces an empty diff.

docs/changelog-fragments.d/828.contrib.rst
julianz-: I get the intent but wouldn't this make error extraction more complex? `requests` wraps the SSL error through `urllib3`, so accessing the underlying message requires a deeper chain than ssl_err.value.args[-1]?

cheroot/test/test_ssl.py
```diff
@@ -697,14 +697,21 @@ def test_https_over_http_error(http_server, ip_addr):
...
         ).request('GET', '/')
-    expected_substring = (
-        'record layer failure'
-        if IS_ABOVE_OPENSSL31
-        else 'wrong version number'
-        if IS_ABOVE_OPENSSL10
-        else 'unknown protocol'
-    )
-    assert expected_substring in ssl_err.value.args[-1]
+    actual_error = ssl_err.value.args[-1]
```
webknjaz: Side note: I wonder if `str(ssl_err)` would give us the same data these days..

docs/changelog-fragments.d/828.contrib.rst
webknjaz: @julianz- I was mainly thinking of trying to avoid relying on the internals instead. I'm not trilled that the tests are fragile due to the fact that OpenSSL works differently under different runtimes. Technically, this is something external to us and so it'd be a good idea to have it abstracted away, but still something will end up having to normalize the error processing.

Anyway.. Let's postpone this for now.

cheroot/test/test_ssl.py
```diff
@@ -697,14 +697,21 @@ def test_https_over_http_error(http_server, ip_addr):
...
+    if IS_ABOVE_OPENSSL31 and IS_WINDOWS:
```
webknjaz: @julianz- looks like it's not just Windows: https://github.com/cherrypy/cheroot/actions/runs/28048216203/job/83032035095?pr=828#step:15:131

cheroot/test/test_ssl.py
webknjaz: Also, can we rename `actual_error` to `underlying_error_string`?

webknjaz:
Looks like there's another test to be addressed? https://github.com/cherrypy/cheroot/actions/runs/28886497404/job/85688299440?pr=828#step:21:393

cheroot/test/test_ssl.py
```diff
@@ -527,7 +527,9 @@ def test_tls_client_auth(  # noqa: C901, WPS213  # FIXME
                 "'Remote end closed connection without response'))",
             )
 
-        assert any(e in err_text for e in expected_substrings)
+        assert any(e in err_text for e in expected_substrings), (
+            f'Unexpected error text: {err_text!r}'
```
webknjaz: @julianz- why did you think this was necessary? pytest should already expose this automatically.

cheroot/test/test_ssl.py
```diff
@@ -697,14 +699,26 @@ def test_https_over_http_error(http_server, ip_addr):
...
-    expected_substring = (
-        'record layer failure'
-        if IS_ABOVE_OPENSSL31
-        else 'wrong version number'
-        if IS_ABOVE_OPENSSL10
-        else 'unknown protocol'
-    )
-    assert expected_substring in ssl_err.value.args[-1]
+    underlying_error_string = str(ssl_err.value)
+    if IS_ABOVE_OPENSSL31 and IS_WINDOWS:
```
webknjaz: Perhaps, we could just keep the first two blocks (if+elif) as one if-block w/o distinguishing by whether this is windows? Just merge them but keep the explanatory comments. WDYT?

cheroot/test/test_ssl.py
julianz-: I had a failure and for some reason could not see what the failure was but you're right - it's not needed.

cheroot/test/test_ssl.py
julianz-: yes makes sense

cheroot/test/test_ssl.py
```diff
@@ -697,14 +697,18 @@ def test_https_over_http_error(http_server, ip_addr):
...
-        else 'unknown protocol'
-    )
-    assert expected_substring in ssl_err.value.args[-1]
+    underlying_error_string = str(ssl_err.value)
+    if IS_ABOVE_OPENSSL31:
+        # 'record layer failure' is typical, but Windows and macOS/Python 3.8
+        # yield 'wrong version number' instead.
+        assert (
+            'record layer failure' in underlying_error_string
+            or 'wrong version number' in underlying_error_string
```
webknjaz, line 706: I was considering having the strings in dedicated variables like `expected_substring_openssl31plus` etc. But it's probably not going to simplify this.

webknjaz:
Another test is failing: https://github.com/cherrypy/cheroot/actions/runs/29217454416/job/86716151926?pr=828#step:15:344. Although, I can imagine it's a flaky incarnation of #736 (sort of).

webknjaz:
Since that failure was flaky and restarting the job made it green, I'll merge this PR, but we should probably add another error message variant to that flaky test.

julianz-:
This failure in test_tls_client_auth on Windows Python 3.14 is actually an example of why I thought having the assertion message could be useful - err_text isn't shown in the output. You asked me why I added it before and failing to find the log in question, I agreed it wasn't needed. But this log shows why it might be useful after all?

`>           assert any(e in err_text for e in expected_substrings)
E           assert False
E            +  where False = any(<generator object test_tls_client_auth.<locals>.<genexpr> at 0x0000022F08750BA0>)
`

webknjaz:
@julianz- no, `err_text` is visible a few lines below together with other local vars: https://github.com/cherrypy/cheroot/actions/runs/29217454416/job/86716151926?pr=828#step:15:344

julianz-:
ah my bad. missed it. the error string is almost identical with one that would have passed. Looks like Python 3.14 is dropping the extra params: `None, 10054, None).`?

webknjaz:
> ah my bad. missed it.

It's configured in `pytest.ini` via `--show-locals`, if you're curious. I also sometimes have a hard time finding the correct line in the log because the vars are printed on many levels of the trace. I think it's useful if you try to build a mental model of what such logs look like with this feature on. Once you know it's there, you can use the search field on the job log page, with the respective step unfolded, and pasting the exact variable name can get you to places where the values are recorded rather quickly.

> the error string is almost identical with one that would have passed. Looks like Python 3.14 is dropping the extra params: `None, 10054, None).`?

I can't look now. But make a PR with the context documented and we can talk about the possibility of simplifying the test, perhaps. This might require in-depth research, though.

julianz-:
> I can't look now. But make a PR with the context documented and we can talk about the possibility of simplifying the test, perhaps. This might require in-depth research, though.

I've added #838. I had forgotten I saw a similar error albeit on a Windows Python 3.15 machine when working on #820.

PR #832, opened 2026-06-27
Reject NUL bytes in HTTP request lines

andrewkernel:

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [x] 📋 tests/coverage improvement
* [ ] 📋 refactoring
* [ ] 💥 other

Related to #201.

HTTP request lines containing embedded NUL bytes could reach the request-line parser instead of being rejected explicitly as malformed input.

Cheroot now rejects request lines containing NUL bytes with `400 Bad Request` before splitting the request line. The malformed request-line test matrix covers the NUL-byte case.

📋 **Other information**:

Added `docs/changelog-fragments.d/832.bugfix.rst`.

Testing:

* `.venv\Scripts\python.exe -m pytest --no-cov -n 0 cheroot/test/test_core.py -k malformed_request_line`

📋 **Contribution checklist:**

* [x] I wrote good commit messages
* [x] I have squashed related commits together after the changes have been approved
* [x] The PR relates to only one subject with a clear title and description in grammatically correct, complete sentences

### DIFF
--- a/cheroot/server.py
+++ b/cheroot/server.py
@@ -800,6 +800,13 @@ class HTTPRequest:
             )
             return False
 
+        if b'\x00' in request_line:
+            self.simple_response(
+                '400 Bad Request',
+                'Malformed Request-Line',
+            )
+            return False
+
         try:
             method, uri, req_protocol = request_line.strip().split(SPACE, 2)
             if not req_protocol.startswith(b'HTTP/'):
--- a/cheroot/test/test_core.py
+++ b/cheroot/test/test_core.py
@@ -315,6 +315,11 @@ def test_large_request(test_client_with_defaults):
             b'Malformed Request-Line',
         ),
         (
+            b'GET /\x00 HTTP/1.1',  # NUL byte
+            HTTP_BAD_REQUEST,
+            b'Malformed Request-Line',
+        ),
+        (
             b'GET / HTTPS/1.1',  # invalid proto
             HTTP_BAD_REQUEST,
             b'Malformed Request-Line: bad protocol',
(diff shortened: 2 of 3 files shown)


cheroot/test/test_core.py
```diff
@@ -420,6 +420,17 @@ def test_garbage_in(test_client):
             raise
 
 
+def test_null_byte_in_request_line(test_client):
+    """Check that NUL bytes in the request line return Bad Request."""
+    c = test_client.get_connection()
```
julianz-: Let's use descriptive vars rather than single characters. `conn` would be better here.

cheroot/server.py
```diff
@@ -802,6 +802,12 @@ def read_request_line(self):  # noqa: C901  # FIXME
 
         try:
             method, uri, req_protocol = request_line.strip().split(SPACE, 2)
+            if b'\x00' in request_line:
```
julianz-: Better to check for null bytes before doing any parsing. Suggest moving this above `request_line.strip().split(SPACE, 2)`

julianz- (commented):
Thanks for this PR. Looks useful but see inline comments. Also in the PR description you did not use the standard template that makes clear what you have and have not addressed.

Your commit also needs a body message and we need a docs/changelog-fragments.d/832.bugfix.rst describing the bug fix in the appropriate towncrier style - see docs/changelog-fragments.d/README.rst.

cheroot/test/test_core.py
webknjaz: Technically, this follows the existing style, which could be okay in such cases. This, however, mainly copies the same test and should've just parametrized it instead of adding another almost-identical test function.

cheroot/server.py
webknjaz: Even above the try-block. But also, it's probably a good idea to have a list of allowed chars.

In general, though, it'd be better to focus on implementing #201 (and resurrect #262).

PR #840, opened 2026-09-17
Fix AttributeError in the unserviceable-connection error handler

### LINKED ISSUE #797: Replace `self.server.error_log()` with `self.error_log()` in `HTTPServer._serve_unservicable()`

cristianchiriac:

* [x] 🐞 bug fix
* [ ] 🐣 feature
* [ ] 📋 docs update
* [x] 📋 tests/coverage improvement
* [ ] 📋 refactoring
* [ ] 💥 other

## What do these changes do?

`HTTPServer._serve_unservicable()` catches unexpected failures while sending a
`503 Service Unavailable` response and logs them, deliberately, so that the
background thread stays alive and later connections still get their 503:

```python
except Exception as ex:
    # We can't just raise an exception because that will kill this
    # thread, and prevent 503 errors from being sent to future
    # connections.
    self.server.error_log(
        repr(ex),
        level=logging.ERROR,
        traceback=True,
    )
```

The problem is `self.server`. Here `self` is the `HTTPServer` itself, and
`HTTPServer` has no `server` attribute — that belongs to `HTTPConnection`,
whose `self.server` does point back at its owning `HTTPServer`. The two other
`self.server.error_log(...)` calls in this file live in
`HTTPConnection.communicate()`, where they are correct.

So the handler raises a fresh `AttributeError` from inside the except-block,
the original error is never logged, and the thread that is supposed to keep
answering overloaded connections dies — the exact opposite of what the comment
above it promises.

The fix is one line: call `self.error_log(...)`, which `HTTPServer` defines a
few dozen lines further down in the same class.

I also added `test_overload_survives_failure_to_send_http_error`, an
end-to-end test that runs against a real overloaded server. Only the failure
to send the 503 is simulated, by making `HTTPRequest.simple_response()` raise
once. It checks that the original error is logged rather than an
`AttributeError`, and that the next client still receives its 503. The
overload server setup now lives in a fixture that
`test_overload_results_in_suitable_http_error` shares. The test design draws
on #825 and #830, whose authors are credited in the commits and in the change
log entry.

I should mention: you noted this is the third PR for this bug, after #825 and
#830, and I only saw those after your comment. The one-line fix is necessarily
the same in all three. If you would rather land one of the earlier PRs, please
do — I am happy for this one to be closed, and can port the test over there
instead if that helps.

Nothing changes on the happy path. When sending a 503 fails, the original
exception is now written to the error log with its traceback, and the
unserviceable-connection thread keeps running instead of dying with an
unrelated `AttributeError`. Anyone who wrote code against that `AttributeError`
(unlikely — it was a crash, not an API) would notice it is gone.

No. It swaps one attribute lookup for the method on the same class and adds a
test in the existing style of `test_server.py`. There is nothing new to keep
working over the next five years, and the test pins the behaviour so the same
mistake cannot come back unnoticed.

## Related issue number

Fixes #797. Same underlying bug as #825 and #830.

## Checklist

* [x] Integration tests for the changes exist (if applicable) — the new test
      runs against a real overloaded server
* [x] Project documentation (in `docs/`) and inline docstrings reflect the changes
* [x] I have added a [change log entry] — `docs/changelog-fragments.d/797.bugfix.rst`,
      named after the issue since the PR number was not known when I wrote it;
      happy to rename it to `840.bugfix.rst` if you prefer
* [x] I'm planning to squash related commits together before final merge
* [x] I have read the contribution guide and the code of conduct

[change log entry]:
https://github.com/cherrypy/cheroot/tree/main/docs/changelog-fragments.d#readme

## Test run

* `pytest cheroot/test/test_server.py cheroot/test/test_conn.py cheroot/test/test_core.py`
  passes on Python 3.8 and 3.13 (95 passed, 11 skipped, 4 xfailed on each).
* The new test fails without the fix (the overload thread dies and the
  second client never gets its 503) and passes with it.
* The repository's `pre-commit` hooks (`flake8` with its plugins, `ruff`,
  `mypy`, `pylint`) pass on the changed files.


julianz-:
@Tatamis Thanks for the submission. This is actually the third time a fix for this problem has been submitted: see #825 and #830, both still open. These were incomplete submissions for which I left comments on how to complete. Your submission is also incomplete in that it does not use our template for new PRs - `PULL_REQUEST_TEMPLATE.md` although you may not have seen this if you used an automated tool for making the submission. Anyway, if you could revise the PR using this template we can take things from there.

cheroot/test/test_server.py
```diff
@@ -616,6 +616,39 @@ def test_overload_results_in_suitable_http_error(request):
...
+    """A failure while sending a 503 must be logged, not raise AttributeError.
+
+    Regression test: ``HTTPServer._serve_unservicable()`` called
+    ``self.server.error_log(...)``, but ``HTTPServer`` has no ``server``
+    attribute -- that belongs to ``HTTPConnection``, which is a different
+    class. So instead of logging the original failure, it raised a fresh
+    ``AttributeError`` from within the except-block meant to keep this
+    background thread alive. See issue #797.
+    """
+    from .. import server as server_module
```
julianz-: This kind of local, whole-module import shouldn't be needed. `_STOPPING_FOR_INTERRUPT` can just move into the top-level `from ..server import ...` block. For `HTTPRequest`, rather than importing the module just to patch an attribute on it, use mocker.patch('cheroot.server.HTTPRequest', ...) instead — no import needed at all.

docs/changelog-fragments.d/797.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fixed :py:meth:`HTTPServer._serve_unservicable() <cheroot.server.HTTPServer._serve_unservicable>`
+raising an unrelated :py:exc:`AttributeError` instead of logging the original failure
```
julianz-:
```suggestion
raising an unrelated :py:exc:`AttributeError` instead of logging the original failure,
```

cristianchiriac:
Pushed 9e7e3c06 addressing this: `_STOPPING_FOR_INTERRUPT` now comes from the top-level `from ..server import (...)` block, and the `HTTPRequest` patch is `mocker.patch('cheroot.server.HTTPRequest', ...)` instead of patching an attribute on the imported module. Also picked up the comma in the changelog fragment.

julianz-:
Thanks for update. @webknjaz please take a look.

cheroot/test/test_server.py
```diff
@@ -616,6 +621,33 @@ def test_overload_results_in_suitable_http_error(request):
     assert response.status_code == HTTPStatus.SERVICE_UNAVAILABLE
 
 
+def test_serve_unservicable_logs_errors_without_crashing(mocker, capsys):
```
webknjaz: Add type annotations to all newly added test functions.

cheroot/test/test_server.py
```diff
@@ -616,6 +621,33 @@ def test_overload_results_in_suitable_http_error(request):
...
+    """A failure while sending a 503 must be logged, not raise AttributeError.
+
+    Regression test: ``HTTPServer._serve_unservicable()`` called
+    ``self.server.error_log(...)``, but ``HTTPServer`` has no ``server``
+    attribute -- that belongs to ``HTTPConnection``, which is a different
+    class. So instead of logging the original failure, it raised a fresh
+    ``AttributeError`` from within the except-block meant to keep this
+    background thread alive. See issue #797.
```
webknjaz: This sounds more like a code comment rather than a docsring, which has a different audience and purpose.

cheroot/test/test_server.py
```diff
@@ -616,6 +621,33 @@ def test_overload_results_in_suitable_http_error(request):
...
+    """
+    httpserver = HTTPServer.__new__(HTTPServer)
```
webknjaz: This object creation style doesn't seem to exist anywhere. I'd rather not use it.

cheroot/test/test_server.py
```diff
@@ -616,6 +621,33 @@ def test_overload_results_in_suitable_http_error(request):
...
+    httpserver.ready = True
+    httpserver._unservicable_conns = queue.Queue()
+    fake_conn = mocker.Mock()
```
webknjaz: Please don't mock everything. We need to use proper e2e tests as much as possible, only mocking external things/side effects where appropriate.

cheroot/test/test_server.py
webknjaz, line 1: Draw inspiration from #825 / #830 to design a better test with stubs, not mocking everything to the point that the mocks are tested more than the overall behavior.

docs/changelog-fragments.d/797.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fixed :py:meth:`HTTPServer._serve_unservicable() <cheroot.server.HTTPServer._serve_unservicable>`
+raising an unrelated :py:exc:`AttributeError` instead of logging the original failure,
+when sending a ``503 Service Unavailable`` response itself raised an unexpected exception.
+
+-- by :user:`Tatamis`
```
webknjaz: We should probably credit the contributors from #825 and #830 in the test rewrite commit and here: https://hynek.me/til/easier-crediting-contributors-github/

docs/changelog-fragments.d/797.bugfix.rst
webknjaz, line 1: Release notes are generally supposed to be targeting the end-users. This seems to be focusing on the low-level internals. Could you think of making this text more high-level?

cristianchiriac:
@webknjaz thanks for the detailed review. I pushed 2753fd9c (two commits) that addresses all of it:

* **Mocking / e2e:** the test is rewritten as `test_overload_survives_failure_to_send_http_error`. It runs against a real overloaded server with real sockets. The only thing simulated is the failure to send the 503 (`HTTPRequest.simple_response()` raises once). It asserts that the original error is logged, with no `AttributeError`, and that the next client still gets its 503. Design inspired by #825 and #830. It fails without the fix and passes with it.
* **Object creation style:** gone, no hand-built server objects or `Mock` stand-ins any more. The overload server setup is a fixture now, shared with `test_overload_results_in_suitable_http_error`.
* **Type annotations:** added to the new fixture and test functions, and to the existing overload test that now uses the fixture.
* **Docstring vs. comment:** the long explanation is gone. The docstring now only says what the test checks, with a reference to #797.
* **Credit for #825 / #830:** both commits carry `Co-authored-by` trailers for @avinashkamat48 and @andrewkernel, and the change log fragment credits them too.
* **Release notes wording:** rewritten around what users see (the server stops answering overloaded connections with 503 after a failed send), without the internals.

I ran the changed tests on Python 3.8 and 3.13, and the repo's pre-commit hooks (flake8 with wemake, ruff, mypy, pylint) pass on the changed files. The PR description is updated to match. Please take another look when you have time.

docs/changelog-fragments.d/797.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fixed the server stopping to answer overloaded connections with
+``503 Service Unavailable`` after a failure to send one of these responses.
+The failure is now logged and the following connections are still answered.
+
+-- by :user:`Tatamis`, :user:`avinashkamat48` and :user:`andrewkernel`
```
julianz-:
```suggestion
Fixed the server's ability to send ``503 Service Unavailable`` whenever the server becomes too busy.  Failures to send caused the handler for this message to crash leaving new connections during busy times unanswered. Any failure in sending this message is now logged, and subsequent connections continue to be answered.

-- by :user:`Tatamis`, :user:`avinashkamat48` and :user:`andrewkernel`
```

julianz-: Think this easier to understand as an end user even if it still hints at the internal message handler.

cristianchiriac: Thanks, applied as suggested in b9967da7 (rewrapped the lines only). That commit also fixes the docs build: the new test annotations refer to `pytest.MonkeyPatch` and `pytest.CaptureFixture`, which Sphinx can't resolve in nitpicky mode, so I added them to `nitpick_ignore`, like the other unresolvable references.

cheroot/test/test_server.py
```diff
@@ -583,11 +584,11 @@ def test_threadpool_multistart_validation(monkeypatch):
         tp.start()
 
 
-def test_overload_results_in_suitable_http_error(request):
-    """A server that can't keep up with requests returns a 503 HTTP error."""
-    localhost = '127.0.0.1'
```
webknjaz, line 588: Why did you change this?

cheroot/test/test_server.py
```diff
@@ -602,19 +603,76 @@ def test_overload_results_in_suitable_http_error(request):
...
 
-    response = requests.get(f'http://{localhost}:{port}', timeout=20)
     assert response.status_code == HTTPStatus.SERVICE_UNAVAILABLE
 
+    # The original failure is reported instead of a bogus `AttributeError`:
+    captured_stderr = capsys.readouterr().err
+    assert (
+        'RuntimeError: unexpected failure sending the 503' in captured_stderr
+    )
+    assert 'AttributeError' not in captured_stderr
```
webknjaz, line 674: Seems weird checking for such internals leaking.

webknjaz:
So.. Cristian, could you tell me why this bug is so important to you? Where did you hit it?

PR #842, opened 2026-09-19
Convert contributing docs to Markdown

julianz-:

* [ ] 🐞 bug fix
* [ ] 🐣 feature
* [x] 📋 docs update
* [ ] 📋 refactoring
* [ ] 💥 other

## What do these changes do?

1. Converts .github/CONTRIBUTING.md and docs/contributing/guidelines.md from RST to Markdown, using MyST-parser (added in #835) to render them.

2.  Fixes the "Also" section so that it doesn't refer back to itself in the rendered version - it only shows up when reading CONTRIBUTING.md directly on GitHub. 

3. Drops CONTRIBUTING.md's opening pointer to the root README, which sends readers to a file with little relevant additional information for contributors.

4. Updates one link in changelog-fragments.d/README.rst that broke once that file got pulled into the new Markdown page.

None — this only affects contributors reading the docs, not end users

No. It's a one-time format migration. Worth noting though:
`changelog-fragments.d/README.rst` remains RST, embedded via `eval-rst`
into an otherwise-Markdown page — a minor format seam, not a burden.

## Related issue number
 
Discussed with @webknjaz in chat; not tied to a filed issue. 

## Checklist

* [x] Project documentation (in `docs/`) and inline docstrings reflect the changes
      (see [good commit messages])
* [x] I have added a [change log entry]
      (can be done once the PR number is known; alternatively, you can
      use a related issue number if one exists)
* [] I'm planning to [squash related commits] together before final merge
* [x] I have read the [contribution guide] and the
      [code of conduct][CoC]


webknjaz:
@julianz- the CI is failing under Python 3.13+ due to a new validation check in coveragepy. Could you look into cherry-picking https://github.com/re-actors/alls-green/commit/8de05dc10f508bf473c58ea66a5ee252fe7e63bd in a standalone PR to fix the CI overall?

docs/changelog-fragments.d/README.rst
```diff
@@ -111,6 +112,3 @@ File :file:`docs/changelog-fragments.d/359.bugfix.rst`:
 
    See :file:`towncrier.toml` for all available categories
    (``tool.towncrier.type``).
-
-.. _Towncrier philosophy:
```
webknjaz, line 115: Why did you do this? I actually intentionally prefer detached references everywhere so that they are clearly labeled and can be reused.

webknjaz: Oh, the commit message says that one link broke. Did you mean this one?

@julianz- pro tip — you can leave additional notes right in the code review diff, as inline threads, like I do here. This helps enrich the context and saves me from having to guess.

docs/changelog-fragments.d/README.rst
```diff
@@ -6,7 +6,8 @@ Adding change notes with your PRs
 It is very important to maintain a log for news of how
 updating to the new version of the software will affect
 end-users. This is why we enforce collection of the change
-fragment files in pull requests as per `Towncrier philosophy`_.
+fragment files in pull requests as per `Towncrier philosophy
```
webknjaz, line 9: Would the link still break if you keep put it after this paragraph? 🤔

docs/contributing/guidelines.md
```diff
@@ -0,0 +1,137 @@
+```{include} ../../.github/CONTRIBUTING.md
+:end-before: "## Also"
+```
+
+## First-time setup
```
webknjaz, line 5: This sort of gives me vibes of a "this is how you use a computer". Wonder if this is useful at all or should link to something external. Although, I realize that a format conversion PR is not a place to discuss this. So this is just a side note for the future.

docs/changelog-fragments.d/README.rst
webknjaz, line 1: @julianz- what if we updated this to be a `README.md` too? Would it be helpful or do you think this is out of the scope here?

webknjaz (approved):
@julianz- I'll let you decide whether you want to extend the scope a little or merge this as is.

In any case, I'd like a follow-up to reconfigure Towncrier to start using MD for the fragment files to make people's lives easier. This will require some amount of integration work, though. Here's a few places to pick up a few cool tricks from: https://github.com/jazzband/pip-tools/pull/2203/changes#diff-e010bb4b0f729415605939bacb0af8800a91a06d404da48b7e3fac5a74cddc90 / https://github.com/jazzband/pip-tools/pull/2343/changes#diff-e010bb4b0f729415605939bacb0af8800a91a06d404da48b7e3fac5a74cddc90R19

docs/changelog-fragments.d/README.md
```diff
@@ -111,6 +112,3 @@ File :file:`docs/changelog-fragments.d/359.bugfix.rst`:
 
    See :file:`towncrier.toml` for all available categories
    (``tool.towncrier.type``).
-
-.. _Towncrier philosophy:
```
julianz-: Sorry I made you hunt for the explanation. Will try to add comments more proactively as you suggest - thanks for the tip! Yes, this is the link that broke - the Towncrier philosophy reference and its target.

I originally thought to retain as an RST file in order to keep the changes smaller and more manageable.  I see you suggested changing it below though, so maybe I will do that after all. The issue with it as an RST file, is that I had to pull it into the new `guidelines.md` page through an `{eval-rst}` block wrapping the old `include` directive. This generates `ERROR: Unknown target name: "towncrier philosophy"`with detached refs no matter where I place them presumably because of the `eval-rst` embedding.

docs/changelog-fragments.d/README.rst
julianz-: Yes tried that and it still breaks alas!

docs/contributing/guidelines.md
julianz-: Agreed - the noob vibes would be less grating if this were a lot shorter or pointed elsewhere as you suggest!

docs/changelog-fragments.d/README.md
julianz-: See above comment. I think maybe we should do this in order to get the more complete conversion over and done with.

docs/changelog-fragments.d/README.md
```diff
@@ -1,13 +1,11 @@
-.. _Adding change notes with your PRs:
+(adding change notes with your prs)=
```
webknjaz: Is this anchor used anywhere? We'll need to avoid using MyST-specific syntax in a few files that are also special-cased and rendered by GitHub itself. It uses GFM and anything MyST would show up broken in the render.

Examples include files like `CONTRIBUTING.md` and similar that are linked and embedded into the GH UI. So we both render them on GH and in the docs. We may want to split them into parts by using inclusions and keeping MyST syntax on the Sphinx side while only using what GFM allows in the shared parts of the docs. `pip-tools`' examples solved this pretty well.

Another example would be README files in any folders but especially next to the changelog fragments — just take a look: https://github.com/cherrypy/cheroot/tree/main/docs/changelog-fragments.d#readme — the context is right where people have expected to add these files. So I consider it important that this should render well. I suggest you to preview how this file renders in your fork, not only in the Sphinx build but also on GH, on the branch view.

docs/changelog-fragments.d/README.md
```diff
@@ -80,35 +76,34 @@ feature, in which case two fragments should be added. It is not
 necessary to make a separate documentation fragment for documentation
 changes accompanying the relevant code changes.
 
-Examples for adding change log entries to your Pull Requests
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+## Examples for adding change log entries to your Pull Requests
 
-File :file:`docs/changelog-fragments.d/77.doc.rst`:
+File {file}`docs/changelog-fragments.d/77.doc.rst`:
```
webknjaz: For the same reason as above, this needs to drop the file role, I think...

docs/changelog-fragments.d/README.md
```diff
@@ -80,35 +76,34 @@ feature, in which case two fragments should be added. It is not
...
-.. tip::
+```rst
+Fixed a regression from :pr:`199` that made the worker threads exit on
+invalid connection attempts and could make the whole server unresponsive
+once there was no workers left -- by :user:`cameronbrunner`.
+```
 
-   See :file:`towncrier.toml` for all available categories
-   (``tool.towncrier.type``).
+```{tip}
```
webknjaz: This is a MyST construct and GFM will not like it. In `pip-tools`, I've integrated some additional plugin or something for it to also support GFM-native admonitions that use Markdown blockquotes and start with `> [!tip]`

